PR #41, opened 2014-08-12
Dependencies for Fedora 20

tzicatl:
A small note on the dependencies of GTG for Fedora 20.


izidormatusov:
This section was dropped from README.md in meanwhile and thus closing this PR.

PR #54, opened 2014-09-09
Backend GTGOnline for GTG Python3

parinporecha:


GTG/backends/backend_gtgonline.py
```diff
@@ -0,0 +1,671 @@
...
+    
+    WEB_STATUS_TO_GTG = {
+        0: 'Active',
+        1: 'Done',
+        2: 'Dismiss',
+    }
+    
+    def __init__(self, params):
+        """ Constructor of the object """
+        super(Backend, self).__init__(params)
```
izidormatusov: Please, use python3 version:

```python
super().__init__(params)
```

diegogangl:
@nekohayo what do you think about this PR? Do we still support gtgonline?

nekohayo:
Tricky question... I've never used it because I've never known if it was considered stable/finished, and because I didn't have time to be sysadmining this kind of thing for myself. I don't know if GTGOnline is "complete" or not. Did GTG already have GTGOnline code in it, and this improves/fixes it, or this is brand new code insertion that GTG didn't have as a backend? If the former then this could be worth merging, if the latter then it could just be relegated to "volunteers/maintainers needed" on some sort of realigned wiki roadmap...

diegogangl:
This is adding a new backend. Actually, I just looked at the code and it looks like it's Python 2 :/ 
I'll close this one, could you add a mention to the wiki/roadmap?

PR #73, opened 2014-11-17
fix: hamster icon is invisible in task editor

tkdchen:
When enable Hamster plugin, open a task, the Hamster icon does not appear in the toolbar, but the button is added. This patch is to fix this issue.


diegogangl:
Hi @tkdchen, thanks for your work on this! Unfortunately the hamster plugin is unmaintained and blacklisted for the next release, we also don't have a toolbar anymore so we can't merge this anymore :(

tkdchen:
@diegogangl Understood. Thanks for your feedback.

Issue #81, opened 2014-12-25
TaskView._detect_url will run in infinite loop under GTK 3.14.6


tkdchen:
Environment: Fedora 21, GTK 3.14.6, `gi.version_info` equals to (3, 14, 0), GTG version: dev in master branch

This causes that task cannot be opened from task browser.

vrutkovs:
+1, works great on my system

izidormatusov:
Solved by #106

PR #82, opened 2014-12-26
fix: #81 TaskView._detect_url run into infinite loop under GTK 3.14.6

tkdchen:
This bug report describes why this patch is needed for users who are using latest GTK+ release 3.14.x https://bugzilla.gnome.org/show_bug.cgi?id=742000

The original way to test whether to reach the end of text buffer by comparing the iterator's offset is not stable in GTK+ 3.14.x.

Without this patch, some tasks cannot be opened and GTG would crash due to the infinite loop.


izidormatusov:
Fixed by #106

PR #84, opened 2014-12-29
Bugfix 83 Closed tasks remover imposes a too low limit for the days period option

### LINKED ISSUE #83: Closed tasks remover imposes a too low limit for the days period option
The Closed tasks remover has an option to specify after how many days a task should be removed. However, this options is limited to value of 100 days, which is unnecessarily low.

rivaldi8:
Fixes #83


izidormatusov:
Thanks for your PR, sorry it took so long to merge it.

PR #106, opened 2015-05-02
Fix startup errors and style issues

izidormatusov:


izidormatusov:
I find that the first commit of this PR (4183e95) solves opening the editor better than #82

jakubbrindza:
Thank you for putting your commit message in a clear way. Now I get why you made the change with forward_word_end(). Everything else seemed reasonable and right to me.

PR #107, opened 2015-05-12
test with more recent python versions and use sudo:false

tomspur:
With sudo:false, travis can use [containers](http://docs.travis-ci.com/user/workers/container-based-infrastructure/) and has a faster startup time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: python
 
 python:
-  - '3.2'
+  - '3.3'
+  - '3.4'
+  - 'nightly'
 
 before_install:
   - sudo apt-get update
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: python
 
 python:
-  - '3.3'
-  - '3.4'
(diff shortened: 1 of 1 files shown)


tomspur:
Unfortunately, `gir1.2-networkmanager-1.0` cannot be installed yet and needs an [approval](https://github.com/travis-ci/apt-package-whitelist#package-approval-process). So maybe the old way of running the tests is just fine and this can be revisited, if the containers have a bigger advantage to the current way of running tests...

izidormatusov:
Can you file a bug for approval of that package in meanwhile? You could
then reopen PA after it is added

tomspur:
The whitelist requst is at travis-ci/travis-ci#3895 now. I don't think it is save to add that package in the container environment, but we'll see...

izidormatusov:
Other possibility is to make network manager as an optional dependency.
It's nicely isolated in GTG/tools/networkmanager.py and if the dependency
is not available, is_connection_up() fall back to safe *always return True*
.

PR #115, opened 2015-06-01
Fix ImportError: No module named 'GTG.core.tools'

tomspur:
This was introduced in 44fb77d with the reorganization of GTG.core.dirs.


izidormatusov:
Thanks! Good catch!

PR #134, opened 2015-06-10
BackendsDialog: Declare on_help as classmethod

tomspur:
The on_help of the BackendsDialog method should be a classmethod like
the one from the PreferencesDialog.

(Found while testing out @quantifiedcode)


izidormatusov:
Thanks!

PR #146, opened 2015-06-25
Re-design of Task Editor

jakubbrindza:
In this PR, I have re-designed the entire GTG task editor to follow the GNOME3 standards. Despite the new look, it lacks some core functionality.

### Additions & working features:
* re-design of toolbar and implementation of headerbar
* new icons for new functionality (italic / bold)
* popovers for tags, parents or start/due date widgets
* insert multiple tags from the tag popover (only insert, yet)
* opening parent task possible only if the task has 1 parent (yet)
* set start/due/closed date in the corresponding GtkEntry *(calendars not working, yet)*
* insert subtask

### Issues & TODOs:
* Mark as Done + Dismiss
* Remove tags
* Search / Add tags from the top GtkEntry in tags popover
* Open popover if a task has more than one parent
* Re-factor code to separate code where possible *(calendar popovers ==> separate .ui and re-factored calendar.py to work with the calendar widgets in start/due/closed popovers)*
* Make space for and bring back Date label to show the number of remaining days for a task in editor.
* Italic & bold (@izidormatusov)

screen shot 2015-06-25 at 14 59 51

screen shot 2015-06-25 at 15 00 01

screen shot 2015-06-25 at 15 00 15

screen shot 2015-06-25 at 13 54 53


GTG/gtk/editor/taskeditor.ui
```diff
@@ -4,403 +4,461 @@
   <object class="GtkWindow" id="TaskEditor">
     <property name="can_focus">False</property>
     <property name="title" translatable="yes">Task</property>
-    <property name="default_width">500</property>
-    <property name="default_height">400</property>
+    <property name="default_width">800</property>
+    <property name="default_height">500</property>
```
izidormatusov: Is it possible to keep the original default_width/default_height? Nice thing about the GTG task editor window was it  was quite small if you wanted.

GTG/gtk/editor/editor.py
```diff
@@ -65,12 +67,12 @@ def __init__(self,
         self.deletebutton = self.builder.get_object("delete_editor")
         self.deletebutton.set_tooltip_text(GnomeConfig.DELETE_TOOLTIP)
         self.subtask_button = self.builder.get_object("insert_subtask")
-        self.subtask_button.set_tooltip_text(GnomeConfig.SUBTASK_TOOLTIP)
-        self.inserttag_button = self.builder.get_object("inserttag")
-        self.inserttag_button.set_tooltip_text(GnomeConfig.TAG_TOOLTIP)
-        self.open_parents_button = self.builder.get_object("open_parents")
-        self.open_parents_button.set_tooltip_text(
-            GnomeConfig.OPEN_PARENT_TOOLTIP)
+        # self.subtask_button.set_tooltip_text(GnomeConfig.SUBTASK_TOOLTIP)
```
izidormatusov: Can you set that tooltip from .ui instead and remove the code?

GTG/gtk/editor/editor.py
```diff
@@ -30,11 +30,12 @@
 from GTG.core.plugins.api import PluginAPI
 from GTG.core.plugins.engine import PluginEngine
 from GTG.core.task import Task
-from GTG.core.translations import _, ngettext
+# from GTG.core.translations import _, ngettext
```
izidormatusov: uncomment this line

GTG/gtk/editor/editor.py
```diff
@@ -30,11 +30,12 @@
...
 from GTG.gtk.editor import GnomeConfig
 from GTG.gtk.editor.calendar import GTGCalendar
 from GTG.gtk.editor.taskview import TaskView
-from GTG.gtk.help import add_help_shortcut
+# from GTG.gtk.help import add_help_shortcut
```
izidormatusov: uncomment this line

GTG/gtk/editor/editor.py
```diff
@@ -30,11 +30,12 @@
...
+from GTG.gtk.tag_completion import tag_filter
```
izidormatusov: Can you add todo next to this import saying you want to refactor this into a separate module?

GTG/gtk/editor/editor.py
```diff
@@ -52,6 +53,7 @@ def __init__(self,
         vmanager is the view manager
         thisisnew is True when a new task is created and opened
         '''
+        self.store = None
```
izidormatusov: You can initiate the store directly here (one less condition in the code bellow).

For extra points, declare the store directly in .ui file and then just do self.builder.get_object()

GTG/gtk/editor/editor.py
```diff
@@ -79,11 +81,12 @@ def __init__(self,
             "delete_clicked": self.delete_task,
             "on_duedate_pressed": lambda w: self.on_date_pressed(
                 w, GTGCalendar.DATE_KIND_DUE),
+            "on_tags_popover": self.open_tags_popover,
             "on_startdate_pressed": lambda w: self.on_date_pressed(
                 w, GTGCalendar.DATE_KIND_START),
             "on_closeddate_pressed": lambda w: self.on_date_pressed(
                 w, GTGCalendar.DATE_KIND_CLOSED),
-            "close_clicked": self.close,
+            # "close_clicked": self.close,
```
izidormatusov: Remove this. I can't find the reference to the original close_clicked signal anywhere

izidormatusov: Thanks to that, you can simplify the declaration of close method, to not accept any arguments

GTG/gtk/editor/editor.py
```diff
@@ -98,8 +101,11 @@ def __init__(self,
                 w, e, GTGCalendar.DATE_KIND_CLOSED),
             "on_insert_subtask_clicked": self.insert_subtask,
             "on_inserttag_clicked": self.inserttag_clicked,
-            "on_open_parent_clicked": self.open_parent_clicked,
+            "on_parent_select": self.open_parent_clicked,
```
izidormatusov: can you also rename the method to open_parent_select

GTG/gtk/editor/editor.py
```diff
@@ -184,46 +194,113 @@ def __init__(self,
...
+
+    #     # Ctrl-I marks task as dismissed
+    #     dismiss_editor = self.builder.get_object('dismiss_editor')
+    #     key, mod = Gtk.accelerator_parse('<Control>i')
+    #     dismiss_editor.add_accelerator('clicked', agr, key, mod,
+    #                                    Gtk.AccelFlags.VISIBLE)
+
+    #     # Ctrl-Q quits GTG
+    #     key, modifier = Gtk.accelerator_parse('<Control>q')
+    #     agr.connect(key, modifier, Gtk.AccelFlags.VISIBLE, self.quit)
```
izidormatusov: There is no reason to comment this method out. If nobody calls it, it isn't going to fail. Please add a todo above the method so we don't forget to *re-enable* it later.

GTG/gtk/editor/editor.py
```diff
@@ -184,46 +194,113 @@ def __init__(self,
...
+    TODO(jakubbrindza): Add the functionality to the existing calendar widgets.
+    This will require ammending and re-factoring the entire calendar.py.
+    '''
+
+    def show_popover_start(self, widget, popover):
+        self.popover = self.builder.get_object("date_popover")
+        self.popover.set_relative_to(self.builder.get_object
+                                     ("startdate_entry"))
+        self.popover.set_modal(False)
+        # popover.hide(self.builder.get_object("fuzzy_btns"))
```
izidormatusov: Why do you have this line commented out? and 3 times?

GTG/gtk/editor/editor.py
```diff
@@ -184,46 +194,113 @@ def __init__(self,
...
+
+    '''
+    TODO(jakubbrindza): Add the functionality to the existing calendar widgets.
+    This will require ammending and re-factoring the entire calendar.py.
+    '''
+
+    def show_popover_start(self, widget, popover):
+        self.popover = self.builder.get_object("date_popover")
+        self.popover.set_relative_to(self.builder.get_object
+                                     ("startdate_entry"))
```
izidormatusov: You can simplify the code:

```python
entry = self.builder.get_object("startdate_entry")
self.popover.set_relative_to(entry)
```

It's 2 lines instead of ~~1~~ 2 lines and it reads a bit nicer.

GTG/gtk/editor/editor.py
```diff
@@ -184,46 +194,113 @@ def __init__(self,
...
+    #     # Ctrl-Q quits GTG
+    #     key, modifier = Gtk.accelerator_parse('<Control>q')
+    #     agr.connect(key, modifier, Gtk.AccelFlags.VISIBLE, self.quit)
+
+    '''
+    TODO(jakubbrindza): Add the functionality to the existing calendar widgets.
+    This will require ammending and re-factoring the entire calendar.py.
+    '''
+
+    def show_popover_start(self, widget, popover):
```
izidormatusov: According to the [documentation](https://developer.gnome.org/gtk3/stable/GtkWidget.html#GtkWidget-focus-in-event), the 3 argument should be event, not popover.

GTG/gtk/editor/editor.py
```diff
@@ -184,46 +194,113 @@ def __init__(self,
...
+        self.popover.set_modal(False)
+        # popover.hide(self.builder.get_object("fuzzy_btns"))
+        self.popover.show_all()
+
+    def open_tags_popover(self, widget):
+        if self.store is None:
+            self.store = Gtk.ListStore(bool, str)
+            self.store.clear()
+
+            self.treeview = self.builder.get_object("tags_tree")
```
izidormatusov: You don't need to set it to the scope of class (it's used only in this method)

This is only an implementation detail, don't expose it to the outside world.

GTG/gtk/editor/editor.py
```diff
@@ -184,46 +194,113 @@ def __init__(self,
...
+
+            renderer_toggle = Gtk.CellRendererToggle()
+            renderer_toggle.connect("toggled", self.on_cell_toggled)
+            column_toggle = Gtk.TreeViewColumn("Toggle", renderer_toggle,
+                                               active=0)
+            self.treeview.append_column(column_toggle)
+
+            renderer_text = Gtk.CellRendererText()
+            column_text = Gtk.TreeViewColumn("Text", renderer_text, text=1)
+            self.treeview.append_column(column_text)
```
izidormatusov: Can you set those attributes in .ui file? It should be pretty easy to do and there will be less of python boilerplate code :)

GTG/gtk/editor/editor.py
```diff
@@ -184,46 +194,113 @@ def __init__(self,
...
+            self.treeview.set_model(self.store)
+            self.treeview.show_all()
+
+    def on_cell_toggled(self, widget, path):
+        self.store[path][0] = not self.store[path][0]
+
+        if self.store[path][0]:
+            self.inserttag(None, self.store[path][1])
+        else:
+            # self.remove_tag_edit(self.store[path][1])
```
izidormatusov: Can you add a TODO here and remove the call?

GTG/gtk/editor/editor.py
```diff
@@ -184,46 +194,113 @@ def __init__(self,
...
+
+            tags = self.req.get_tag_tree().get_all_nodes()
+
+            used_tags = self.task.get_tags()
+
+            for tagname in tags:
+                tag = self.req.get_tag(tagname)
+                if tag_filter(tag):
+                    is_used = tag in used_tags
+                    self.store.append([is_used, tagname])
```
izidormatusov: Can you add a TODO to sort the tags?

GTG/gtk/editor/editor.py
```diff
@@ -184,46 +194,113 @@ def __init__(self,
...
+
+            self.treeview.set_model(self.store)
+            self.treeview.show_all()
+
+    def on_cell_toggled(self, widget, path):
+        self.store[path][0] = not self.store[path][0]
```
izidormatusov: make the code a bit readable:

```python
tag_row = self.store[path]
tag_row[0] = not tag_row[0]
if tag_row[0]:
   self.inserttag(None, tag_row[1])
```

Also write a docstring explaining that tag_row is expected to be (is_used, tagname)

GTG/gtk/editor/editor.py
```diff
@@ -536,14 +614,28 @@ def inserttag_clicked(self, widget):
             self.textview.insert_text("@", itera)
         else:
             self.textview.insert_text(" @", itera)
-        self.textview.grab_focus()
+        # self.textview.grab_focus()
```
izidormatusov: Please, remove this method, as you don't use it any more (you have the new fancy popoup don't you?)

GTG/gtk/editor/editor.py
```diff
@@ -536,14 +614,28 @@ def inserttag_clicked(self, widget):
...
 
     def inserttag(self, widget, tag):
         self.textview.insert_tags([tag])
-        self.textview.grab_focus()
+        # self.textview.grab_focus()
```
izidormatusov: Remove this method and inline the call directly in handling cell_toggle

GTG/gtk/editor/editor.py
```diff
@@ -536,14 +614,28 @@ def inserttag_clicked(self, widget):
...
+
+    def remove_tag_edit(self, widget, tagname):
```
izidormatusov: I don't think it is worth to hide the single call in a special method. Remove it, please.

GTG/gtk/editor/editor.py
```diff
@@ -536,14 +614,28 @@ def inserttag_clicked(self, widget):
...
+        '''
+        TODO(jakubbrindza): add this functionality to taskview.py.
+        It is crucial for GTG 0.4!!!
+        '''
+        self.textview.remove_tag_editor(tagname)
 
     def open_parent_clicked(self, widget):
-        self.vmanager.open_task(self.task.get_parents()[0])
+        if len(self.task.get_parents()) == 1:
```
izidormatusov: Simplify the code a bit:

```python
parents = self.task.get_parents()
if len(parents) == 1:
  self.vmanager.open_task(parents[0])
else:
  # ...
```

GTG/gtk/editor/editor.py
```diff
@@ -536,14 +614,28 @@ def inserttag_clicked(self, widget):
...
+            self.vmanager.open_task(self.task.get_parents()[0])
+        elif len(self.task.get_parents()) > 1:
+            pass
+        '''
+        TODO (jakubbrindza): If there is more than 1 parent,open a popover with
+        their names and allow choosing which one to open.
+        '''
```
izidormatusov: Can you put this todo above pass? The cocde should go there :)

GTG/gtk/editor/taskview.py
```diff
@@ -463,6 +463,37 @@ def insert_tag(self, tag):
         line_mark = self.buff.create_mark("lastline", lastline, False)
         self.insert_at_mark(self.buff, line_mark, tag)
 
+    def remove_tag_editor(self, tagname):
```
izidormatusov: Choose a better, consistent name. When the method to add tags is `insert_tags`, the method to remove tags should be called `remove_tags` (and accept multiple tags at the same time)

GTG/gtk/editor/taskview.py
```diff
@@ -463,6 +463,37 @@ def insert_tag(self, tag):
...
+        enable removing a tag from a task in task editor.
+        '''
+        self.buff = self.get_buffer()
+        mark_start = self.buff.get_mark(tagname)
+        mark_end = self.buff.get_mark("/%s" + tagname)
+
+        it = self.buff.get_iter_at_mark(mark_start)
+        for tag in it.get_toggled_tags():
+            if hasattr('is_tag', tag) and tag.tagname == tagname:
+                        if mark_start:
```
izidormatusov: This has too deep indentation (maybe 8 spaces instead of 4?)

GTG/gtk/editor/taskview.py
```diff
@@ -463,6 +463,37 @@ def insert_tag(self, tag):
         line_mark = self.buff.create_mark("lastline", lastline, False)
         self.insert_at_mark(self.buff, line_mark, tag)
 
+    def remove_tag_editor(self, tagname):
+        '''
+        TODO(jakubbrindza): Finish works on this function which will
+        enable removing a tag from a task in task editor.
+        '''
+        self.buff = self.get_buffer()
```
izidormatusov: AFAIK, self.buff is already set at this point

GTG/gtk/editor/taskview.py
```diff
@@ -463,6 +463,37 @@ def insert_tag(self, tag):
...
+        mark_end = self.buff.get_mark("/%s" + tagname)
+
+        it = self.buff.get_iter_at_mark(mark_start)
+        for tag in it.get_toggled_tags():
+            if hasattr('is_tag', tag) and tag.tagname == tagname:
+                        if mark_start:
+                            offset1 = self.buff.get_iter_at_mark(mark_start)\
+                                            .get_offset()
+                            if mark_start.get_offset() <= offset1 <= \
+                                    mark_end.get_offset():
```
izidormatusov: This looks strange. If you set offset from mark_start, then it should be logical that mark_start.get_offset() == offset1

I think you can get rid of the condition.

GTG/gtk/editor/taskview.py
```diff
@@ -463,6 +463,37 @@ def insert_tag(self, tag):
...
+                            offset2 = self.buff.get_iter_at_mark(mark_end)\
+                                            .get_offset()
+                            if mark_start.get_offset() <= offset2 <= \
+                                    mark_end.get_offset():
+                                self.buff.delete_mark_by_name("/%s" % tagname)
+
+        it_end = self.buff.get_iter_at_mark(mark_end)
+        self.buff.delete_text(it, it_end)
+        self.buff.delete_mark(mark_start)
+        self.buff.delete_mark(mark_end)
```
izidormatusov: You are deleting marks 2 times. You don't need that.

GTG/gtk/editor/taskview.py
```diff
@@ -463,6 +463,37 @@ def insert_tag(self, tag):
...
+        '''
+        TODO(jakubbrindza): Finish works on this function which will
+        enable removing a tag from a task in task editor.
+        '''
+        self.buff = self.get_buffer()
+        mark_start = self.buff.get_mark(tagname)
+        mark_end = self.buff.get_mark("/%s" + tagname)
+
+        it = self.buff.get_iter_at_mark(mark_start)
+        for tag in it.get_toggled_tags():
```
izidormatusov: You might not need the whole for loop as you want to delete that tag anyway.

GTG/gtk/editor/taskview.py
```diff
@@ -463,6 +463,37 @@ def insert_tag(self, tag):
...
+                                self.buff.delete_mark_by_name(tagname)
+                        if mark_end:
+                            offset2 = self.buff.get_iter_at_mark(mark_end)\
+                                            .get_offset()
+                            if mark_start.get_offset() <= offset2 <= \
+                                    mark_end.get_offset():
+                                self.buff.delete_mark_by_name("/%s" % tagname)
+
+        it_end = self.buff.get_iter_at_mark(mark_end)
+        self.buff.delete_text(it, it_end)
```
izidormatusov: This method doesn't exist. However, there is [delete](http://lazka.github.io/pgi-docs/Gtk-3.0/classes/TextBuffer.html#Gtk.TextBuffer.delete) method.

GTG/gtk/editor/taskview.py
```diff
@@ -463,6 +463,37 @@ def insert_tag(self, tag):
...
+        self.buff.delete_mark(mark_start)
+        self.buff.delete_mark(mark_end)
+        self.remove_tag_callback()
```
izidormatusov: Remove tag callback requires the tagname as an argument.

GTG/gtk/editor/taskview.py
```diff
@@ -463,6 +463,37 @@ def insert_tag(self, tag):
         line_mark = self.buff.create_mark("lastline", lastline, False)
         self.insert_at_mark(self.buff, line_mark, tag)
 
+    def remove_tag_editor(self, tagname):
+        '''
+        TODO(jakubbrindza): Finish works on this function which will
+        enable removing a tag from a task in task editor.
```
izidormatusov: I think you shouldn't include this function in this PR as it is not ready and it is not called anyway.

GTG/gtk/editor/editor.py
```diff
@@ -164,7 +174,7 @@ def __init__(self,
             self.task.set_to_keep()
         self.textview.modified(full=True)
         self.window.connect("destroy", self.destruction)
-        self.calendar.connect("date-changed", self.on_date_changed)
+        # self.calendar.connect("date-changed", self.on_date_changed)
```
izidormatusov: Can you add todo for this? Either provide the functionality or drop it completely?

GTG/gtk/editor/editor.py
```diff
@@ -184,46 +194,113 @@ def __init__(self,
...
+            for tagname in tags:
+                tag = self.req.get_tag(tagname)
+                if tag_filter(tag):
+                    is_used = tag in used_tags
+                    self.store.append([is_used, tagname])
+
+            self.treeview.set_model(self.store)
+            self.treeview.show_all()
+
+    def on_cell_toggled(self, widget, path):
```
izidormatusov: can you name this method better? Something like `on_tag_toggled`? There could be potentially many cells here.

GTG/gtk/editor/editor.py
```diff
@@ -333,40 +403,40 @@ def refresh_editor(self, title=None, refreshtext=False):
...
+        #         txt = ""
+        #     elif result > 0:
+        #         txt=ngettext("Due tomorrow!", "%(days)d days left", result) \
+        #             % {'days': result}
+        #     elif result == 0:
+        #         txt = _("Due today!")
+        #     elif result < 0:
+        #         abs_result = abs(result)
+        #         txt = ngettext("Due yesterday!", "Was %(days)d days ago",
+        #                        abs_result) % {'days': abs_result}
```
izidormatusov: Can you refactor the computation of this label in a separate function? It should have been there in the first place. You don't have to call that function, just put a commented out call in here.

GTG/gtk/editor/editor.py
```diff
@@ -381,22 +451,22 @@ def refresh_editor(self, title=None, refreshtext=False):
                 mi.connect("activate", self.inserttag, tagname)
                 mi.show()
                 menu.append(mi)
-        if tag_count > 0:
-            self.inserttag_button.set_menu(menu)
+        # if tag_count > 0:
+        #      self.inserttag_button.set_menu(menu)
```
izidormatusov: I don't think this code will be needed, you will show the popover menu instead anyway

GTG/gtk/editor/editor.py
```diff
@@ -381,22 +451,22 @@ def refresh_editor(self, title=None, refreshtext=False):
...
+        # if len(parents) > 0:
+        #     for parent in self.task.get_parents():
+        #         task = self.req.get_task(parent)
+        #        mi = Gtk.MenuItem(label=task.get_title(), use_underline=False)
+        #         mi.connect("activate", self.open_parent, parent)
+        #         mi.show()
+        #         menu.append(mi)
+        #     self.open_parents_button.set_menu(menu)
+        # else:
+        #     self.open_parents_button.set_sensitive(False)
```
izidormatusov: I think this will be handled by completely different code, remove this code :)

GTG/gtk/editor/editor.py
```diff
@@ -505,7 +583,7 @@ def change_status(self, widget):
     def delete_task(self, widget):
         # this triggers the closing of the window in the view manager
         if self.task.is_new():
-            # self.req.delete_task(self.task.get_id())
+            self.req.delete_task(self.task.get_id())
```
izidormatusov: Could you provide a reason why you uncommented this line?

GTG/gtk/editor/taskeditor.ui
```diff
@@ -4,403 +4,461 @@
...
+    <child type="titlebar">
+      <object class="GtkHeaderBar" id="headerbar">
         <property name="visible">True</property>
-        <property name="can_focus">False</property>
-        <property name="orientation">vertical</property>
+        <property name="show_close_button">True</property>
+        <property name="title" translatable="yes">Task</property>
         <child>
-          <object class="GtkToolbar" id="task_tb1">
+          <object class="GtkMenuButton" id="mark_as_done_editor">
```
izidormatusov: why not `mark_as_done_button` or `mark_as_done` ? No need to put there `_editor` suffix

GTG/gtk/editor/taskeditor.ui
```diff
@@ -4,403 +4,461 @@
...
-              <packing>
-                <property name="expand">False</property>
-                <property name="homogeneous">True</property>
-              </packing>
-            </child>
+            <property name="sensitive">True</property>
+            <property name="label" translatable="yes">Mark as Done</property>
+            <property name="tooltip_text" translatable="yes">Mark the task as done</property>
+            <property name="valign">center</property>
+            <signal handler="mark_as_done_clicked" name="clicked" swapped="no"/>
```
izidormatusov: Can you follow the convention to name signals `on_$something`? E.g. `on_mark_as_done_clicked`

GTG/gtk/editor/taskeditor.ui
```diff
@@ -4,403 +4,461 @@
...
         </child>
         <child>
-          <object class="GtkBox" id="box3">
+          <object class="GtkMenuButton" id="editor_menu">
             <property name="visible">True</property>
-            <property name="can_focus">False</property>
+            <property name="sensitive">True</property>
+            <property name="tooltip_text" translatable="yes">Task action menu</property>
+            <property name="valign">center</property>
+            <property name="popover">menu_popover</property>
```
izidormatusov: it might be better to name it `editor_menu_popover` as you will have more popover menus

GTG/gtk/editor/taskeditor.ui
```diff
@@ -4,403 +4,461 @@
...
+            <property name="pack_type">end</property>
           </packing>
+          <style>
+            <class name="image-button"/>
+          </style>
         </child>
+      </object>
+    </child>
+    <child>
+      <object class="GtkBox" id="vbox4">
```
izidormatusov: What is `vbox4`? Provide a more descriptive name, please

GTG/gtk/editor/taskeditor.ui
```diff
@@ -4,403 +4,461 @@
...
-              <packing>
-                <property name="expand">False</property>
-                <property name="fill">True</property>
-                <property name="padding">10</property>
-                <property name="position">0</property>
-              </packing>
-            </child>
             <child>
-              <object class="GtkBox" id="box1">
+              <object class="GtkBox" id="task_toolbar_box">
```
izidormatusov: You might want to group semantically similar buttons into button boxes: https://developer.gnome.org/gtk3/stable/GtkButtonBox.html

It creates the effect of button groups, like in this screenshot (list icons are separate from formatting icons):
x

GTG/gtk/editor/taskeditor.ui
```diff
@@ -4,403 +4,461 @@
...
+                  <object class="GtkButton" id="parent">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
-                    <property name="invisible_char">●</property>
-                    <property name="width_chars">10</property>
-                    <signal name="changed" handler="closeddate_changed" swapped="no"/>
-                    <signal name="focus-out-event" handler="closeddate_focus_out" swapped="no"/>
+                    <property name="valign">center</property>
+                    <property name="tooltip_text">Open the parent task</property>
+                    <!--<property name="popover">parent_popover</property>-->
```
izidormatusov: Remove this property and set it from the code when needed.

GTG/gtk/editor/taskeditor.ui
```diff
@@ -4,403 +4,461 @@
...
                 <property name="expand">False</property>
                 <property name="fill">True</property>
-                <property name="position">5</property>
+                <property name="position">0</property>
               </packing>
             </child>
+          </object>
+        </child>
+        <child>
+          <object class="GtkBox" id="box3">
```
izidormatusov: What is box3? Provide a better name. Do you even need this box?

GTG/gtk/editor/taskeditor.ui
```diff
@@ -4,403 +4,461 @@
...
+  <object class="GtkPopoverMenu" id="parent_popover">
+    <child>
+      <object class="GtkModelButton" id="parent_button">
+        <property name="visible">True</property>
+        <property name="centered">True</property>
+        <property name="text" translatable="yes">Parent Button</property>
+      </object>
+    </child>
+  </object>
+  <object class="GtkPopoverMenu" id="start_date_popover">
```
izidormatusov: This object isn't used anywhere.

GTG/gtk/editor/taskeditor.ui
```diff
@@ -4,403 +4,461 @@
...
+            <property name="expand">True</property>
             <property name="fill">True</property>
-            <property name="position">3</property>
+            <property name="position">1</property>
           </packing>
         </child>
       </object>
     </child>
   </object>
+  <object class="GtkPopoverMenu" id="parent_popover">
```
izidormatusov: this object isn't used anywhere

GTG/gtk/editor/taskeditor.ui
```diff
@@ -4,403 +4,461 @@
...
+        <child>
+          <object class="GtkModelButton" id="dismiss_editor">
+            <property name="visible">True</property>
+            <property name="centered">True</property>
+            <property name="text" translatable="yes">Dismiss Task</property>
+            <signal handler="on_dismiss" name="clicked" swapped="no"/>
+          </object>
+        </child>
+        <child>
+          <object class="GtkSeparatorMenuItem" id="separator4">
```
izidormatusov: you don't need id for separator, at least not `separator4`

tkdchen:
Like this new UI.

A star image is usually used for bookmark or giving a rate to something. In GTG, '@' is the prefix of a tag, will '@' and star together confuse users?

izidormatusov:
@tkdchen that's a good idea! @jakubbrindza can you make it happen?

jakubbrindza:
@tkdchen, @izidormatusov that's certainly a great point and I've spent quite a lot of time trying to find the best fit for this particular icon, however working with the best source of Gtk icon names I'm aware of: http://standards.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html I didn't find anything better that would be near-tag (maybe I'm partially blind :smile: ).

I'll have another look for sure, or try to come up with a custom icon :blush:

GTG/gtk/editor/editor.py
jakubbrindza: done

GTG/gtk/editor/editor.py
jakubbrindza: done

GTG/gtk/editor/editor.py
jakubbrindza: good note! done.

GTG/gtk/editor/editor.py
jakubbrindza: done

GTG/gtk/editor/editor.py
jakubbrindza: Great, I removed it.
However, can you explain the latter comment? What can be simplified?

GTG/gtk/editor/taskeditor.ui
jakubbrindza: Sure thing, the reason for bigger size was that I considered the view less cluttered that way, however 500x400 does not affect functionality or visibility of widgets, so we can live with that :)

GTG/gtk/editor/taskeditor.ui
jakubbrindza: I wanted to distinguish between browser & editor mark as done / dismiss / delete buttons, but you're right, there's no need to do that. Thanks

GTG/gtk/editor/taskeditor.ui
jakubbrindza: that will make the code simpler to use, good! Done.

GTG/gtk/editor/taskeditor.ui
jakubbrindza: Done

GTG/gtk/editor/taskeditor.ui
jakubbrindza: :+1:

GTG/gtk/editor/taskeditor.ui
jakubbrindza: :+1:

GTG/gtk/editor/taskeditor.ui
jakubbrindza: It was useless, I removed it.

GTG/gtk/editor/taskeditor.ui
jakubbrindza: :+1:

GTG/gtk/editor/taskview.py
jakubbrindza: but then the entire functionality of tags will not be complete. Is that an acceptable way to handle this? I think it will only raise tons of complaints and bugs from the users.

However if you think that this is extra feature (which it is) and would deserve a separate PR then I cannot oppose and I remove the entire body of the function for now :blush:

GTG/gtk/editor/taskview.py
jakubbrindza: well I want to delete both, no? Therefore I'm doing it twice...

GTG/gtk/editor/taskview.py
jakubbrindza: :+1:

GTG/gtk/editor/taskview.py
jakubbrindza: :+1:

GTG/gtk/editor/taskview.py
jakubbrindza: good catch :+1:

GTG/gtk/editor/taskview.py
jakubbrindza: good for consistency, thanks! :+1:

GTG/gtk/editor/taskeditor.ui
jakubbrindza: :+1:  removed

GTG/gtk/editor/taskeditor.ui
jakubbrindza: :+1:  removed

GTG/gtk/editor/editor.py
jakubbrindza: However this is the reason for these things being commented out:
Since some of the core functions using these things are commented out / not in the code probably anymore, tests will fail and that's not what we want

screen shot 2015-07-01 at 11 16 06

GTG/gtk/editor/editor.py
jakubbrindza: I don't really know if that's a good idea. The reason for having it like this is that later in the code, we are changing the tooltips based on whether your task is among active, dismissed or marked as done tasks, you know... and therefore it might as well be a matter of convenience to have it like this, done in one way. 

What do you say?

GTG/gtk/editor/editor.py
jakubbrindza: :+1:  adding it to TODO

GTG/gtk/editor/editor.py
jakubbrindza: :+1:  thanks

GTG/gtk/editor/editor.py
jakubbrindza: :+1:

GTG/gtk/editor/editor.py
jakubbrindza: :+1:  nice!

GTG/gtk/editor/editor.py
jakubbrindza: Sorry, left-overs from me playing around with hiding some of the buttons in the start/due popovers. I'll implement it more nicely

GTG/gtk/editor/editor.py
jakubbrindza: :+1:

GTG/gtk/editor/editor.py
jakubbrindza: :+1:

GTG/gtk/editor/editor.py
jakubbrindza: :+1:

GTG/gtk/editor/editor.py
jakubbrindza: :+1:

GTG/gtk/editor/editor.py
jakubbrindza: :+1:

GTG/gtk/editor/editor.py
jakubbrindza: :+1:

GTG/gtk/editor/editor.py
jakubbrindza: :+1:  yup, you're right here too

GTG/gtk/editor/editor.py
jakubbrindza: to be honest? By a mistake :smiley_cat: I just noticed it and can't find a reason for me doing that (or for someone else leaving it commented out!)

Hmm, apparently it was not needed before

GTG/gtk/editor/editor.py
jakubbrindza: It's used twice (it'll be used also in def refresh_editor), so I'm leaving it as a separate method for now

GTG/gtk/editor/editor.py
jakubbrindza: removed

GTG/gtk/editor/editor.py
jakubbrindza: :+1:

GTG/gtk/editor/editor.py
jakubbrindza: Sorry for that, resolved :blush:

GTG/gtk/editor/editor.py
jakubbrindza: :+1:

GTG/gtk/editor/editor.py
jakubbrindza: One thing, I think it could be a bit complicated. 
Secondly, this way we know exactly what happens from one place which is a positive thing.

GTG/gtk/editor/editor.py
jakubbrindza: sure! :+1:

GTG/gtk/editor/taskeditor.ui
jakubbrindza: This is a great idea!
However, there is a problem with the default padding property for child elements in GtkButtonBox (yes, I'm convinced the problem is there, I've spent a lot of time on this and I cannot find any other possible explanation) which causes the child buttons to be separated and not nicely aligned.

If you have an idea, I'll be glad to implement it, but for now, child-internal-pad-x is a property which I cannot implement and re-set to the value of 0, which would solve our problem.

jakubbrindza:
@izidormatusov I tried to address all comments possible, please review when you have some free time :blush: 

(I'll just try to come up with a good tag icon replacement)

GTG/gtk/editor/editor.py
jakubbrindza: I'll think about this one...

izidormatusov: You doing neither of the proposed alternatives. Please, define the store directly in .ui file

GTG/gtk/editor/editor.py
izidormatusov: subtask button doesn't change it's tooltip. Delete menu item doesn't change it's tooltip. They should be defined in .ui file

I'm also of the opinion, that there should be multiple buttons (mark as done, mark as undone). You should not modify those buttons, just choose which one you decide to choose. Put the description how the element should look like into .ui file and the logic of when to show what in python. You can do it very easily:

```python
if condition:
  button_a.show()
  button_b.hide()
else:
  button_a.hide()
  button_b.show()
```

The same thing for menu items

GTG/gtk/editor/__init__.py
```diff
@@ -31,8 +31,8 @@ class GnomeConfig:
 
     MARK_DONE = _("Mark as Done")
     MARK_UNDONE = _("Mark as not Done")
-    MARK_DISMISS = _("Dismiss")
-    MARK_UNDISMISS = _("Undismiss")
+    MARK_DISMISS = _("Dismiss Task")
+    MARK_UNDISMISS = _("Undismiss Task")
 
     MARK_DONE_TOOLTIP = _("Mark this task as done")
     MARK_UNDONE_TOOLTIP = _("Mark this task as to be done")
```
izidormatusov: Those should be in .ui file

GTG/gtk/editor/__init__.py
```diff
@@ -31,8 +31,8 @@ class GnomeConfig:
 
     MARK_DONE = _("Mark as Done")
     MARK_UNDONE = _("Mark as not Done")
-    MARK_DISMISS = _("Dismiss")
-    MARK_UNDISMISS = _("Undismiss")
+    MARK_DISMISS = _("Dismiss Task")
+    MARK_UNDISMISS = _("Undismiss Task")
```
izidormatusov: Those should be in .ui file

GTG/gtk/editor/editor.py
```diff
@@ -60,30 +65,30 @@ def __init__(self,
...
         self.subtask_button = self.builder.get_object("insert_subtask")
-        self.subtask_button.set_tooltip_text(GnomeConfig.SUBTASK_TOOLTIP)
-        self.inserttag_button = self.builder.get_object("inserttag")
-        self.inserttag_button.set_tooltip_text(GnomeConfig.TAG_TOOLTIP)
-        self.open_parents_button = self.builder.get_object("open_parents")
-        self.open_parents_button.set_tooltip_text(
-            GnomeConfig.OPEN_PARENT_TOOLTIP)
+        # self.subtask_button.set_tooltip_text(GnomeConfig.SUBTASK_TOOLTIP)
+        self.inserttag_button = self.builder.get_object("tags")
+        # self.inserttag_button.set_tooltip_text(GnomeConfig.TAG_TOOLTIP)
```
izidormatusov: put this thing into .ui file and delete this line

GTG/gtk/editor/editor.py
```diff
@@ -60,30 +65,30 @@ def __init__(self,
...
+        self.open_parents_button = self.builder.get_object("parent")
+        # self.open_parents_button.set_tooltip_text(
+        #      GnomeConfig.OPEN_PARENT_TOOLTIP)
```
izidormatusov: put this into .ui file and delete this line

GTG/gtk/editor/editor.py
```diff
@@ -536,14 +609,20 @@ def inserttag_clicked(self, widget):
             self.textview.insert_text("@", itera)
         else:
             self.textview.insert_text(" @", itera)
-        self.textview.grab_focus()
 
     def inserttag(self, widget, tag):
         self.textview.insert_tags([tag])
```
izidormatusov: Remove this function and inline the call directly in the popover callback

GTG/gtk/editor/__init__.py
jakubbrindza: @izidormatusov well, shouldn't all of them be placed to a .ui file?

I mean, I can create, as you suggest a greater variety of unused (property visible == False) widgets which will be inter-changed with those default ones and set all the tooltip properties automatically there and all of these can go away, right?

GTG/gtk/editor/editor.py
```diff
@@ -330,9 +415,32 @@ def refresh_editor(self, title=None, refreshtext=False):
             self.closeddate_widget.set_text(str(closeddate))
 
         # refreshing the day left label
+        # self.refresh_day_left()
```
izidormatusov: Add todo to re-enable this

GTG/gtk/editor/editor.py
```diff
@@ -330,9 +415,32 @@ def refresh_editor(self, title=None, refreshtext=False):
...
+        tag_count = 0
+        for tagname in taglist:
+            tag_object = self.req.get_tag(tagname)
+            if not tag_object.is_special() and \
+                    not self.task.has_tags(tag_list=[tagname]):
+                tag_count += 1
+                mi = Gtk.MenuItem(label=tagname, use_underline=False)
+                mi.connect("activate", self.inserttag, tagname)
+                mi.show()
+                menu.append(mi)
```
izidormatusov: You don't need this block as you use something completely else now. Drop it

GTG/gtk/editor/editor.py
```diff
@@ -330,9 +415,32 @@ def refresh_editor(self, title=None, refreshtext=False):
...
+
+        if refreshtext:
+            self.textview.modified(refresheditor=False)
+        if to_save:
+            self.light_save()
+
+    def refresh_day_left(self, title=None, refreshtext=False):
```
izidormatusov: You don't need attributes title and refreshtext, drop them!

GTG/gtk/editor/editor.py
izidormatusov: If it wasn't needed before, you shouldn't need it now. Delete the line

GTG/gtk/editor/editor.py
```diff
@@ -60,30 +65,30 @@ def __init__(self,
...
         self.deletebutton.set_tooltip_text(GnomeConfig.DELETE_TOOLTIP)
         self.subtask_button = self.builder.get_object("insert_subtask")
-        self.subtask_button.set_tooltip_text(GnomeConfig.SUBTASK_TOOLTIP)
-        self.inserttag_button = self.builder.get_object("inserttag")
-        self.inserttag_button.set_tooltip_text(GnomeConfig.TAG_TOOLTIP)
-        self.open_parents_button = self.builder.get_object("open_parents")
-        self.open_parents_button.set_tooltip_text(
-            GnomeConfig.OPEN_PARENT_TOOLTIP)
+        # self.subtask_button.set_tooltip_text(GnomeConfig.SUBTASK_TOOLTIP)
+        self.inserttag_button = self.builder.get_object("tags")
```
izidormatusov: as you set the tooltip in .ui file, you don't need the reference to this object any more. Drop the line! :fire:

GTG/gtk/editor/editor.py
```diff
@@ -252,6 +344,10 @@ def init_dimensions(self):
     # Refreshtext is whether or not we should refresh the TaskView
     # (doing it all the time is dangerous if the task is empty)
     def refresh_editor(self, title=None, refreshtext=False):
+        '''
+        TODO(jakubbrindza): Repair refresh functionality for tags and parents
+        (a.k.a. adjust the commented out code for current needs)
+        '''
```
izidormatusov: This is not true, and you already solved this by deleting them! Kill it with :fire:!

GTG/gtk/editor/editor.py
```diff
@@ -60,30 +65,30 @@ def __init__(self,
         self.clipboard = clipboard
         self.builder = Gtk.Builder()
```
izidormatusov: builder should not be accessed from anywhere else tan this method. Create reference to all buttons that you might need later, connect to all signals. Nothing else should be needed!

izidormatusov: Taking this back, that breaks the assumptions from plugins API quite horribly :/

GTG/gtk/editor/taskview.py
izidormatusov: Currently you can't remove tag from the toolbar, only to add it. After this big editor PR, I'm expecting that you do N small ones, each addressing specific issue (go to parent popover, tag management, start date, due date)

GTG/gtk/editor/__init__.py
izidormatusov: Yep

GTG/gtk/editor/editor.py
izidormatusov: Not really. *Business logic* shouldn't depend on how the information is presented but what you want to achieve (show me tags I can remove/add to this tag).

Have a look at gedit implementation in .ui file, it is easy:

 - [column definition](https://github.com/GNOME/gedit/blob/4378e052141fda0b7f567a9768e39782352437a6/gedit/resources/ui/gedit-encodings-dialog.ui#L128-L161)
 - [list store definition](https://github.com/GNOME/gedit/blob/4378e052141fda0b7f567a9768e39782352437a6/gedit/resources/ui/gedit-encodings-dialog.ui#L18-L27)

GTG/gtk/editor/editor.py
izidormatusov: No, it is used only once (I asked you remove that block of code from refresh_editor that calls it, as you don't need it any more)

GTG/gtk/editor/editor.py
izidormatusov: I believe this was resolved :smile:

GTG/gtk/editor/taskeditor.ui
izidormatusov: Good news, bad news! Good news, I found a solution for this issue. Put those buttons in a separate GtkBox and assign it style `linked`:

```xml
<object class="GtkBox">
  <style>
    <class name="linked"/>
  </style>
  <child>
    <object class="GtkButton" id="bold">
```

Found this in [gnome-builder ui](https://github.com/chergert/gnome-builder/blob/master/data/ui/gb-view-stack.ui#L192-L199)

Result:
button groups

This is second time I've burned myself on Gtk styles :-1: I found there is a list of all styles in [GtkStyleContext](https://developer.gnome.org/gtk3/stable/GtkStyleContext.html)

Bad news: it doesn't have any common with GtkButtonBox. Sorry for leading you onto a wrong path to solve this :cold_sweat:

GTG/gtk/editor/taskeditor.ui
```diff
@@ -7,400 +7,419 @@
     <property name="default_width">500</property>
```
izidormatusov: As we talked about it, please move this file to GTG/gtk/ui/taskeditor.ui

GTG/gtk/editor/editor.py
jakubbrindza: solved!

GTG/gtk/editor/editor.py
jakubbrindza: Removed

GTG/gtk/editor/editor.py
jakubbrindza: Removed

GTG/gtk/editor/editor.py
jakubbrindza: Dropped!

GTG/gtk/editor/editor.py
jakubbrindza: Did so!

GTG/gtk/editor/editor.py
jakubbrindza: Cool, done!

GTG/gtk/editor/editor.py
jakubbrindza: Yes, it was :blush:

GTG/gtk/editor/editor.py
```diff
@@ -60,30 +65,27 @@ def __init__(self,
...
+        self.undismissbutton = self.builder.get_object("undismiss")
+        self.deletebutton = self.builder.get_object("delete")
         self.subtask_button = self.builder.get_object("insert_subtask")
-        self.subtask_button.set_tooltip_text(GnomeConfig.SUBTASK_TOOLTIP)
-        self.inserttag_button = self.builder.get_object("inserttag")
-        self.inserttag_button.set_tooltip_text(GnomeConfig.TAG_TOOLTIP)
-        self.open_parents_button = self.builder.get_object("open_parents")
-        self.open_parents_button.set_tooltip_text(
-            GnomeConfig.OPEN_PARENT_TOOLTIP)
+        self.inserttag_button = self.builder.get_object("tags")
```
izidormatusov: You don't need this button anywhere, you can delete this line

GTG/gtk/editor/editor.py
```diff
@@ -60,30 +65,27 @@ def __init__(self,
...
+        self.open_parents_button = self.builder.get_object("parent")
```
izidormatusov: you don't need this button anywhere, delete this line

GTG/gtk/editor/editor.py
```diff
@@ -60,30 +65,27 @@ def __init__(self,
...
-        self.donebutton = self.builder.get_object("mark_as_done_editor")
-        self.dismissbutton = self.builder.get_object("dismiss_editor")
-        self.deletebutton = self.builder.get_object("delete_editor")
-        self.deletebutton.set_tooltip_text(GnomeConfig.DELETE_TOOLTIP)
+        self.donebutton = self.builder.get_object("mark_as_done")
+        self.undonebutton = self.builder.get_object("mark_as_undone")
+        self.dismissbutton = self.builder.get_object("dismiss")
+        self.undismissbutton = self.builder.get_object("undismiss")
+        self.deletebutton = self.builder.get_object("delete")
         self.subtask_button = self.builder.get_object("insert_subtask")
```
izidormatusov: You don't reference this button anywhere, drop it

GTG/gtk/editor/editor.py
```diff
@@ -60,30 +65,27 @@ def __init__(self,
...
         self.builder.add_from_file(GnomeConfig.EDITOR_UI_FILE)
-        self.donebutton = self.builder.get_object("mark_as_done_editor")
-        self.dismissbutton = self.builder.get_object("dismiss_editor")
-        self.deletebutton = self.builder.get_object("delete_editor")
-        self.deletebutton.set_tooltip_text(GnomeConfig.DELETE_TOOLTIP)
+        self.donebutton = self.builder.get_object("mark_as_done")
+        self.undonebutton = self.builder.get_object("mark_as_undone")
+        self.dismissbutton = self.builder.get_object("dismiss")
+        self.undismissbutton = self.builder.get_object("undismiss")
+        self.deletebutton = self.builder.get_object("delete")
```
izidormatusov: You don't reference this button anywhere, drop it

GTG/gtk/editor/editor.py
```diff
@@ -164,7 +173,11 @@ def __init__(self,
             self.task.set_to_keep()
         self.textview.modified(full=True)
         self.window.connect("destroy", self.destruction)
-        self.calendar.connect("date-changed", self.on_date_changed)
+        # self.calendar.connect("date-changed", self.on_date_changed)
+        '''
+        TODO(jakubbrindza): make this function work alongside
+        the new popover calendar
+        '''
```
izidormatusov: move todo above the line that todo references, please

GTG/gtk/editor/editor.py
```diff
@@ -204,19 +221,19 @@ def init_accelerators(self):
         agr.connect(key, modifier, Gtk.AccelFlags.VISIBLE, self.new_task)
 
         # Ctrl-Shift-N creates a new subtask
-        insert_subtask = self.builder.get_object("insert_subtask")
+        insert_subtask = self.builder.get_object("add_subtask")
         key, mod = Gtk.accelerator_parse("<Control><Shift>n")
         insert_subtask.add_accelerator('clicked', agr, key, mod,
```
izidormatusov: Please reuse the references from the above instead. (This means that some can't be removed because they will be used)

```
# Ctrl-Shift-N creates a new subtask
key, mod = Gtk.accelerator_parse("<Control><Shift>n")
self.insert_subtask.add_accelerator('clicked', agr, key, mod,
                                                Gtk.AccelFlags.VISIBLE)
```

GTG/gtk/editor/editor.py
```diff
@@ -204,19 +221,19 @@ def init_accelerators(self):
...
                                        Gtk.AccelFlags.VISIBLE)
 
         # Ctrl-D marks task as done
-        mark_as_done_editor = self.builder.get_object('mark_as_done_editor')
+        mark_as_done_editor = self.builder.get_object('mark_as_done')
```
izidormatusov: some here, use reference to self.donebutton

GTG/gtk/editor/editor.py
```diff
@@ -204,19 +221,19 @@ def init_accelerators(self):
...
         key, mod = Gtk.accelerator_parse('<Control>d')
         mark_as_done_editor.add_accelerator('clicked', agr, key, mod,
                                             Gtk.AccelFlags.VISIBLE)
 
         # Ctrl-I marks task as dismissed
-        dismiss_editor = self.builder.get_object('dismiss_editor')
+        dismiss_editor = self.builder.get_object('dismiss')
```
izidormatusov: self.dismiss_button

GTG/gtk/editor/editor.py
```diff
@@ -225,6 +242,78 @@ def init_accelerators(self):
...
+    TODO(jakubbrindza): Add the functionality to the existing calendar widgets.
+    This will require ammending and re-factoring the entire calendar.py.
+    '''
+
+    def show_popover_start(self, widget, event):
+        entry = self.builder.get_object("startdate_entry")
+        self.popover = self.builder.get_object("date_popover")
+        self.popover.set_relative_to(entry)
+        self.popover.set_modal(False)
+        self.popover.show_all()
```
izidormatusov: popover is not referenced from anywhere else than just this function. Make it local

```
popover = ...
```

GTG/gtk/editor/editor.py
```diff
@@ -225,6 +242,78 @@ def init_accelerators(self):
...
+
+    def show_popover_due(self, widget, popover):
+        self.popover = self.builder.get_object("date_popover")
```
izidormatusov: same here, make popover local only

GTG/gtk/editor/editor.py
izidormatusov: I still see it here :/

GTG/gtk/editor/editor.py
izidormatusov: I still see it here

jakubbrindza:
@izidormatusov please review, I hope it's all done now! :blush:

GTG/gtk/editor/editor.py
```diff
@@ -60,30 +64,25 @@ def __init__(self,
...
-        self.inserttag_button.set_tooltip_text(GnomeConfig.TAG_TOOLTIP)
-        self.open_parents_button = self.builder.get_object("open_parents")
-        self.open_parents_button.set_tooltip_text(
-            GnomeConfig.OPEN_PARENT_TOOLTIP)
+        self.donebutton = self.builder.get_object("mark_as_done")
+        self.undonebutton = self.builder.get_object("mark_as_undone")
+        self.dismissbutton = self.builder.get_object("dismiss")
+        self.undismissbutton = self.builder.get_object("undismiss")
+        self.add_subtask = self.builder.get_object("add_subtask")
+        self.store = self.builder.get_object("tag_store")
```
izidormatusov:
```python
self.tag_store = ...
```

GTG/gtk/editor/editor.py
```diff
@@ -204,27 +219,87 @@ def init_accelerators(self):
...
+
+    def open_tags_popover(self, widget):
+        self.store.clear()
+
+        treeview = self.builder.get_object("tags_tree")
+        column_toggle = self.builder.get_object("bool_column")
+        column_text = self.builder.get_object("tagname_column")
+
+        treeview.append_column(column_toggle)
+        treeview.append_column(column_text)
```
izidormatusov: this must be doable in .ui file

GTG/gtk/editor/editor.py
```diff
@@ -204,27 +219,87 @@ def init_accelerators(self):
...
+            if tag_filter(tag):
+                is_used = tag in used_tags
+                self.store.append([is_used, tagname])
+                '''
+                TODO(jakubbrindza): add sorting of the tags based on
+                True | False and within each sub-group arrange them
+                alphabetically
+                '''
+
+        treeview.set_model(self.store)
```
izidormatusov: This should be doable directly from .ui file

GTG/gtk/editor/editor.py
```diff
@@ -204,27 +219,87 @@ def init_accelerators(self):
...
+        treeview.show_all()
```
izidormatusov: I don't think you need this line anymore

GTG/gtk/editor/editor.py
```diff
@@ -251,6 +326,7 @@ def init_dimensions(self):
     # the new window title. If not, we will look for the task title.
     # Refreshtext is whether or not we should refresh the TaskView
     # (doing it all the time is dangerous if the task is empty)
+
```
izidormatusov: drop the extra line, it's not needed

GTG/gtk/ui/taskeditor.ui
```diff
@@ -0,0 +1,500 @@
...
+                        <property name="fill">True</property>
+                        <property name="padding">10</property>
+                        <property name="position">0</property>
+                      </packing>
+                    </child>
+                    <child>
+                      <object class="GtkEntry" id="startdate_entry">
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="invisible_char">●</property>
```
izidormatusov: drop this line

GTG/gtk/ui/taskeditor.ui
```diff
@@ -0,0 +1,500 @@
...
+                        <property name="fill">False</property>
+                        <property name="padding">10</property>
+                        <property name="position">0</property>
+                      </packing>
+                    </child>
+                    <child>
+                      <object class="GtkEntry" id="duedate_entry">
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="invisible_char">●</property>
```
izidormatusov: drop this line

GTG/gtk/ui/taskeditor.ui
```diff
@@ -0,0 +1,500 @@
...
+                        <property name="fill">False</property>
+                        <property name="padding">10</property>
+                        <property name="position">0</property>
+                      </packing>
+                    </child>
+                    <child>
+                      <object class="GtkEntry" id="closeddate_entry">
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="invisible_char">●</property>
```
izidormatusov: drop this line

GTG/gtk/editor/editor.py
jakubbrindza: :+1:

GTG/gtk/editor/editor.py
jakubbrindza: it is not shown without it

GTG/gtk/ui/taskeditor.ui
jakubbrindza: :+1:

GTG/gtk/ui/taskeditor.ui
jakubbrindza: :+1:

GTG/gtk/ui/taskeditor.ui
jakubbrindza: :+1:

GTG/gtk/editor/editor.py
jakubbrindza: :+1:

khurshid-alam:
@jakubbrindza  Do you intend to complete calendar widget in start due popover?

jakubbrindza:
@khurshid-alam unfortunately, I am unable to devote time to GTG anymore due to University :/ sorry about that.

khurshid-alam:
@jakubbrindza Well that is really unfortunate. Gtg has been left in broken state so many times before.Hopefully someone will pick up the project.

@izidormatusov May be we should revert to old ui where most thing was working. Thoughts?

PR #148, opened 2015-06-28
core/task: Support for 'Added date'

jdiez17:
I found it pretty annoying that subtasks are ordered alphabetically by default. I wanted to see the subtasks in the order I entered them, and there's no support for this in GTG.

I also felt that an "added date" field was missing, and this is a good way to sort the tasks in the order they were entered.

In the future, someone should add support for user-customisable ordering (and being able to reorder via drag and drop).

### DIFF
--- a/GTG/core/task.py
+++ b/GTG/core/task.py
@@ -794,11 +794,9 @@ class Task(TreeNode):
         return toreturn
 
     def __str__(self):
-        s = ""
-        s = s + "Task Object\n"
-        s = s + "Title:  " + self.title + "\n"
-        s = s + "Id:     " + self.tid + "\n"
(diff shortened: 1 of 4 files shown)


jdiez17:
@tkdchen I agree with both of your suggestions - I refactored `__str__` because it was a quick fix, but `treeview_factory` needs a bit more of attention. I think that'd be better suited for a different PR.

tkdchen:
@jdiez17 +1

diegogangl:
Recent changes have caused a ton of conflicts for this PR, so I ended up cherry-picking the commits and pushing them manually. Thanks for the work @jdiez17 !

Issue #92, opened 2015-03-17
Replace deprecated Gtk.ImageMenuItem by Gtk.MenuItem


izidormatusov:
[Documentation](https://developer.gnome.org/gtk3/stable/GtkImageMenuItem.html) contains many deprecation warnings for Gtk.ImageMenuItem. Replace it by Gtk.MenuItem.

izidormatusov:
@jakubbrindza , great job!

PR #155, opened 2015-07-10
Remove GtkImageMenuItem occurence

jakubbrindza:
**This PR solves #92**

I'm removing a left-behind ImageMenuItem widget and putting MenuItem instead.
It would be complicated to keep there the image. If you look at the documentation of ImageMenuItem it suggests creating a MenuItem, placing a Box inside and within that box place a Lable and an Image which would be a real mess to do in code!


jakubbrindza:
@izidormatusov I want to close the bug #92, therefore the PR.

Issue #117, opened 2015-06-01
Remove task count from title for task browser window


izidormatusov:
We should not display the task count in the title of task browser window. Somebody (can't find the email) pointed out that it goes against GNOME usability standard:

 - Only first couple characters/words might be available to describe the window when switching showing Alt+Tab dialog or representing the window in task switcher. Information that window belongs to GTG is much more useful than *21 ative tasks - GTG*.
 - Information about the task count is visible directly in UI
 - We moved to Headerbar, so the task count is hidden by default

Hence we should remove the task count from the title and set in `taskbrowser.ui` to a sensible value. `_update_window_title` and its references should be removed.

izidormatusov:
@jakubbrindza do you mind to have a look at this? It should be as simple as deleting definition and refrences of `_update_window_title`

PR #158, opened 2015-07-27
Remove task count from title in task browser

jakubbrindza:
This commit is a response to #117. The feature of displaying the number of tasks in stock is not needed anymore as we transfered to header bar and we are not interested in providing this kind of information there anymore.

Therefore, I removed the parts of code which took care of displaying this information.


jakubbrindza:
@izidormatusov take a look. If you'd like to see some further enhancements, let me know :blush:

jakubbrindza:
@tomspur I removed ngettext import :) hope looks better now

Issue #104, opened 2015-04-21
Search feature is not intuitive


ePaulo:
Make search feature more easily discoverable:
- [x] icon in the head bar
- [x] CTRL+F shortcut to start searching
- [x] separate search bar

Original text:

> A simple search tool would be helpful. I have accumulated a long list of tasks for one project and a search tool to find all tasks with a specific keyword or phrase would be nice.

izidormatusov:
You can search directly from task browser:

![screen shot 2015-05-07 at 09 06 05](https://cloud.githubusercontent.com/assets/98588/7510035/6057bc9a-f498-11e4-9b18-9cd2f05c3b3b.png)

Start typing into quick add bar and then select search. UI is not intuitive right now but @jakubbrindza is going to fix that over the summer.

nekohayo:
Merged in master.

PR #162, opened 2015-08-03
Add CTRL+F shortcut for search

jakubbrindza:
This commit solves the final suggestion in #104 - adds the keyboard shortcut for browser window. By pressing Control + F, user can substitute the search button's toggle functionality and view the search bar with the focus in search entry.

Once merged, #104 can be closed, too. Please, do that, so that our bugs list gets lower :blush:

### DIFF
--- a/GTG/gtk/browser/browser.py
+++ b/GTG/gtk/browser/browser.py
@@ -30,7 +30,6 @@ from GTG.core.dirs import ICONS_DIR
 from GTG.core.search import parse_search_query, InvalidQuery
 from GTG.core.tag import SEARCH_TAG
 from GTG.core.task import Task
-from GTG.core.translations import _
 from GTG.gtk.browser import GnomeConfig
 from GTG.gtk.browser.custominfobar import CustomInfoBar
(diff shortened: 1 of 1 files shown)


jakubbrindza:
I removed the translations _ import, since it broke the Travis Cl check.

izidormatusov:
This implementation is very simple but it doesn't work 100%. Let's say I start searching (by pressing CTRL+F). I find a task, open it but it is not the right task. I close it and want to search again. I press CTRL+F but instead of searchbar to be getting focus, it closes the search bar.

In other words, make it behave more like a browser: When you press CTRL+F, it should open and focus searchbar, so you can start typing. When the searchbar has focus and you press ESC, it should go away.

diegogangl:
This MR works as intended and is easily mergeable

nekohayo:
Thank you for spotting this low-hanging fruit Diego, I have now applied the patch to master. I had to manually grab the git-formatted patch file and apply with `git am --reject < 162.patch` to be able to ignore the conflicting first line, then staged the changes and did `git am --continue` before pushing.

PR #166, opened 2015-09-18
Update/complete french translation

poum:


jakubbrindza:
Hi, this is very useful! :blush: 
I can do a check of your translations and improvements if you'd like, so that we can merge this and have this complete :blush: 

By the way, what happens to your ci/circleci test? It fails right now, can you make sure it passes?
@izidormatusov may have other suggestions or ideas...

poum:
Hi, and thanks for GTG ! 
For the CI test error, I'm not sure (I've never use this before) but apparently nothing to do with my modifications: a python missing dependency, I guess this will not cause you any trouble ...
And of course, do any check you judge usefull

poum:
Should be better, found to 2 broken strings... sorry

izidormatusov:
circleci tests are probably broken

@ploum Do you mind to review the french translation messages?

Issue #172, opened 2015-10-12
broken in gnome 3.18


tomdurrant:
After updating to gnome 3.18, gtg is broken. Running from the command lines yeilds:

Traceback (most recent call last):
  File "./GTG/gtg", line 227, in <module>
    main()
  File "./GTG/gtg", line 173, in main
    manager = Manager(req)
  File "/home/tdurrant/git/gtg/GTG/gtk/manager.py", line 70, in __init__
    self.browser = TaskBrowser(self.req, self)
  File "/home/tdurrant/git/gtg/GTG/gtk/browser/browser.py", line 97, in __init__
    self.builder.add_from_file(GnomeConfig.BROWSER_UI_FILE)
GLib.Error: gtk-builder-error-quark: /home/tdurrant/git/gtg/GTG/gtk/browser/taskbrowser.ui:29:1 Can't use <style> here (4)

fturco:
I think I have the same problem. I have just updated to GNOME 3.18 on my Parabola GNU/Linux-libre system. GTG is version r5170.a795d66-1 from Git. I'm using a clean profile, so this is not a configuration problem. Error message:

```
$ gtg
sys:1: PyGIWarning: Gdk was imported without specifying a version first. Use gi.require_version('Gdk', '3.0') before import to ensure that the right version gets loaded.
sys:1: PyGIWarning: GnomeKeyring was imported without specifying a version first. Use gi.require_version('GnomeKeyring', '1.0') before import to ensure that the right version gets loaded.
sys:1: PyGIWarning: Gtk was imported without specifying a version first. Use gi.require_version('Gtk', '3.0') before import to ensure that the right version gets loaded.
sys:1: PyGIWarning: NetworkManager was imported without specifying a version first. Use gi.require_version('NetworkManager', '1.0') before import to ensure that the right version gets loaded.
sys:1: PyGIWarning: NMClient was imported without specifying a version first. Use gi.require_version('NMClient', '1.0') before import to ensure that the right version gets loaded.
2015-10-12 20:17:09,820 - WARNING - __init__:__init__:70 - Backend backend_launchpad could not be loaded: No module named 'launchpadlib'
2015-10-12 20:17:09,822 - WARNING - __init__:__init__:70 - Backend backend_rtm could not be loaded: No module named 'dateutil'
2015-10-12 20:17:09,824 - WARNING - __init__:__init__:70 - Backend backend_evolution could not be loaded: No module named 'dateutil'
[...]

PR #173, opened 2015-10-14
Resolving the issue in #172 with out-dated style property for GtkButton

jakubbrindza:
According to the latest standards for GNOME: 

https://developer.gnome.org/gtk3/stable/GtkButton.html#GtkButton.style-properties

no style properties distinguishing the type of the button are required, therefore this was the reason probably for fails mentioned in #172. 

I haven't tested the outcome on GNOME 3.18 yet, though, although the functionality and the design of GTG remains unchanged after these style properties were removed, therefore I think this will also enable launching on the new GNOME version. They were essentially useless and just made the code cluttered.

### DIFF
--- a/GTG/gtk/browser/taskbrowser.ui
+++ b/GTG/gtk/browser/taskbrowser.ui
@@ -25,9 +25,6 @@
           <packing>
             <property name="pack_type">start</property>
           </packing>
-          <style>
-            <class name="text-button"/>
-          </style>
         </child>
         <child>
           <object class="GtkBox" id="customisation_box">
@@ -45,9 +42,6 @@
                 <signal handler="on_view_sidebar_toggled" name="clicked" swapped="no"/>
                 <property name="valign">center</property>
               </object>
-              <style>
-                <class name="text-button"/>
-              </style>
               <packing>
                 <property name="pack_type">start</property>
               </packing>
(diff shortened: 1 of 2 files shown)


fturco:
I tested your version (branch "button_styles") and now gtg starts correctly with GNOME 3.18.1. I'm using the Arch Linux [AUR package](https://aur.archlinux.org/packages/gtg-git/) with the following source line:

    source=("${pkgname}::git+https://github.com/jakubbrindza/gtg#branch=button_styles")

fturco:
Unfortunately I just noticed that I cannot edit tasks because of the following error message:

    Lib.GError: gtk-builder-error-quark: /usr/lib/python3.5/site-packages/GTG/gtk/ui/taskeditor.ui:373:1 Can't use <style> here (4)

jakubbrindza:
@fturco that's extremely helpful, thank you! I hoped someone would do this testing for me, I really appreciate that!

Sorry for the inconvenience, I forgot to remove one more occurrence of the style property for one of the buttons in the task editor. The commit is now updated, so if you can, try it out for me please, and let me know, I am convinced though that there should be no other issues of this kind, used silver-searcher to check through the entire repo for these outdated style properties.

Once you approve, we can merge.

PR #192, opened 2016-03-23
Help not accessible from main window

### LINKED ISSUE #189: Help not accessible from main window
I tried to summon the help file from the main window, but couldn't find how: no button, no menu, and pressing F1 didn't do anything either.

There is a help button in the synchronisation dialog, so I eventually got back to the main page from there, but it's a bit convoluted.

sagarghuge:
Added shortcut key "F1" to open help

Fixes #189


diegogangl:
This works and can be merged easily

PR #208, opened 2017-05-09
GTM synchronization bug fix

volkov:
task.get_tags_name() do not contains '@' in tags but attached_tags do.

This is hack, and actually attached_tags should be changed, I'm not good at python and have no time to write good fix=(


diegogangl:
Hey @volkov. Oddly this was fixed in specific backends but not in the generic one. Instead of slicing it used `if "@" + tag in attached_tags:`. I've copied that fix for the generic backend instead of yours for consistency. Thanks for your pointing to this and your work though!

PR #219, opened 2019-11-24
UI Polish / Improvements

diegogangl:
Hi guys, this PR includes a few changes to the UI to bring it more in line with the current HIG and general UI style.

- Moved tags, synchronization, plugins and preferences to a main (burger) menu
- Brought back About and Help which were probably lost in the app menu
- Removed "tasks" from the names of the tabs to avoid repetition
- Added emoji input and button to quick entry. (I don't know if the button is really needed TBH, but it does look nice)
- Removed double borders in the sidebar
- Added some padding to the quick entry to avoid double edges there too
- Improved the padding in the sidebar
- Used symbolic icons for sidebar

There are two issues:
- The F9 shortcut only works if the main menu popup is open. I couldn't figure out how to fix this. Maybe we need to add a custom signal to the window or something like that instead?
- I wanted to add the `.sidebar` class to make the sidebar look more like Nautilus', but the tree widget is doing something to the styling of the cells and I couldn't figure out how to change it.

Mandatory Screen:
Screenshot from 2019-11-23 21-34-04

### DIFF
--- a/GTG/core/treefactory.py
+++ b/GTG/core/treefactory.py
@@ -73,7 +73,7 @@ class TreeFactory(object):
         # Build the "all tasks tag"
         alltag = tag.Tag(tag.ALLTASKS_TAG, req=req)
         alltag.set_attribute("special", "all")
-        alltag.set_attribute("label", "<span weight='bold'>%s</span>"
+        alltag.set_attribute("label", "%s"
                              % _("All tasks"))
         alltag.set_attribute("icon", "gtg-tags-all")
         alltag.set_attribute("order", 0)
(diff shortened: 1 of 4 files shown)


nekohayo:
And now the real (global) comment I wanted to post, first and foremost: What a pleasant surprise! **THANKS!!** The proposed improvements are as if you had read my mind, with exactly the changes that were needed for the main UI! So from the UX perspective, definite :+1:

From the code mergeability perspective: from a cursory glance, it looks good to me, though I didn't pay excruciating attention to detail in some areas like the 1st commit cfcf39ad which seems to be mostly Glade's eternal "Let's rename/move the order of lines and parameters in the XML all over the place for fun and giggles", so overall harmless noise. The rest of the commits seem to be trivial changes and so unless @ploum spots something odd I think we could just hit the merge button (as far as I understand Github lets you do that without even having to touch git by hand?)

diegogangl:
Hey, awesome! :)
About the icons: Search is always going to be search. The others are a documents emblem, and task-past-due. We could bundle them in case they change in the future. I don't know if icon themes also change the symbolic icons though, sounds like a lot of work.

ploum:
merging it as it can only improve the trunk.

ploum:
Thanks a lot for the work :-)

diegogangl:
Thanks for merging :D

PR #227, opened 2019-12-28
Preferences fixes

### LINKED ISSUE #154: Setting Autostart and "New Task" shortcut in preferences not working
In current preferences window, when trying to set GTG to autostart on log in or to set the New Task shortcut, the system seems not to remember my decision in this.

I set it to "ON" (a.k.a activate the check button), close the preferences window, open it again, and it is set to "OFF" (a.k.a. check button de-activated). This means that I can actually never set these two settings to active and make them functional.

Maybe some dependencies required for them which I'm missing? 
If not dependencies, this needs to be addressed anyway for #153 or its descendants...

diegogangl:
This PR updates the preferences dialog and removes a few options (cue the Gnome devs removing features meme)

- Modernizes the settings dialog (following Gnome Builder prefs for ideas)
- Also bump versions for all glade files and fix the general preferences ui file
- Remove autostart setting. We mentioned this in #154 
- Remove global shortcut setting. I spent some time trying to get this to work until I asked in #gnome-hackers. I was told applications messing with shell settings is frowned upon, and it wouldn't work with flatpak anyways. If we want to implement this feature in the future we could add a parameter to gtg that open the task editor automatically and then have a short tutorial on how to add a global shortcut to it.

fixes #154 (nuking everything is also a way of fixing things ¯\\\_(ツ)_/¯ )


ploum:
Awesome!

PR #235, opened 2020-01-31
UI changes

diegogangl:
Hey o/

I'm pushing this one through the review system since it includes some UI changes you might want to discuss.

## Task browser

- Change the tabs name to "Open" and "Next Up" (instead of Active and Workview). I think this is more intuitive, the tabs follow a single convention and they look like a factory line for the life of a task.
open -> next -> closed. This means updating the help/wikis though, so more work for @nekohayo :)

Screenshot from 2020-01-30 21-14-42

## Task editor
- Follow the new HIG guidelines for titles
- Simplify "Mark as Done" to "Done"
- Make it a suggested action (blue button)
- Add some whitespace for the task text
- Make subtasks a bit more subtle and integrated

(updated image here)
Screenshot from 2020-03-10 12-40-58

### DIFF
--- a/GTG/gtk/browser/taskbrowser.ui
+++ b/GTG/gtk/browser/taskbrowser.ui
@@ -437,8 +437,8 @@
                             </child>
                           </object>
                           <packing>
-                            <property name="name">active_view</property>
-                            <property name="title">Active</property>
+                            <property name="name">open_view</property>
+                            <property name="title">Open</property>
                           </packing>
                         </child>
                         <child>
@@ -451,8 +451,8 @@
                             </child>
                           </object>
                           <packing>
-                            <property name="name">work_view</property>
-                            <property name="title">Workview</property>
+                            <property name="name">actionable_view</property>
+                            <property name="title">Actionable</property>
                             <property name="position">1</property>
                           </packing>
                         </child>
--- a/GTG/gtk/editor/taskview.py
+++ b/GTG/gtk/editor/taskview.py
@@ -106,10 +106,12 @@ class TaskView(Gtk.TextView):
         # but set in self.modified)
         self.table = self.buff.get_tag_table()
         # Tag for title
-        self.title_tag = self.buff.create_tag("title", foreground="#007bff",
-                                              scale=1.6, underline=1)
-        self.title_tag.set_property("pixels-above-lines", 10)
-        self.title_tag.set_property("pixels-below-lines", 10)
+        self.title_tag = self.buff.create_tag("title",
+                                              weight=Pango.Weight.ULTRABOLD,
+                                              size_points=20,
+                                              pixels_above_lines=10,
+                                              pixels_below_lines=20)
+
         # Tag for highlight (tags are automatically added to the tag table)
         self.buff.create_tag("fluo", background="#F0F")
         # Tag for bullets
--- a/GTG/gtk/editor/taskview.py
+++ b/GTG/gtk/editor/taskview.py
@@ -43,8 +43,8 @@ separators = [' ', ',', '\n', '\t', '!', '?', ';', '\0', '(', ')']
 # are part of the word
 specials_separators = ['.', '/']
 
-bullet1_ltr = '→'
-bullet1_rtl = '←'
+bullet1_ltr = '→  '
+bullet1_rtl = '  ←'
 
 
 class TaskView(Gtk.TextView):
@@ -98,8 +98,8 @@ class TaskView(Gtk.TextView):
                        'underline': Pango.Underline.SINGLE}
         self.hover = {'background': 'light gray'}
         self.tag = {'background': "#FFea00", 'foreground': 'black'}
-        self.indent = {'scale': 1.4, 'editable': False, 'left-margin': 10,
-                       "accumulative-margin": True}
+        self.indent = {'scale': 1, 'editable': False,
+                       "weight": Pango.Weight.ULTRABOLD }
 
         # Tag we will use ###########
         # We use the tag table (tag are defined here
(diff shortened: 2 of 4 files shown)


ploum:
Hello @diegogangl,

Thanks for your wonderful work. Due to lot of unexpected stuff happening in my life, I will totally trust you and @nekohayo to continue GTG. If you need a specific advice, send me an email.

I'm really happy to see the effort and the work you are doing. GTG seems to be in very good hand :-)

diegogangl:
Hey @ploum, thanks for the kind words :)

tkdchen:
I feel "Done" could be ambiguous and confuse people who don't know the changes to task content are stored automatically. That is, it is for marking current opened task as Done state, or just meaning "task content edition is done, click the button to close the task window".

I prefer to the "Mark as Done", or some other shorter version, or even use a GtkCheckButton.

nekohayo:
> Simplify "Mark as Done" to "Done"

I had thought of that too, a long while ago, but auto-convinced myself that it was a dangerous idea. The problem with naming it just "Done" is that, as Chenxion Qi mentioned, it is ambiguous in a GUI, particularly in English where it can be interpreted as "I'm done with the changes / dismiss this window", in the same way as just having it labelled "Close" would have people interpret it as "Close the window" (GitHub doesn't have this problem because nobody closes a window from a form inside a webpage :)

Conversely, I do use the "Mark as not to be done anymore" button (which has no label for obvious reasons, similarly to "Close/Done" you can't just label it "Dismiss"...) and the "Mark as not done" buttons quite often in GTG 0.3.x

I wish I had better shorter-but-still-accurate wording suggestions, but so far I don't :)

GTG/gtk/browser/taskbrowser.ui
```diff
@@ -452,7 +452,7 @@
                           </object>
                           <packing>
                             <property name="name">work_view</property>
-                            <property name="title">Workview</property>
+                            <property name="title">Next Up</property>
```
nekohayo: Wouldn't "Actionable tasks" be clearer?

And this could also benefit from a (translateable) GtkTooltip that says that this view only shows tasks that have no subtasks/dependencies and that are not set to begin in the future or associated with a tag that is forbidden to be shown in this view (I guess this is why it was called workview; it allowed you to exclude certain tags from the work view...)

GTG/gtk/browser/taskbrowser.ui
```diff
@@ -438,7 +438,7 @@
                           </object>
                           <packing>
                             <property name="name">active_view</property>
-                            <property name="title">Active</property>
+                            <property name="title">Open</property>
```
nekohayo: Aaah, the great debate of "Active" vs "Open" vs "To do" vs "All tasks" vs "Non-closed" tasks ;) I'm not sure which formulation is best.

GTG/gtk/editor/taskview.py
```diff
@@ -43,8 +43,8 @@
 # are part of the word
 specials_separators = ['.', '/']
 
-bullet1_ltr = '→'
-bullet1_rtl = '←'
+bullet1_ltr = '→  '
+bullet1_rtl = '  ←'
```
nekohayo: Won't this break if the user deletes the extra white spaces by mistake, or inserts their contents (with the mouse) inbetween the two whitespaces?

GTG/gtk/ui/taskeditor.ui
```diff
@@ -120,12 +120,15 @@
         <property name="show_close_button">True</property>
         <child>
           <object class="GtkButton" id="mark_as_done">
-            <property name="label" translatable="yes">Mark as Done</property>
+            <property name="label" translatable="yes">Done</property>
```
nekohayo: I would recommend not relabelling that one, for clarity.

GTG/gtk/editor/taskview.py
diegogangl: Surprisingly no, you have to backspace at the first character of the task name to delete the the icon

diegogangl:
@tkdchen @nekohayo thanks for the comments on the button. You are right, it's kind of confusing. On top of that, I took a look at the HIG and we are supposed to use imperative verbs.
I'll drop/fix that commit before merging.

A GtkCheckButton could be an interesting idea to explore, though I've never seen one in a headerbar

GTG/gtk/browser/taskbrowser.ui
diegogangl: Actionable could work. The only problem I see is that it's GTD lingo. I know this is what GTG is based on, but using domain-specific words set a bar of entry for users. I think most new users would wonder what actionable means, and what's the difference with active/open.

> this could also benefit from a (translateable) GtkTooltip

Yeah, we should add tooltips to all of them.

> (I guess this is why it was called workview; it allowed you to exclude certain tags from the work view...)

I thought it was called work view because it was a button in the toolbar, and the name made more sense in that context :thinking:

GTG/gtk/browser/taskbrowser.ui
diegogangl: Been thinking about this:
- "All tasks" would also include closed tasks, which we don't show in that pane
- "Active" sounds like the tasks are being worked on or actionable right now
- "To Do" could work, but the second pane also includes to-dos

nekohayo:
To clarify on my previous comment on why I think "Actionable" would be much clearer than "Next up": as someone who also practices kanban, "Next up" means a very different thing to me, it means "Whatever I explicitly chose to be the next things to work on", a shortlist of "next priorities I decided to tackle", whereas  GTG's workview is "_all_ the tasks that are actionable". Calling it "Next up" instead of "Actionable" _might_ confuse some people into thinking that they can somehow pick/flag/pin tasks into this view?

diegogangl:
I did some changes manually and cherry others, so closing this PR

PR #236, opened 2020-02-05
Add missing dependent packages

tkdchen:
These two packages are required when trying to run GTG from a local
virtual environment by executing script gtg.sh

Signed-off-by: Chenxiong Qi <qcxhome@gmail.com>


diegogangl:
LGTM, thanks!

PR #239, opened 2020-02-07
Add tox support

freyes:
Tox is a widely spread testing automation tool for python projects,
it still uses the Makefile under the hood, but tox will take care
of managing virtualenvs and running the tests inside of them.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,12 @@
+[tox]
+envlist = py37,pep8
+sitepackages=True
+
+[testenv]
+sitepackages=True
+deps=-rrequirements.txt
+commands=make tests
+whitelist_externals=/usr/bin/make
+
+[testenv:pep8]
+commands=make pep8
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ tmp
 .pydevproject
 .project
 .settings
+.tox
 **/**.glade.h
 **/**.gladep
 **/**.gladep.bak
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@
 # this program.  If not, see <http://www.gnu.org/licenses/>.
 # -----------------------------------------------------------------------------
 
-PEP8=pep8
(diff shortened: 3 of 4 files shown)


tox.ini
```diff
@@ -0,0 +1,12 @@
+[tox]
+envlist = py37,pep8
+sitepackages=True
+
+[testenv]
+sitepackages=True
```
tkdchen: Is this necessary? All the required packages listed in requirements.txt can be installed into a virtual environment.

diegogangl:
Hey @freyes thanks for contributing! I've been reading about Tox, but I'm still wondering what's the benefit of this over just using venv + pip (for us)? Does it really warrant adding another dependency?

tkdchen:
@diegogangl In general, IMO, as a Python developer, when see a project has `tox.ini`, he/she can simply know tests could run by running command `tox`. The `tox` could be installed from Linux distro package repos, or PyPI into a venv.

tox.ini
freyes: without this some things are left out, for instance dbus python module, I could chase down the missing dependencies (most likely gtk and friends) so we could remove this line.

diegogangl:
Ok, I looked a bit more into this and it seems like this would be an improvement for our tests and could make CI easier too. Is using `sitepackages=True` a problem, or does it have any drawbacks?

tox.ini
tkdchen: dbus python module is included in https://github.com/getting-things-gnome/gtg/commit/d9c4a3bb710bc693bd83baa0235afafe4e0220e9

Currently, I'm running gtg just from a venv created without `--system-site-packages`.

diegogangl: Tox still runs fine for me without those lines

diegogangl: Friendly ping. @federicomenaquintero do you want to try to add the dependencies? Or are they not needed? @tkdchen what do you think?

PR #248, opened 2020-02-27
debug.sh: Drop -l from usage help as it is not a valid option anymore

leio:
The option is gone from main gtg binary since commit 5131f6fae6c4efed,
where it was also dropped from debug.sh script, but appears it was
forgotten to be dropped from the usage help string.


diegogangl:
@leio nice catch, thanks!

PR #249, opened 2020-03-04
gi.require_version raises uncatched exception

liberforce:
This happens when GNOME Keyring is not found.


diegogangl:
Merged, thanks! :D

PR #254, opened 2020-03-15
Adapt delete tag dialog to modern UI

### LINKED ISSUE #253: Adapt delete tag dialog to modern UI
The delete tag dialog still follows the old GTK2/early GTK3 style, we need to update it to follow the modern Gnome style. If you are wondering what this looks like try deleting a task in the browser.

This is pretty easy since you can adapt the code from the task delete dialog. 

Files: 
- `GTG/gtk/delete_dialog.py` has the task delete dialog
- `GTG/gtk/deletetags_dialog.py` is what you need to hack
- `GTG/gtk/deletetags_dialog.ui` needs to be removed

kristho91:
Fixes #253


diegogangl:
Great work, thanks!

PR #300, opened 2020-05-03
Fix the docs folder link in README.md

gszy:


nekohayo:
Pushed your patch to master, thanks for spotting my mistake :)

PR #301, opened 2020-05-04
Hide delete tag option in context menu for saved search

### LINKED ISSUE #283: Saved searches has "Edit Tag..." and both "Delete tag" and "Delete" contextual menu items
If you save a search, and then right-click it inside the sidebar, you have these options:

* Edit Tag... (should probably be named differently; it lets you change the "nickname" of the search and assign a color or icon, which is interesting, but the UI was clearly for tags...)
* Generate Color
* Delete Tag (doesn't do anything)
* Delete

"Delete Tag" should be removed and "Delete" should probably become "Delete saved search".

jscn:
Deleting the tag in this case doesn't do anything since it's actually a saved search, without a '@' prepended to the text.

Fixes #283 

There is a bigger issue here, in that it doesn't seem possible to actually alter the search term(s) once the search has been saved. So having a "Edit Search..." option that just displays the same window as for editing tags is misleading. Probably the "Edit Tag..." option should not be displayed at all for saved searches. Instead there should be a window specifically for editing saved searches, which actually allows you update the saved search terms. I thought that was probably out of scope for this ticket.

### DIFF
--- a/GTG/gtk/browser/tag_context_menu.py
+++ b/GTG/gtk/browser/tag_context_menu.py
@@ -90,8 +90,8 @@ class TagContextMenu(Gtk.Menu):
     def on_mi_ctag_activate(self, widget):
         random_color = generate_tag_color()
         present_color = self.tag.get_attribute('color')
-        if(present_color is not None):
-                color_remove(present_color)
+        if present_color is not None:
+            color_remove(present_color)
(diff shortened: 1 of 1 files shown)


diegogangl:
Works as advertised :), thanks! 
You are correct about editing searches, I've opened a net ticket for that: #302

Issue #237, opened 2020-02-07
Code Quality


diegogangl:
This issue is related to small things that can be done to improve code quality and make it more approachable to new contributors.
This is an on-going effort.

- [ ] Organize all .ui files in one place
- [ ] Rename and organize .py files in gtk (following a consistent system)
- [ ] Replace old % formatting with f-strings
- [ ] Use GTK resources to manage and load ui and css files
- [ ] Remove old icons and things that aren't used anymore
- [ ] Use Gtk.Template decorator to load ui files (when possible?)
- [ ] Use Gapplication and GtkApplicationWindow (looks useful, but needs some research)
- [ ] Add [Pylint as a commit hook](https://git-pylint-commit-hook.readthedocs.io/en/latest/usage.html#support-for-pylintrc-files) to keep a consistent style and check for code smells (we need to make a `pylintrc` file first, since it's way too strict)
- [ ] Look around for TODOs/FIXMEs/etc and either remove them, fix them or add them as a ticket here
- [ ] Remove Python2-isms, like `class Something(object)`

nekohayo:
That all makes sense! Thanks for laying out the plan for it all.

If I may add my 2¢ of peanut gallery opinion, I would like to suggest officially dumping the prehistoric "Manually wrap anything that exceeds 72 or 79 characters/columns" rule, but otherwise respecting PEP8. 

When I peeked at the GTG codebase a few months ago, the code was filled with those over-zealous linebreaks. I personally think the 79-columns rule is stupid (hey, we're not running 640x480 screens anymore, folks!) and hurts code legibility more than it helps in this millenium.

In Pitivi we had agreed upon my proposed "compromise" approach (wrap if the line is really unbearably long, otherwise try to keep it on one line), as documented here: https://gitlab.gnome.org/GNOME/pitivi/blob/master/docs/Coding_style_guide.md 

* If you think manually breaking lines at arbitrary positions (instead of truly semantic positions or not at all) is ugly, we could make it a code style recommendation in GTG as well (I guess it would just be a matter of documenting that policy and slacking off the pylint checker config and such). Since I'm not the one who has to wrestle with the majority of the codebase I'm merely suggesting here of course, but I do think it makes sense.
* Hint: if you want to save some time, you could totally steal the pre-commit hooks and pylint configs linked from the intro of that Pitivi code style documentation page ;)

diegogangl:
[...]

PR #311, opened 2020-05-16
Shellcheck fixes

asadana:
Minor cleanup of the shell scripts to fix common shellcheck warnings and errors

### DIFF
--- a/gtg.sh
+++ b/gtg.sh
@@ -5,4 +5,4 @@
 # It refers to the GTG debug script originally placed in /scripts/debug.sh.
 # Makes life much easier.
 
-exec ./scripts/debug.sh $@
+exec ./scripts/debug.sh "$@"
--- a/gtg.sh
+++ b/gtg.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # This script allows contributors to run GTG directly from the gtg directory without the need to navigate
 # to the scripts folder. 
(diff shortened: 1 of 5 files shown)


asadana:
While not listed in #237, this should help with Clean Up section of the issue

asadana:
I see the problem. Sorry about that, fix coming up

asadana:
@diegogangl 

There is currently `yes` command piped to `cp`. This is not necessary in the modern versions of `cp`, shellcheck throws a warning for this, but I haven't changed it incase this was put in for a purpose. Please let me know if you want these removed.

```sh
yes|cp -Rf ~/.local/share/gtg/* tmp/default/xdg/data/gtg/
yes|cp -Rf ~/.config/gtg/* tmp/default/xdg/config/gtg/
```

diegogangl:
Indeed, cp doesn't read stdin. Yeah, it'd be nice to remove those too.
About debug.sh: it doesn't look like arguments are being passed to gtg at the end. You can try this by passing `-d`. You should see a ton of log messages and a gear icon in the window headerbar

asadana:
> Indeed, cp doesn't read stdin. Yeah, it'd be nice to remove those too.
> About debug.sh: it doesn't look like arguments are being passed to gtg at the end. You can try this by passing `-d`. You should see a ton of log messages and a gear icon in the window headerbar

I reverted the commit to see what I was missing since I made sure the title was working yesterday.
It seems double-quoting of `args` in debug script messes up python argument flags besides explicitly specified `-t`.

I have fixed and tested `-d` flag. This required me to disable shellcheck for that specific launcher line (commented with informational message for future contributors)

I have also removed the `yes` pipe before `cp` command.

PR #315, opened 2020-05-17
Initial meson port

leio:
Initially this only tries to mostly replicate what distutils
was doing, with further changes planned separately after setup.py
can go away, as these would require adapting setup.py to the changes
as well.
There are some difference, however:
* gtcli and gtg_new_task are not installed, as they are broken after
  dbus removal and will get removed soon separately.
* setup.py installed all *.py and other files unconditionally, but
  with meson we list them explicitly - as such we intentionally don't
  install unmaintained backends and plugin files anymore. gtg main
  binary doesn't get installed twice either anymore (setup.py put it
  together with the rest in site-packages, but then also into bindir).
* as it's not distutils, the *.egg-info file isn't installed.
* appdata file is installed to /usr/share/metainfo instead of deprecated
  /usr/share/appdata location.
* as gtcli isn't installed, the bash completion file for it isn't either.
* gtg manpage is actually installed - didn't seem to be with setup.py
* Russian user manual is not installed, as the modern system expects
  the translation to be in a PO catalog file, not all the *.page files
  replicated with the strings changed. To preserve it, it'd have to be
  somehow converted, but probably not worth it - especially if the
  english text will be overhauled still, and seems to already have
  compared to when the translation was made.


leio:
As this version doesn't yet remove setup.py, nor change the result as installed by setup.py by my testing, maybe it can go into 0.4 as an alternative option that is marked experimental? Then we have a version that supports both (unless we end up deciding that we put in more meson stuff into 0.4 that would need setup.py changes that we don't bother with and remove it instead).

diegogangl:
I gave this a try and everything looks good :+1: 
It'd be nice to include some documentation on how to build/install/generate *.pot files once we replace setup.py

Issue #355, opened 2020-06-05
Update Needed for Wording in the General Preferences UI


vansia43:
This UI uses British English for the word "coloured" unlike the rest of the UI for the Enabled Colored Background setting.

nekohayo:
Fix pushed as commit 90e12286

PR #356, opened 2020-06-05
Update UI text for General Preferences

vansia43:
Addresses #355 as well as a few additional grammar and style updates.

### DIFF
--- a/GTG/gtk/data/general_preferences.ui
+++ b/GTG/gtk/data/general_preferences.ui
@@ -361,7 +361,7 @@
                           <object class="GtkLabel">
                             <property name="visible">True</property>
                             <property name="can_focus">False</property>
-                            <property name="label" translatable="yes">Automatically remove old, closed tasks</property>
+                            <property name="label" translatable="yes">Automatically remove old closed tasks</property>
                             <property name="xalign">0</property>
                           </object>
                           <packing>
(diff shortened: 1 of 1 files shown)


GTG/gtk/data/general_preferences.ui
```diff
@@ -361,7 +361,7 @@
                           <object class="GtkLabel">
                             <property name="visible">True</property>
                             <property name="can_focus">False</property>
-                            <property name="label" translatable="yes">Remove old closed tasks automatically</property>
+                            <property name="label" translatable="yes">Automatically remove old, closed tasks</property>
```
nekohayo, line 364: I see a potential issue with the comma here... "old closed" is clearer to me than "old, closed" because "old, closed" could be two separate things, if you see what I mean?

vansia43: My thought process on this: "old" and "closed" are coordinate adjectives that equally describe the noun "tasks". The task is both old and closed. Coordinate adjectives are separated by a comma; however, I can remove it if you think that this would impede someone's understanding. I am fine either way :)

nekohayo: Yeah, for the avoidance of doubt in the user's mind about whether or not it's an enumeration of independent things, I'd tend to avoid the comma if you don't mind! Because if I was casually reading this I'd wonder "wait so will it delete old tasks and closed tasks?"

vansia43: Removed per below commit. Thanks!

nekohayo:
Alright, I've squashed your two commits together and pushed them to master.

Issue #362, opened 2020-06-12
Some plugin related strings aren't translatable


Neui:
The strings are translated in the german (de) files, but they aren't translated in the application itself.

The following screenshot shows Gtranslator in the background with the original and translated strings (which aren't marked specially), and the plugins window showing an plugin that you can't load and another window showing the plugin information, again with untranslated strings.

![Screenshot showing translated strings in gtranslator but in the application the strings arent translated](https://user-images.githubusercontent.com/1196130/84554485-5976ce00-ad18-11ea-8a7b-8db81b475241.png)  
(Also, shouldn't it be "This plugin can not be loaded"?)

These strings are defined in [`GTG/core/plugins/__init__.py`](https://github.com/getting-things-gnome/gtg/blob/adf61eee45c7a98c031d1216e02532678c4e6571/GTG/core/plugins/__init__.py), maybe they get initialized before the gettext part is initialized?

Using git master (abe2a9110dd0fc6a46f2d095013972877ea67d78).

Neui:
[The strings in `dates.py`](https://github.com/getting-things-gnome/gtg/blob/master/GTG/core/dates.py#L43-L48) suffer from a similar thing. In this case, I copied that dictionary to the [`to_readable_string`](https://github.com/getting-things-gnome/gtg/blob/master/GTG/core/dates.py#L445) method and it shows up translated for me in the main window tree view thing.

Issue #357, opened 2020-06-09
In the due date (and similar) treeview columns, the "now", "soon", "someday" strings don't show up in their translated versions, although others do


nekohayo:
Let's say you run with `LANG=fr_CA.UTF8 ./launch.sh`. In the main window, in the due date column (and probably also the start date, closure date, etc.) these strings will show up translated:

* Today
* Tomorrow
* In X days
* etc.

But not these ones, even if they are picked up by gettext for the .pot and translated in fr.po:
* now
* soon
* someday

The last time I looked at it, I couldn't figure out why, because the code in GTG/core/dates.py and GTG/gtk/browser/main_window.py didn't seem to do something particularly wrong...

Neui:
Seems to have a similar problem to #362

In this case, the [`dates.py` file also initializes the translated strings](https://github.com/getting-things-gnome/gtg/blob/abe2a9110dd0fc6a46f2d095013972877ea67d78/GTG/core/dates.py#L43-L48) before gettext (the translation library) is initialized.
An workaround could be copying that dictionary to [`to_readable_string`](https://github.com/getting-things-gnome/gtg/blob/abe2a9110dd0fc6a46f2d095013972877ea67d78/GTG/core/dates.py#L445) (and indent it properly, because Python), but that would (I assume) always recreate the dictionary that is otherwise static and thus not that efficent.

PR #365, opened 2020-06-13
Tasks column translatable

Neui:
Note that, at least in the german version, there is existing translation for this string in an #~ "comment" (gettext complained about duplicate string when I added that string manually).
This translation may be re-used next time translations are refreshed.

Additionally, in `common_desc_for_tasks` the parameter `title_label` is getting translated again.
This is removed since callers already translate their label and has the benefit of gettext finding them.

### DIFF
--- a/GTG/gtk/browser/treeview_factory.py
+++ b/GTG/gtk/browser/treeview_factory.py
@@ -325,7 +325,7 @@ class TreeviewFactory():
 
     def active_tasks_treeview(self, tree):
         # Build the title/label/tags columns
-        desc = self.common_desc_for_tasks(tree, "Tasks")
+        desc = self.common_desc_for_tasks(tree, _("Tasks"))
 
         # "startdate" column
         col_name = 'startdate'
--- a/GTG/gtk/browser/treeview_factory.py
+++ b/GTG/gtk/browser/treeview_factory.py
@@ -422,7 +422,7 @@ class TreeviewFactory():
         # "label" column
         col_name = 'label'
         col = {}
-        col['title'] = _(title_label)
+        col['title'] = title_label
         render_text = Gtk.CellRendererText()
         render_text.set_property("ellipsize", Pango.EllipsizeMode.END)
         col['renderer'] = ['markup', render_text]


nekohayo:
I'm not sure about ee5c9a4, doesn't GTG use that string mostly as a placeholder and then dynamically replace it by the translatable strings for each view mode (ex: "Closed Tasks" or something like that)?

The other commit, 8d28234 : could this be the fix for https://github.com/getting-things-gnome/gtg/issues/357 ?

Neui:
> I'm not sure about [ee5c9a4](https://github.com/getting-things-gnome/gtg/commit/ee5c9a42a246f376d6ce7bd28257ace24fd69723), doesn't GTG use that string mostly as a placeholder and then dynamically replace it by the translatable strings for each view mode (ex: "Closed Tasks" or something like that)?

[I was just copying what the path for "Closed Tasks" does](https://github.com/getting-things-gnome/gtg/blob/ee5c9a42a246f376d6ce7bd28257ace24fd69723/GTG/gtk/browser/treeview_factory.py#L359).
Note that, yes, it would eventually do `_("Tasks")` in the old code, but since this is the only occurrence of `"Tasks"`, and since it isn't marked to be translated (the `_("String")` thing), gettext (the library that "does" the translations) sees that (old) string as not being used. It is marked as `#~` inside the PO file that, [according to a website](http://pology.nedohodnik.net/doc/user/en_US/ch-poformat.html), means "obsolete", and is otherwise ignored by gettext.

According to a grep, `common_desc_for_tasks` is only called in two places (the one I modified in the first commit and the existing one with `_("Closed Tasks")`). If I would kept it, it would just try to "translate" the string again. It won't hurt, but it is doing unnecessary stuff.

> The other commit, [8d28234](https://github.com/getting-things-gnome/gtg/commit/8d28234003745d874069275853a2e00d029943b0) : could this be the fix for #357 ?

No, [but I've also noticed that problem](https://github.com/getting-things-gnome/gtg/issues/362#issuecomment-643615621), it seems it has something to do when the code translates strings before it sets up gettext. I suggested you could copy that dictionary to the place where it is used, which might not be as efficient (recreating the otherwise static dictionary every time).

nekohayo:
Rebased and pushed your commits to master, thanks!

PR #368, opened 2020-06-16
Added reference to the AUR package

sandervankasteel:


diegogangl:
Hey thanks for the work! Could you add a note that this are non-official third-party packages? 
We are ok with people packaging for their distro but we only officially support flatpak.

Pinging @nekohayo to help here (the readme is your baby after all :) )

sandervankasteel:
Yes of course! Will do so :)

nekohayo:
Hmmm... actually I think this is a better fit for https://wiki.gnome.org/Apps/GTG/install than the README file, so I added a mention of your package there instead (otherwise I suspect that eventually our README would become a mess to maintain with all the distro packages). Hoping that's OK!

sandervankasteel:
@nekohayo It's definitely okay! Don't sweat about it. :) Now that I think about it, eventually the README file would indeed because a mess, filled with references to distro packages.

Issue #299, opened 2020-05-03
Better export plugin


diegogangl:
The current export plugin is based on several dependencies that are dead for most distros or require  serious hunting. This could be simplified by having a simpler plugin that exports to markdown.
Then we can hook it up with [pandoc](https://pandoc.org/) to convert to html, pdf, latex, cuneiform, etc.

This way we only have one dependency that's well maintained and packaged.

nekohayo:
A small note while I'm at it (and to avoid creating yet another ticket): currently the export plugin "almost" works, when running on Ubuntu it can be enabled and the "Export the tasks currently listed" (sic) menu item shows up in the main window's menubutton, however for some obscure reason that menu item cannot be selected, when you hover the mouse over it there is no selection highlight and you can't click it. It's as if it was insensitive, witohut being insensitive.

kocio-pl:
With current git version exporting is possible with 3/5 templates. There is a easy fix for PDF template (#833) and only simple text template has no fix at the moment (#832).

PR #370, opened 2020-06-19
Export plugin: Make it clickable in the menu

Neui:
Looks like since the GTK3 Headerbar switch, the new menus use ModelButtons instead of the old MenuItems.
[Fixes the comment by nekohayo](https://github.com/getting-things-gnome/gtg/issues/299#issuecomment-638259673) in #299.

Found using the [GTK Inspector](https://developer.gnome.org/gtk3/3.24/gtk-running.html#interactive-debugging) (Ctrl+Shift+i doesn't work for some reason, I used `GTK_DEBUG=interactive ./launch.sh`: (The export plugins menu entry is called "Exportiere die aktuell gelistete Aufgaben" here)
Image of the GTK Inspector showing the button being GtkMenuItem instead of GtkModelButton like the others

Note that the plugin itself seems to be buggy, I couldn't export anything (to check what it really exports, like does it take into account the current filter being applied?)
When trying to export (using the open button), I just get an error message that the document couldn't be exported, and a reason is some python error, so there is still debugging to be done.
Possibly remove the plugin temporarily for the upcoming release, if there isn't enough time.

(BTW the "open" string when exporting conflicts with the "open" (view) string in the main window for the open tasks; There is an distinction in german, I would likely make an PR or Issue to mark the "open" (view) string in the main window specially, so it has its own translation)

### DIFF
--- a/GTG/plugins/export/export.py
+++ b/GTG/plugins/export/export.py
@@ -180,8 +180,9 @@ class ExportPlugin():
         """ Initialize all the GTK widgets """
         self.menu_entry = False
 
-        self.menu_item = Gtk.MenuItem(_("Export the tasks currently listed"))
-        self.menu_item.connect('activate', self.show_dialog)
+        self.menu_item = Gtk.ModelButton()
+        self.menu_item.set_label(_("Export the tasks currently listed"))
+        self.menu_item.connect('clicked', self.show_dialog)
         self.menu_item.show()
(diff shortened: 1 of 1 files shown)


diegogangl:
Thanks, merged. We'll probably include this plugin for this version and then move on to the new one (whenever it's done) for 0.5

PR #371, opened 2020-06-19
Add translation context for the tasks view names

Neui:
The open tasks view (to view all open tasks) is rightfully called "Open", however, it is also used as an verb (like in the Export plugin).
This is problematic for languages that has distinct words for those two situations (such as in german: "Offen" for the former one, and "Öffnen" for the latter one).
Because the latter is the more "usual" case, the name of the view names get an special gettext message context.

Note that this means translators would have to re-check the translation as it would be marked as fuzzy by gettext.
From very little testing (as I am not very familiar with the gettext tools), it did that for me, but also correctly separated the "Open" from the export plugin and the "Open" from the task view name in the main window.

### DIFF
--- a/GTG/gtk/data/main_window.ui
+++ b/GTG/gtk/data/main_window.ui
@@ -480,7 +480,7 @@
                       </object>
                       <packing>
                         <property name="name">open_view</property>
-                        <property name="title" translatable="yes">Open</property>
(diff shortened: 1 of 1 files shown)


diegogangl:
Merged, thanks!

PR #372, opened 2020-06-19
Fix opening of the help files

leio:
webbrowser python module just opens things in the browser for me, which
doesn't understand what a "help:gtg" is and routes it back to external
program dialog if lucky.
Use Gtk.show_uri instead, which will properly open it in help handler (yelp).

I believe we should be using Gtk.show_uri instead of webbrowser in other
places as well, but that's something for a different release.


diegogangl:
Merged, thanks!

Issue #392, opened 2020-07-05
Use glib functions to handle some XDG-related functions and drop the pyxdg dependency


nekohayo:
As the title says, we currently depend on pyxdg for some things but this functionality is reportedly available directly in glib nowadays, so we could just use that through gobject introspection and drop the pyxdg dependency.

Probably easy enough, so up for grabs for anyone who wants to attempt this.

PR #396, opened 2020-07-08
Replace XDG related functions

izidormatusov:
Fix for #392

### DIFF
--- a/GTG/backends/unmaintained/backend_rtm.py
+++ b/GTG/backends/unmaintained/backend_rtm.py
@@ -24,12 +24,12 @@ from functools import reduce
 import html
 import datetime
 import os
-import subprocess
 import threading
 import time
 import uuid
 
 from dateutil.tz import tzutc, tzlocal
+from gi.repository import Gio
 
 from GTG.backends.backend_signals import BackendSignals
 from GTG.backends.generic_backend import GenericBackend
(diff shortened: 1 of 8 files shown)


GTG/backends/unmaintained/backend_rtm.py
```diff
@@ -562,7 +562,7 @@ def _authenticate(self):
             if not self.token:
                 self.rtm = createRTM(
                     self.PUBLIC_KEY, self.PRIVATE_KEY, self.token)
-                subprocess.Popen(['xdg-open', self.rtm.getAuthURL()])
+                webbrowser.open(self.rtm.getAuthURL())
```
leio, line 565: Use Gtk.show_uri instead?

This is just a comment on this bit, not a full review

diegogangl: This should be `Gio.AppInfo.launch_default_for_uri()` ([link](https://lazka.github.io/pgi-docs/#Gio-2.0/classes/AppInfo.html#Gio.AppInfo.launch_default_for_uri)) so it uses the right portal in the flatpak version.

izidormatusov: Used `Gio.AppInfo.launch_default_for_uri()`

diegogangl:
First commit of 0.5! Thanks @izidormatusov :D

PR #401, opened 2020-07-08
Port XML code to Lxml

### LINKED ISSUE #285: Port to lxml and improve xml-related code
[Lxml](https://lxml.de/) is apparently *40 times* faster than minidom and has a nicer API. We should definitely switch to it. We can also drop all the code to prettify the XML, since lxml takes care of that.

And once we are sticking our hands in the parser/serialize code, we should also do a deep refactor and improve that code. The whole thing is full of "fixme" comments and _that can't be good..._

diegogangl:
This PR adds [lxml](https://lxml.de/) as a dependency and ports all the xml code to use it. I also simplified some stuff (like the first run tasks module). There was also a small versioning code that I removed since I doubt anyone is still using 0.2 or older.

Hit this PR with all your testing power!

closes #285


diegogangl:
Marking this as WIP since I need to write some tests and add it to the flatpak manifest

GTG/core/xml.py
```diff
@@ -0,0 +1,316 @@
...
+    except (IOError, FileNotFoundError):
+        log.error(f'Could not write XML file at {filepath}')
+        create_dirs(filepath)
+
+
+def write_empty_file(filepath: str, root_tag: str) -> None:
+    """Write an empty tasks file."""
+
+    root = etree.Element(root_tag)
+    save_file(filepath, etree.ElementTree(root))
```
johnnybubonic, line 316: So there's two ways you can go with this.

1. This way
2. A formatted string of the stub file (`f'<?xml version="1.0" encoding="UTF-8" ?>\n<{root_tag}...'`)

Option 1 will let you populate with default attributes if you want (as they'd be defined in the schema - #431).
Option 2, if you don't need attributes populated with defaults or anything, will save you some memory footprint since you don't need to instantiate an `lxml.etree._Element` object. Maybe. I'd be curious to see a benchmark between the two.

GTG/gtk/editor/serialize.py
```diff
@@ -230,7 +230,7 @@ def parsexml(self, buf, ite, element):
                         s = buf.get_iter_at_mark(start)
                         e = buf.get_iter_at_mark(end)
                         if n.nodeName == "link":
-                            anchor = n.getAttribute("target")
+                            anchor = n.get("target")
```
johnnybubonic, line 233: My entire life was changed by this, by the way. I always did `elem.attrib.get('attrName')`; I had no idea I could just do `elem.get('attrName`)`. So thanks. ;)

scripts/anonymize_task_file.py
```diff
@@ -87,8 +87,8 @@ def main():
             data_dir = os.path.join(GLib.get_user_data_dir(), "gtg")
             project_filepath = os.path.join(data_dir, "projects.xml")
             dom = parse(project_filepath)
-            xmlproject = dom.getElementsByTagName("backend")[0]
-            xmlfile = str(xmlproject.getAttribute("path"))
+            xmlproject = dom.findall("//backend")[0]
```
johnnybubonic, line 90: Would we want to skip stdlib XML in this as well and replace with LXML? If this is intended to be LXML, I didn't see the modification of the import of `xml.dom.minidom.parse`.

If I missed that and it's an LXML `_Element` or `_ElementTree`, I think you'd want to do one of these instead:

* `dom.xpath('//backend')[0]`
* `dom.find('backend')` (might only work if it's an `_Element`, not an `_ElementTree`)
* `dom.findall('backend')[0]` (might only work if it's an `_Element`, not an `_ElementTree`)

GTG/core/xml.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+# Total amount of backups
+BACKUPS = 7
+
+# Information on whether a backup was used
+backup_used = {}
+
+
+def task_from_element(task, element: etree.Element):
```
johnnybubonic, line 36: If you wanted to be technical, `element: etree._Element` but I believe this should work just fine.

diegogangl: Nah, the IDE picks it up fine

scripts/anonymize_task_file.py
diegogangl: This is one sneaky bug you found :)

GTG/gtk/editor/serialize.py
diegogangl: Glad to hear! :D

GTG/core/xml.py
```diff
@@ -0,0 +1,316 @@
...
+    temp_file = filepath + '__'
+
+    if os.path.exists(filepath):
+        os.rename(filepath, temp_file)
+
+    try:
+        write_xml(filepath, root)
+
+        if os.path.exists(temp_file):
+            os.unlink(temp_file)
```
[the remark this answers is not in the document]
diegogangl: I just copied that from the old code and didn't change it, good catch

GTG/core/xml.py
```diff
@@ -0,0 +1,316 @@
...
+        except etree.XMLSyntaxError as e:
+            log.debug(f'Syntax error in {filepath}. {e}. Trying next.')
+            continue
+
+    if root:
+        return root
+
+    # We couldn't open any file :(
+    else:
+        # Try making a new empty file and open it
```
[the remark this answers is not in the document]
diegogangl: I wonder how useful that would be though. When we have an empty file we fill it with the first run tasks, so we never actually have an empty file.

GTG/core/xml.py
```diff
@@ -0,0 +1,316 @@
...
+        - file.bak.1
+        - .... until BACKUP_NBR
+
+    If file doesn't exist, create a new file."""
+
+    global backup_used
+
+    files = [
+        xml_path,            # Main file
+        xml_path + '__',     # Temp file
```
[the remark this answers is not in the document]
diegogangl: Maybe, but we would need to find some way to keep a reference of that name around so we can try to open it next time

GTG/core/xml.py
```diff
@@ -0,0 +1,316 @@
...
+    backup_file = f"{filename}.bak.{i}"
+
+    return os.path.join(dirname, 'backup', backup_file)
+
+
+def get_xml_tree(filepath: str) -> etree.Element:
+    """Parse XML file at filepath and get tree."""
+
+    with open(filepath, 'rb') as stream:
+        tree = etree.parse(stream)
```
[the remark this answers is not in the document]
diegogangl: We do use `getroot()` though (`datastore.py`, line 221). Maybe we can use Element once we have the new format, since we will have to redo most of the code anyways

diegogangl:
This was merged manually

Issue #407, opened 2020-07-10
Add synchronisation to CalDAV


mildred:
Hello,

I was a user of GTG long ago, and with the new version 0.4, I'd like to use it again. I'm willing to contribute too to the extent that I can afford (with the limited time that I have available). Before using GTG again, I'd like to get working synchronization available using open standards like CalDAV because last time, I just lost all my tasks somewhere in the homedir of an old computer.

I also would like the CalDAV synchronization to be compatible with https://tasks.org which supports many features and arbitrary nesting of tasks.

I'm asking here if there is any leftover of synchronization code left around (I'll look) and what's the best way to proceed.

Thank you.

nekohayo:
Bonjour Mildred,
We didn't delete the old plugins and synch backends, they are simply moved to subfolders to indicate their unmaintained status. Here they are:

* https://github.com/getting-things-gnome/gtg/tree/master/GTG/plugins/unmaintained
* https://github.com/getting-things-gnome/gtg/tree/master/GTG/backends/unmaintained

If you would like to resurrect one of them (or make a derivative for a different sync service) then we will be happy to include your new contributed feature, as long as it works well (and that you will be able to stay around to maintain it, ideally :)

mildred:
I take a simple look and I enabled back the synchronization dialog and registered a caldav backend, seems easy so far and it seems I'll just have to implement the talking to the backend part, so I am pleasantly surprised. There is just an issue around the icons where `data/icons` is not used at all and I worked around by using standard icons.

I'll need some more time to get the backend working and I should be able to maintain it if I know what needs to be done on it.

Thank you for the great work so far

diegogangl:
Hi Mildred, thanks for working on this! 

> There is just an issue around the icons where data/icons is not used at all and I worked around by using standard icons.

I think most of the stuff in data/icons isn't even getting exported to the flatpak. We'll need to revisit this once we are using gresources. Using standard icons is the best way for now.

oyren:
Maybe not the right place, but whatever. I'm planning a setup similar to @mildred, but I'm thinking about whether [EteSync](https://www.etesync.com/) would be the better solution.
Advantage:
- End-to-end encrypted
- [Python API](https://github.com/etesync/pyetesync)
- [Works with Tasks.org](https://tasks.org/docs/etesync_intro.html)
- Desktop Bridge for CalDAV
[...]

PR #419, opened 2020-07-14
Add CalDAV backend

mildred:
Implements #407

Add a CalDAV backend. Add dependency to`caldav` package. Apart from adding the CalDAV backend, this PR:

- enable back the synchronization backends (disabled backends stay disabled for now)
- use generic icons for backends instead of custom icons that were not loaded
- update keyring support to use libsecret in addition to existing methods. This should make it compatible with Flatpak. (see https://people.gnome.org/~dueno/libsecret-guadec.pdf and https://guadec.ubicast.tv/videos/desktop-secrets-management-for-the-future/)

This backend is still a bit experimental yet, I managed to wipe all my TODOs a few time, but hopefully it is no longer the case. Debugging issues should be possible with the debug log lines provided.

### DIFF
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -1,0 +1,831 @@
+# -----------------------------------------------------------------------------
+# Getting Things GNOME! - a personal organizer for the GNOME desktop
+# Copyright (c) 2008-2013 - Lionel Dricot & Bertrand Rousseau
+# Copyright (c) 2020 - Mildred Ki'Lya
+#
+# This program is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free Software
+# Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program.  If not, see <http://www.gnu.org/licenses/>.
+# -----------------------------------------------------------------------------
+
+"""
+Backend for storing/loading tasks in CalDAV Tasks
+"""
+
+from dateutil.tz import tzutc, tzlocal
+import datetime
+import os
+import time
+import uuid
+import threading
+
+import caldav
+from vobject.icalendar import dateTimeToString, stringToDate
+
+from GTG.backends.backend_signals import BackendSignals
+from GTG.backends.generic_backend import GenericBackend
+from GTG.backends.periodic_import_backend import PeriodicImportBackend
+from GTG.backends.sync_engine import SyncEngine, SyncMeme
+from GTG.core.tag import ALLTASKS_TAG
+from GTG.core.task import Task
+from gettext import gettext as _
+from GTG.core.dates import Date
+from GTG.core.interruptible import interruptible
+from GTG.core.logger import log
+from GTG.core.tag import extract_tags_from_text
+
+# Dictionaries to translate GTG tasks in CalDAV ones
+_GTG_TO_CALDAV_STATUS = \
+    {Task.STA_ACTIVE: 'NEEDS-ACTION',
+     Task.STA_DONE: 'COMPLETED',
+     Task.STA_DISMISSED: 'CANCELLED'}
+
+_CALDAV_TO_GTG_STATUS = \
+    {'NEEDS-ACTION': Task.STA_ACTIVE,
+     'IN-PROCESS': Task.STA_ACTIVE,
+     'COMPLETED': Task.STA_DONE,
+     'CANCELLED': Task.STA_DISMISSED}
+
+
+class Backend(PeriodicImportBackend):
+    """
+    CalDAV backend
+    """
+
+    _general_description = {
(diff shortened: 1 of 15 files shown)


mildred:
I believe the code is ready for review. I got it working with my CalDAV server and it seems to work in both directions. Nesting is supported and compatible with my Tasks application on Android.

README.md
```diff
@@ -27,8 +27,8 @@ Later, when you want to update to the latest development version (assuming you a
 
 The main software dependencies are listed in the "requirements.txt" file, which is a file that Python's "PIP" dependency solving tool can use to install everything you need and make things easier. You can install PIP by typing one of the following commands:
 
-    sudo dnf install python3-pip  # On Fedora
-    sudo apt install python3-pip  # On Ubuntu/Debian
+    sudo dnf install python3-pip python3-devel  # On Fedora
+    sudo apt install python3-pip python3-dev # On Ubuntu/Debian
```
diegogangl, line 31: These two changes should be below, in the _Solving dependencies for plugins_  section

GTG/core/keyring.py
```diff
@@ -27,6 +33,31 @@
 from GTG.core.borg import Borg
 from GTG.core.logger import log
 
+class SecretKeyring(Borg):
+    def __init__(self):
+        super().__init__()
+        self._SECRET_SCHEMA = Secret.Schema.new("io.github.getting-things-gnome.gtg.v1",
```
diegogangl, line 39: We use org.gnome.GTG everywhere (as is our app id). Is there any reason to use github here? (Also we will probably migrate to gitlab in the future)

GTG/core/keyring.py
```diff
@@ -27,6 +33,31 @@
 from GTG.core.borg import Borg
 from GTG.core.logger import log
 
+class SecretKeyring(Borg):
```
diegogangl, line 36: It'd be nice to have a docstring for this class explaining what it's for, etc.

GTG/backends/backend_caldav.py
```diff
@@ -0,0 +1,831 @@
...
+                todo.add_tag(tag)
+        # tags to remove
+        for tag in todo_tags.difference(gtg_tags):
+            todo.remove_tag(tag)
+
+
+        try:
+            todo.save(save_new=save_new)
+        except:
+            log.debug(f'_populate_todo todo.instance={todo.instance}')
```
diegogangl, line 442: Are there specific errors that can be caught rather than a bare except? Also the debug could be a bit more informative ("Todo saving failed for instance x" or something like that)

GTG/backends/backend_caldav.py
```diff
@@ -0,0 +1,831 @@
...
+            ev = self._find_event_by_url(url) # network lookup
+            try:
+                parent_url = ev.parent.url
+                uid = ev.instance.vtodo.uid.value
+            except AttributeError:
+                parent_url = None
+                uid = None
+        return parent_url, uid
+
+class Todo():
```
diegogangl, line 651: I know it's obvious but a docstring here to explain what the class represents, technical details, etc. would be nice

README.md
diegogangl: Actually, why are these needed? Shouldn't this be something like `pip install caldav`

requirements.txt
```diff
@@ -7,3 +7,4 @@ mock
 dbus-python
 PyGObject
 pycairo
+caldav
```
diegogangl, line 10: We're not using requirements anymore since we're 100% meson. Feel free to leave this change (since the file will be removed eventually), but it can't be depended on for dependencies.

ezickler:
Hi, i tried these changes as i would really like to have this feature. I first tried compiling the project in gnome builder as flatpak as i got GTG from flathub and thought flatpak to be the goto. 
But that resulted in the sync menu enable with no caldav plugin loaded as there was nothing to add. 
Then i used the README variant of compiling/launching via launch.sh and it works. But i still have an issues with my caldav syncing. My provider uses the Sabre_DAV server and it seams as only pulling task from the server is possible.

mildred:
>
> Then i used the README variant of compiling/launching via launch.sh 
> and it works. But i still have an issues with my caldav syncing. My 
> provider uses the Sabre_DAV server and it seams as only pulling task 
> from the server is possible.
>
Yep, for dev I run using launch.sh.

Try with -d flag, that's the debug logs, also wait a little to let gpg 
time to catch up. I'd be interested in the debug logs and the issue 
(which tasks are not syncing). Warning: the debug logs will probably 
contain your personal tasks in clear.

ezickler:
@mildred I hesitate to post the log because there are to many private information in the raw form and i don't have the time to properly filter it for a public post. 
I checked the logs there were no errors. My provider has a sabredav server and what i noticed is that i specified a new test calendar for gtg but all calendars are synced. so maybe there is an error some where in processing the path correctly.

Format is 
domain.tld/calendar/username/calendarid 
and calendar id get ignored and task from all my calendar ids get synced (pull down to gtg) but no new gtg tasks show up in any calendre on the server or other clients.

I checked with the 2do ios app and added a task with a tag which came through to gtg but still not the other way around.

BeatLink:
Any updates on this?

diegogangl:
Hi @mildred, is this still being worked on?

diegogangl:
Closing since we merged the other caldav backend PR, which was almost complete. Hope this doesn't dissuade you from contributing again sometime :)
Thanks for the work!

Issue #279, opened 2020-04-11
Better XML format [RFC]


diegogangl:
We can (and _should_) improve the file format of the local backend. A more strict structure would help us move everything into the treemodel faster, and avoid many bugs. 

## Current format

For reference this is what a task currently looks like:

```xml
	<task id="1@1" status="Active" tags="" uuid="d1312d77-d42d-4403-b080-ca065dde166f">
		<title>Learn How To Use Subtasks</title>
		<addeddate>2020-04-10T20:48:11</addeddate>
		<modified>2020-04-10T20:37:02</modified>
		<subtask>6000caf7-6197-4d77-a50e-8bd8804c5694</subtask>
		<subtask>7ab5f54c-dc3c-4cae-a968-1361d7269eb6</subtask>
		<subtask>3749c090-0d84-48b5-ad85-941657c3cff9</subtask>
		<content>A &amp;quot;Subtask&amp;quot; is something that you need to do first before being able to accomplish your task. In GTG, the purpose of subtasks is to cut down a task in smaller subtasks that are easier to achieve and to track down.

To insert a subtask in the task description (this window, for instance), begin a line with &amp;quot;-&amp;quot;, then write the subtask title and press Enter.

Try inserting one subtask below. Type &amp;quot;- This is my first subtask!&amp;quot;, for instance, and press Enter:

→   One subtask
→   Another subtask
→   Yet another one



Alternatively, you can also use the &amp;quot;Insert Subtask&amp;quot; button.

Note that subtasks obey to some rules: first, a subtask's due date can never happen after its parent's due date and, second, when you mark a parent task as done, its subtasks will also be marked as done.

And if you are not happy with your current tasks/subtasks organization, you can always change it by drag-and-dropping tasks on each other in the tasks list.
</content>
		<task-remote-ids/>
	</task>

```

Here's another example with the "alternative IDs":

```xml
	<task id="0@1" status="Active" tags="" uuid="1990638b-4f95-4645-af3a-df1b27b35c10">
		<title>Getting Started With GTG</title>
		<addeddate>2020-04-10T20:48:21</addeddate>
		<modified>2020-04-10T20:48:11</modified>
		<subtask>1@1</subtask>
		<subtask>2@1</subtask>
		<subtask>3@1</subtask>
		<subtask>4@1</subtask>
		<subtask>5@1</subtask>
		<subtask>6@1</subtask>
		<subtask>7@1</subtask>
		<subtask>8@1</subtask>
		<content>
[...]

Issue #169, opened 2015-09-24
Periodic / repeating / recurring tasks


ghost:
It would be cool if periodic tasks were implemented in Getting Things GNOME. For example I have many tasks that must be carried out every week. So people like me need a new option for setting the period (1 week). When clicking the "Mark as done" button two things should happen then:
1. The task should be copied in the Closed tasks tab
2. The task's start and end dates should be automatically changed according to the period set (1 week), but the task must otherwise stay in the Active tasks tab

With the current version of Getting Things GNOME instead I have to do the following things:
1. Manually create an identical task and mark it as done (for record keeping)
2. Manually change the start and end dates of the main task without marking it as done

What do you think?

fturco:
Oops, I just found [this](https://wiki.gnome.org/Apps/GTG/blueprints/repeated_tasks) related blueprint on GNOME wiki.

nekohayo:
As a user myself I would love to have this. Someone needs to step forward to implement the feature, patches welcome, otherwise I don't know when/if that will happen.

diegogangl:
Some thoughts on this based on @leio's idea in the chat.

- Only start _or_ due can be set to repeat, but not both at the same time
- We can add this as separate properties to a task in the xml (`repeat-type` and `repeat-interval` or something like that)
- The UI for this should be separated from the start/due fields to avoid confusion
- Also, this means we can auto-set a due or start date but the user can later change it without affecting the repeating tasks
- We can also have a UI to modify how the task repeats without affecting due/start
- When the task is set to done or dismisses, we shift to the next date and make a copy with a `closed` status to keep track
- We should have an option in the menu to "stop repeating" the task
- If the task isn't done by the time it should repeat, we clone the task and remove the repeat in the original. 
- If it gets deleted it stops repeating
- Once we have ported the sidebar to a listview we could add a section or a filter for repeating tasks

zeddo123:
Hi,

I'm almost done with the implementation of repeating tasks.
However, the current blueprint does not say anything about subtasks of repeating tasks. Should they automatically be repeating and have the same repeating period?

diegogangl:
Great news! 

> However, the current blueprint does not say anything about subtasks of repeating tasks. Should they automatically be repeating and have the same repeating period?

Yes. Subtasks are usually steps in a bigger task, so if you have to do the task again you probably need to go through all the steps too.

diegogangl:
[...]

PR #431, opened 2020-07-15
Proposal for schema

johnnybubonic:

## Why?

This, combined with the porting to LXML (#401) (LXML natively supports XSD), will let you do some REALLY nice things programmatically:

## Review

```xml
<!--
  TODO: ...
-->
```

## Outer scope changes

The gtg_tasks.xml, etc. may need to have namespaces added.

```xml
<?xml version="1.0" ?>
<project>
<!-- ... -->
</project>
```

it'd be:

```xml
<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xmlns="https://wiki.gnome.org/Apps/GTG?tasks"
         xsi:schemaLocation="http://SOMEDOMAIN.TLD/SOME/PATH/TO/tasks.xsd">
<!-- ... -->
</project>
```

This can be done programmatically, and I can provide code to convert existing documents over to the namespaced version.

## Implementation

### DIFF
--- a/GTG/backends/unmaintained/backend_rtm.py
+++ b/GTG/backends/unmaintained/backend_rtm.py
@@ -24,12 +24,12 @@ from functools import reduce
 import html
 import datetime
 import os
-import subprocess
 import threading
 import time
 import uuid
 
 from dateutil.tz import tzutc, tzlocal
+from gi.repository import Gio
 
 from GTG.backends.backend_signals import BackendSignals
 from GTG.backends.generic_backend import GenericBackend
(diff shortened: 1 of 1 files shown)


johnnybubonic:
Also mentioning @diegogangl on this since I get the impression he's doing a lot of the LXML porting.

diegogangl:
While I really like the idea of having a schema, wouldn't it be better to wait for the new file format? We would end up doing this twice in the end.

johnnybubonic:
yep, it would definitely be better. heh whoops. :)

any suggestions for issues/PR #'s to watch besides #279 ? i can happily write the schemas (i kind of enjoy it, to be honest), all i need to know is structure and what types of data/valid strings to accept. i can probably pick/make the proper datatypes and container types etc. from there.

diegogangl:
Sure, #169 will also include some changes to the xml to support repeatable tasks.

johnnybubonic:
Closed to switch branches.

PR #437, opened 2020-07-20
Proposal for schema (Redux)

johnnybubonic:
_**NOTE**: This is a total redo of #431 to match the updated example in #279 as it progresses._

---

## Why?

This, combined with the porting to LXML (#401) (LXML natively supports XSD), will let you do some REALLY nice things pro grammatically:

## Review

```xml
<!--
  TODO: ...
-->
```

## Outer scope changes

The `gtg_tasks.xml` (filename subject to change per #279 approval/implementation), etc. will need to have namespaces added.

```xml
<?xml version="1.0" ?>
<gtgData appVersion="0.5" xmlVersion="2">
<!-- ... -->
</gtgData>
```

it'd be:

```xml
<gtgData appVersion="0.5"
         xmlVersion="2"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xmlns="https://wiki.gnome.org/Apps/GTG"
         xsi:schemaLocation="http://SOMEDOMAIN.TLD/SOME/PATH/TO/data.xsd">
<!-- ... -->
</project>
```

This can be done pro grammatically, and I can provide code to convert existing documents over to the namespaced version.

## Implementation

## Further Reading

It is recommended to review #279 in full and take note of the example proposed document format. This PR will attempt to stay in tandem with current proposed format per #279..


johnnybubonic:
I messed this one up proper, whoops! All because I forgot to add `.idea/` to the .gitignore and I was trying to keep it clean. Let's try this again.

PR #438, opened 2020-07-20
Proposal for schema (Redux mk. II)

johnnybubonic:
_**NOTE**: This is a total redo of #431 (and #437, oops - need to stop playing with branches I have open PRs for) to match the updated example in #279 as it progresses._

---

## Why?

This, combined with the porting to LXML (#401) (LXML natively supports XSD), will let you do some REALLY nice things programmatically:

## Review

```xml
<!--
  TODO: ...
-->
```

## Outer scope changes

The `gtg_tasks.xml` (filename subject to change per #279 approval/implementation), etc. will need to have namespaces added.

```xml
<?xml version="1.0" ?>
<gtgData appVersion="0.5" xmlVersion="2">
<!-- ... -->
</gtgData>
```

it'd be:

```xml
<gtgData appVersion="0.5"
         xmlVersion="2"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xmlns="https://wiki.gnome.org/Apps/GTG"
         xsi:schemaLocation="http://SOMEDOMAIN.TLD/SOME/PATH/TO/data.xsd">
<!-- ... -->
</project>
```

This can be done programmatically, and I can provide code to convert existing documents over to the namespaced version.

## Implementation

## Further Reading

It is recommended to review #279 in full and take note of the example proposed document format. This PR will attempt to stay in tandem with current proposed format per #279, and will exit draft stage and enter a review/pulling stage once the specification in #279 hits master.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ flatpak/cache
 flatpak/.flatpak-builder
 flatpak/bundles
 /.local_build/
+.idea/
 
 # src/ usually has development version of liblarch
 src/
--- a/data/schema/_elements.xsd
+++ b/data/schema/_elements.xsd
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+<xs:schema targetNamespace="https://wiki.gnome.org/Apps/GTG"
+           xmlns="https://wiki.gnome.org/Apps/GTG"
+           xmlns:gtgData="https://wiki.gnome.org/Apps/GTG"
+           xmlns:xs="http://www.w3.org/2001/XMLSchema"
            elementFormDefault="qualified"
            attributeFormDefault="unqualified">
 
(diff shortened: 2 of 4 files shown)


johnnybubonic:
@diegogangl the schemas should now match the current proposed format for #279, with the EXCEPTION of still hashing out how to handle `<content>`.

if you'd like to validate yourself, you can do it without lxml- just make sure libxml2 is installed (which it almost assuredly is if you're using lxml), and do:

```bash
cd /tmp
git clone --single-branch --branch schema https://github.com/johnnybubonic/gtg.git r00t_gtg
cd r00t_gtg/data/schema
xmllint --noout --schema data.xsd /path/to/your/proposed/format.xml
```

The validation errors can be a little confusing, but they're pretty easily googled. If you get any, let me know. The only one I think you'll get is it choking on `<content>` unless you either escape the chars or enclose it in a CDATA.

diegogangl:
A couple of things to change here:

- The icon is now a string with a single character (an emoji). 
- There's been some changes to the recurrent tasks stuff

I can send you a `gtg_data.xml` file if you want

johnnybubonic:
@diegogangl yeah, please send an updated `gtg_data.xml`! sorry for the delay.

diegogangl:
And of course I never sent you an updated xml -_-

Here it is:
https://gist.github.com/diegogangl/8f08a660c9453f49f4602d8bb36ceab1

pentux-GitHub:
Reviewing dormant PRs while 0.7 wraps up. This one aged better than most: gtg_data.xml is still the current format and Diego was feeding this thread with an updated sample. A schema would also pair nicely with the data integrity ideas discussed in #735. @johnnybubonic if you feel like picking it back up, a fresh sample file from current master is easy to provide. Otherwise, maintainers, would 0.8 be the right shelf so the work is not lost?

PR #440, opened 2020-07-24
Fix Untouched Task timer upper limit

### LINKED ISSUE #412: Untouched tasks: Set timer to something over 100 days?
Hello.

I wanted to use the Untouched Tasks plugin, and configure it to tag tasks after 6 months, which translates to roughly 180 days. I was surprised to see that after I type in 180 days and shift focus in the UI, it immediately changes it to 100 days.

Is there some reason behind this limitation, or was it an arbitrary choice? Would it be possible to extend the timer?

Thanks.

clobrano:
Current upper limit in Untouched Task timer is 100 days, which seems
arbitrary and might confuse users.

Since the Entry widget for the timer is limited to show up to 3 digits,
it seems more reasonable to set this upper limit to 999 days.

Closes #412


diegogangl:
Merged, thanks!

PR #442, opened 2020-07-29
Save and restore task view in the main window

### LINKED ISSUE #398: Remember last view on closing/opening
At first, thanks for this great relaunching.

I've installed the flatpak 0.4 release to replace my old fedora gtg i've been keeping since ages, and I'm surprised that GTG 0.4 dosen't remember which view I was when I closed windows.
I mean I used actionable view (formerly known as workview, I think) by default until then, but when I close GTG 0.4 and reopen it, it always start on open view.
Is it a bug  in my setup, or is it like that for everybody ?
(in my workflow, GTG starts with gnome login, that's why I'd like it to be at first a reminder of what is actionable)

Neui:
The config entry existed, but from what I can tell it is never used (read nor written to).
This uses the same convention as get_selected_pane(), because it looks nice when peeking at the config.

Thus, this could have been done simpler by just using the "internal" child name of the GtkStack.
If that is desired, I can change that easily.

Fixes #398

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -41,6 +41,13 @@ from GTG.gtk.tag_completion import TagCompletion
 from GTG.core.dates import Date
 from GTG.core.logger import log
 
+PANE_STACK_NAMES_MAP = {
+    'closed_view': 'closed',
+    'open_view': 'active',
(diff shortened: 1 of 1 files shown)


diegogangl:
Great work, thanks!

PR #456, opened 2020-09-05
Export plugin: Remove useless settings dialog

Neui:
There is an settings dialog to set whenever it should include the menu button to export be visible or not.
However, it is basically useless, since if an user would hide it, it then can't interact with the plugin any more. It is basically deactivating the entire plugin, at least from what I can see.
The more "invisible" setting of the last template used should be continued to be stored and retrieved as usual.

Later on, this commit could be reverted and changed to add the preferences when needed.


diegogangl:
Sounds good, thanks!

PR #457, opened 2020-09-05
Translatable plugin file (.gtg-plugin) strings

Neui:
Make the `.gtg-plugin` files use the [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys), so that  [GLib Key-value file parser](https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html) can be used to pull localized strings.
It technically breaks the format since the GLib parser doesn't support most of pythons ConfigParser features.
This is unfortunate, but this was the easiest way to get this working, especially with gettext integration.

The probably biggest hack would be introducing `.gtg-plugin.desktop` files, which are just used so that xgettext (which is run by `ninja gtg-pot`) uses the Desktop-language on them, so the appear in the .pot and .po files correctly.

A current issue is that when translation is being updated, a re-launch (via launch.sh) doesn't regenerate the files. To force regenerating the files, run `rm -f .local_build/GTG/plugins/*.gtg-plugin`.

It has been tested and works (at least for german, the translation is temporary for testing):
Screenshot of the translated plugins list
(Plugins in order are: Untouched Tasks; Export and print; Urgency color; Send per email)

Note that I had to comment out `GTG/core/cleanxml.py` and `GTG/core/taskxml.py` lines in `po/POTFILES.in` to make it work; This change isn't in this PR, but I'll let you know anyway when you try this PR.

### DIFF
--- a/GTG/core/plugins/engine.py
+++ b/GTG/core/plugins/engine.py
@@ -17,7 +17,7 @@
 # -----------------------------------------------------------------------------
 import imp
 import os
-import configparser
+from gi.repository import GLib
 
 from GTG.core.dirs import PLUGIN_DIRS
 from GTG.core.borg import Borg
@@ -147,9 +147,10 @@ class PluginEngine(Borg):
             for f in os.listdir(path):
                 info_file = os.path.join(path, f)
                 if os.path.isfile(info_file) and f.endswith('.gtg-plugin'):
-                    info = configparser.ConfigParser()
-                    info.read(info_file)
-                    info = dict(info.items("GTG Plugin", True))
+                    parser = GLib.KeyFile.new()
+                    parser.load_from_file(info_file, GLib.KeyFileFlags.NONE)
+                    keys = parser.get_keys("GTG Plugin")[0] # The list of keys
+                    info = {key: parser.get_locale_string("GTG Plugin", key, None) for key in keys}
                     p = Plugin(info, PLUGIN_DIRS)
                     self.plugins[p.module_name] = p
 
--- a/GTG/plugins/export.gtg-plugin
+++ b/GTG/plugins/export.gtg-plugin
@@ -2,9 +2,7 @@
 module=export
 name=Export and print
 short-description=Exports the tasks in the current view into a variety of formats.
-description=Exports the tasks in the current view into
-    a variety of formats. You can also personalize the format
-    of your tasks by writing your own template
+description=Exports the tasks in the current view into a variety of formats. You can also personalize the format of your tasks by writing your own template.
 authors=Luca Invernizzi <invernizzi.l@gmail.com>, Izidor Matušov <izidor.matusov@gmail.com>
 version=0.2
 enabled=False
--- a/GTG/plugins/export.gtg-plugin.desktop
+++ b/GTG/plugins/export.gtg-plugin.desktop
@@ -1,0 +1,1 @@
+export.gtg-plugin
--- a/GTG/plugins/hamster.gtg-plugin
+++ b/GTG/plugins/hamster.gtg-plugin
@@ -1,9 +1,9 @@
 [GTG Plugin]
-Module=hamster
-Name=Hamster Time Tracker Integration
-Short-description=Track time of GTG task with Hamster applet.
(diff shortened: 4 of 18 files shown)


diegogangl:
So we can't make these translatable without adding a .desktop to them?

> Note that I had to comment out GTG/core/cleanxml.py and GTG/core/taskxml.py lines in po/POTFILES.in to make it work; This change isn't in this PR, but I'll let you know anyway when you try this PR.

`cleanxml` was renamed to xml, and `taskxml` was merged into it during the port to lxml. So it would be better to just remove them. `xml.py` doesn't have any user facing strings so it doesn't need to be there either.

Neui:
> So we can't make these translatable without adding a .desktop to them?

I haven't found an easy way to do that with the "built-in" meson stuff. From what I can remember, POTFILES is fed directly to xgettext and it tires to detect based on the file extension, with no way to customize it on a per-file/extension basis. There is something for XML, but it is for XML, not "Desktop".

diegogangl:
What concerns me about this PR is how it will work with third party plugins. We want to have some way for users to install and remove plugins, so we don't have to mainatain all of them. How would that work here? It seems like third party plugins wouldn't be able to include multiple translations (in say a zip). Wouldn't it be easier to support multiple strings with a language identificator in the gtg-plugin file (kind of like .desktop files)?

Neui:
> Wouldn't it be easier to support multiple strings with a language identificator in the gtg-plugin file (kind of like .desktop files)?

This is exactly what this PR introduces. The generated .gtg-plugin files would have `name[de_DE]=translated` or similar, and the [`GLib.KeyFile.get_locale_string()`](https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html#g-key-file-get-locale-string) method gets the translated value for a key.

diegogangl:
Ah I see, then I misunderstood this PR. Could you rebase on current master (the hamster plugin came back to life)?

Neui:
> Ah I see, then I misunderstood this PR. Could you rebase on current master (the hamster plugin came back to life)?

Done, also added the `hamster.gtg-plugin.desktop` symlink.

Not tested, but an potential problem I saw is that the hamster gtg-plugin file uses uppercase in the key names, while the others use all lowercase, and the `GLib.KeyFile` class works with case-sensitive keys and groups.
I fixed this in a new commit if it is OK.

diegogangl:
I tried this with the french locale (`LANG=fr_FR ./launch.sh -d`) but got a glib error:


```
  File "/home/januz/code/gtg/.local_build/install/lib/python3.8/site-packages/GTG/gtk/application.py", line 111, in do_startup
    self.plugins_dialog = PluginsDialog(self.req)
  File "/home/januz/code/gtg/.local_build/install/lib/python3.8/site-packages/GTG/gtk/plugins.py", line 137, in __init__
    self.pengine = PluginEngine()
  File "/home/januz/code/gtg/.local_build/install/lib/python3.8/site-packages/GTG/core/plugins/engine.py", line 152, in __init__
    keys = parser.get_keys("GTG Plugin")[0] # The list of keys
gi.repository.GLib.Error: g-key-file-error-quark: Le fichier de clés n’a pas de groupe « GTG Plugin » (4)
  File "/home/januz/code/gtg/.local_build/install/lib/python3.8/site-packages/GTG/gtk/application.py", line 121, in do_activate
    self.init_shared()
  File "/home/januz/code/gtg/.local_build/install/lib/python3.8/site-packages/GTG/gtk/application.py", line 158, in init_shared
    self.init_actions()
  File "/home/januz/code/gtg/.local_build/install/lib/python3.8/site-packages/GTG/gtk/application.py", line 242, in init_actions
    self.plugins_dialog.dialog.insert_action_group('app', self)
AttributeError: 'NoneType' object has no attribute 'dialog'
```

Neui:
Very weird. Works on my machine™. I asked about it in the chat, it seems for some reason, at least on @diegogangl side, the `*.gtg-plugin.desktop` files are normal files that just contain the correct filename rather than being a symlink, and thus there is of course no section called `[GTG Plugin]`.

I am not sure what is causing that. I would point my finger to git, but I am not sure.

Issue #352, opened 2020-06-05
Ctrl+Delete should be mapped only within the main window as a keyboard shortcut, not in the Task Editor


nekohayo:
I was wondering why I was always triggering the Delete task action in the git version... well it seems the Ctrl+Delete shortcut activates even if you do it within the Task Editor, not just within the main window (task browser).

Now why the heck would I want to use Ctrl+Delete in an editor? Because that's the standard way to delete the word that is after the text cursor (the opposite of Ctrl+Backspace)...

nekohayo:
Also should not be active when the quick add entry is focused in the main window

diegogangl:
The main reason I picked `ctrl+del` instead of `del` was to avoid conflicts like this!
I guess there's no other way but to remove accels on focus out and re-add them on focus in (in the mainwindow)

nekohayo:
Yup... Pitivi deals with the same problems when focusing/defocusing the timeline, it enables/disables a bunch of accels (I spent some quality time making that work back then ;)

flavin:
Hi @nekohayo I create this PR to handle this issue, could you take a look?
https://github.com/getting-things-gnome/gtg/pull/460

thanks in advance for your time!

nekohayo:
Thank you for fixing this (and many other issues) @flavin, and thank you @diegogangl for reviewing and merging :)

PR #460, opened 2020-09-14
Disabled "ctrl+del" on text focus #352 

flavin:
Disable delete task shortcut when focus is in quick entry or task editor.
why someone could want to use `Ctrl+Delete` in an editor? Because that's the standard way to delete the word that is after the text cursor (the opposite of Ctrl+Backspace)

To do it we add a signal to catch event on focus in and focus out to remove `ctrl+del` accelerator or add it respectively.

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -48,6 +48,7 @@ PANE_STACK_NAMES_MAP = {
 }
 PANE_STACK_NAMES_MAP_INVERTED = {v: k for k, v in PANE_STACK_NAMES_MAP.items()}
 
+
 class MainWindow(Gtk.ApplicationWindow):
     """ The UI for browsing open and closed tasks,
     and listing tags in a tree """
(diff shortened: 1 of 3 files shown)


GTG/gtk/browser/main_window.py
```diff
@@ -735,6 +735,16 @@ def focus_quickentry(self, action, param):
 
         self.quickadd_entry.grab_focus()
 
+    def on_quickadd_focus_in(self, widget, event):
+        self.toggle_delete_accel(False)
+
+    def on_quickadd_focus_out(self, widget, event):
+        self.toggle_delete_accel(True)
+
+    def toggle_delete_accel(self, enable_delete_accel):
```
diegogangl, line 744: Could you add a docstring here?

diegogangl (commented):
LGTM, only missing one docstring that I would consider important. The other thing is that commit messages have summaries longer than 50 characters.  Could you keep the first line to a single sentence of 50  (or less)? The rest of the message (the body) can be 72-characters long.

Also please add "closes #352" in a new line at the end of the message in the first commit? That way the ticket gets closed automatically.

Thanks!

flavin:
thank you @diegogangl  do you mean make the pull request comment no longer than 50? 
because I don't know how to make commit smaller, I only know just creating a new branch new commits new pull request.

diegogangl:
Merged, thanks!

PR #462, opened 2020-09-21
Hungarian translation

gerazo:


po/hu.po
```diff
@@ -1619,24 +1830,30 @@ msgid ""
...
+"Részfeladatok:\n"
+"%s"
 
 #: GTG/plugins/send_email/sendEmail.py:61
 #, python-format
 msgid ""
 "\n"
 "Task content:\n"
-"%s"
+""
```
Neui, line 1842: Changed source string which you shouldn't do.
https://github.com/getting-things-gnome/gtg/blob/27b04d3c86a7217ccae712ee0b7fa9891f12858c/GTG/plugins/send_email/sendEmail.py#L61
Note that the translated version also lacks the "%s" part.

Neui (commented):
Found a small mistake, easy to fix. 
I don't speak Hungarian, so I can't check the translation itself.

po/hu.po
gerazo: Huhh. Thanks. Nice catch.

gerazo: Fixed. (Not other similar was found.)

diegogangl:
@gerazo How do you want to move forward with this? Do you have someone who knows Hungarian that you could invite to review? Or should I go ahead and merge?

Files look good

kristofking007:
I checked the translation, it looks fine to me. ( Hungarian is my first language. )

YuKaiTe:
I just checked too, translation looks good. (Also native Hungarian speaker)

diegogangl:
Merged, thanks guys!

gerazo:
Thanks a lot for everyone.

PR #471, opened 2020-09-25
Use _("...").format instead of _(f"...")

### LINKED ISSUE #380: Python f-strings can't be really translated
Take a look at:

https://github.com/getting-things-gnome/gtg/blob/abe2a9110dd0fc6a46f2d095013972877ea67d78/GTG/gtk/browser/delete_task.py#L109

What happens is that while gettext correctly finds the `"\nAnd {missing_titles_count:d} more tasks"` string, however at run time, Python formats the string first (like `"\nAnd 4 more tasks"`) before feeding it into gettext, and because there isn't explicit translations for each number, the string is left untranslated.

There is an  [StackOverfow question about this](https://stackoverflow.com/questions/49797658/how-to-use-gettext-with-python-3-6-f-strings), and there are 2 solutions:
* Don't use Pythons `f`-strings, use the `format()` method
* Use an wrapper method (`f(_("\nAnd {missing_titles_count:d} more tasks"))`), but seems somewhat hacky

There might be other strings that have the same faith, such as (seems gettext mark those strings as `python-brace-format`, so it is somewhat easy to find, or run `grep -r "_(f"`):

https://github.com/getting-things-gnome/gtg/blob/abe2a9110dd0fc6a46f2d095013972877ea67d78/GTG/gtk/backends/__init__.py#L68

(Only found those two)

Neui:
The latter would format the string first before translating, and having an translation for every possible combination is basically impossible.

I am using .format() specifically because diegogangl commented about having to drop it for .format(), so I am using that.

I tried to not modify the original string to not break existing translations, but I only was half-successful.

Fixes #380
Couldn't test the string about Synchronization Services because there aren't any (active) plugins for that.

### DIFF
--- a/GTG/gtk/backends/__init__.py
+++ b/GTG/gtk/backends/__init__.py
@@ -65,7 +65,8 @@ class BackendsDialog():
         builder = Gtk.Builder()
         self._load_widgets_from_builder(builder)
         # Load and setup other widgets
-        self.dialog.set_title(_(f"Synchronization Services - {info.NAME}"))
+        dialog_title = _("Synchronization Services - {name}")
+        self.dialog.set_title(dialog_title.format(name=info.NAME))
         self._create_widgets_for_add_panel()
         self._create_widgets_for_conf_panel()
         self._setup_signal_connections(builder)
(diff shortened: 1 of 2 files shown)


diegogangl:
Thanks, yeah I guess changing it without breaking strings can't be done if they were already translated with f-formatted string

PR #474, opened 2020-09-28
Make module-level translations work

### LINKED ISSUE #362: Some plugin related strings aren't translatable

### LINKED ISSUE #378: Search commands don't use translated versions
The search commands (such as `!now`) are translatable, but the translated versions aren't use.
For example, in the german version, "now" is translated to "jetzt".
Then create some tasks with the due date be Now(/Jetzt).
After that, open search, using `!now` works, using `!jetzt` doesn't.

Looking at the code, it looks like it has the same faith as #362 and #357: Translation is done at the high level scope

https://github.com/getting-things-gnome/gtg/blob/43b5374715a46438ecd13ae6cdd607380f3f54a5/GTG/core/search.py#L85-L97

### LINKED ISSUE #379: Quick Add doesn't use translated translated versions for date parsing (such as due:x)
These commands set the due date correctly to `now`: (`fällig` is `due` in german here)
* `some due:now`
* `some due:Now`
* `some fällig:now`
* `some fällig:Now`

But not these: (`jetzt` is `now` in german here)
* `some due:jetzt`
* `some due:Jetzt`
* `some fällig:jetzt`
* `some fällig:Jetzt`

Looks very similar to #357 : Translation is done at the high level scope, before gettext initalized

https://github.com/getting-things-gnome/gtg/blob/abe2a9110dd0fc6a46f2d095013972877ea67d78/GTG/core/dates.py#L50-L60

Neui:
I saw this being done in D-Feet:
https://gitlab.gnome.org/GNOME/d-feet/-/blob/b312c714fc01c3d14e2c6e1da6e87645674cca05/src/d-feet.in#L38-41

The problem is that some code uses gettext "at the module level", which is run when the module is being imported.
Before this change, basically all of GTG gets imported, before locale gets set up.
After this change, it gets setup before importing it, so it'll work as expected. Not sure why I haven't thought of this earlier.

I've put this in a if `__name__ == "__main__"` block, because the original code was also in such block.

Fixes #362
Fixes #378
Fixes #379

Haven't tested every string, but I did more of the for me easier tests and it works fine as expected.
Before-After

### DIFF
--- a/GTG/gtg.in
+++ b/GTG/gtg.in
@@ -50,6 +50,20 @@ import gettext
 import locale
 import signal
 
+if __name__ == "__main__":
+    # Set up UI i18n so module-level translations work
+    LOCALE_DIR = '@localedir@'
+
+    try:
+        locale.bindtextdomain('gtg', LOCALE_DIR)
+        locale.textdomain('gtg')
+    except AttributeError as e:
+        # Python built without gettext support doesn't have bindtextdomain() and textdomain()
+        print("Couldn't bind the gettext translation domain. Some translations won't work.\n{}".format(e))
+
+    gettext.bindtextdomain('gtg', LOCALE_DIR)
+    gettext.textdomain('gtg')
+
 import gi
 gi.require_version('Gdk', '3.0')
 gi.require_version('Gtk', '3.0')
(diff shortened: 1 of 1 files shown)


diegogangl:
LGTM, Thanks!

PR #475, opened 2020-09-30
Recurring/Repeating tasks through the TaskEditor

### LINKED ISSUE #169: Periodic / repeating / recurring tasks

zeddo123:
In the previous pull request, we have merged the essential logic for recurring tasks as well as a way to add them with the Quick Add Entry. With this new pull request, users will be able to enable/disable repeating task with the GUI ( a new context menu in the task view and a popover menu in the TaskEditor).

**Using the TaskEditor**
A menu button is added with a circular icon that would open a popover to let the user choose to enable and set a recurring term. Inside it, there is a label that will display the current recurring term. Furthermore, from the TaskEditor the icon will light blue whenever the task is recurring.
94424693-8beeee80-0182-11eb-9c12-ab5be07e1ba0

**Using the TaskView**
Although the context menu is less flexible, it provides a simpler, quicker way to enable recurrency in the task. If a task is recurring, a Unicode icon will be appended to the title.
94364359-8f21a600-00c0-11eb-8e5e-7865b3e8ed9c

**Backend update**
We have added the option of a task that starts recurring from the start date. ref 8dd2ba1

close #169
This work could not be done without the help of @Maistrox9 and @diegogangl.


diegogangl:
Hey, this is awesome :)
I probably won't have time until tomorrow to sit down and review this, but some things I've noticed from the commits:

- Some commits from the previous branch are in this one. Those are already on master, so they are blocking the merge
- Can the "merge into" commits be removed?
- Can the "new recurrence" commits be squashed?
- Capitalize the words instead of using uppercase for commit 47736b6 
- Add a "close #169" to the last commit or the PR message

zeddo123:
I apologize, I am kind of a new-be when dealing with rebasing.

> Some commits from the previous branch are in this one. Those are already on master, so they are blocking the merge

Sorry, I should have deleted and created a new fork, when I started working on the GUI. I thought I fixed the conflict, should I rebase the branch? can it be done easily?

> Can the "merge into" commits be removed?

Can they be removed without a problem?

diegogangl:
> Sorry, I should have deleted and created a new fork, when I started working on the GUI. I thought I fixed the conflict, should I rebase the branch? can it be done easily?

No problem, I've pulled this branch so if something goes wrong I can force push and bring it back.
You can just drop those commit in this branch and then force push. I don't know if there any GUI apps that can do this, but my favorite tool for this kind of things is the interactive rebase. 

https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History
https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d

Just add a `d` in front of the commits to remove them. Also remove the last "merge master into..." commit.

Either 4fbb0ab or 42bb01c should be removed too. They seem to be changing the same thing.

> Can they be removed without a problem?

I think so.  16b1a7b  is a simple merge commit and doesn't have any changes, so it can be removed. Not too sure about the other two, but try it doing it before the other changes. If anything goes wrong you can fetch and pull the branch back.

zeddo123:
> https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History
https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d

Thank you for the links, they were very helpful.

Everything should be set correctly now. Would you like me to squash the commits even more? and should I resolve the conflicts?

diegogangl:
Yes please fix the conflicts. Though I'm not sure why there are conflicts? Most of these files haven't been touched in master in a while. Make sure to pull from master and then rebase.

Also, this section looks like code from the previous PR? 
https://github.com/getting-things-gnome/gtg/pull/475/files#diff-61031fc2ea1504c730c5cd63411b896dR326

zeddo123:
> Also, this section looks like code from the previous PR?
https://github.com/getting-things-gnome/gtg/pull/475/files#diff-61031fc2ea1504c730c5cd63411b896dR326

Here we found an 'issue'. When the user, set the recurrency every week, it would set the due date in 7 days, not the current day. So we changed the offset to 0 when a task is newly created.

diegogangl:
Here's the problem with the conflicts. All these commits are already on master:

Screenshot from 2020-10-03 20-56-26

When git switches to this branch, it tries to re-apply all those commits on top of master. Which fails because they can't be applied twice. Try interactive rebasing again and drop them all, then push --force again here. You may end up having to drop the last two "resolve conflict" commits too.

zeddo123:
Hi,
I dropped those commits and rebased as you have suggested, but I always seem to have conflicts when I want to merge with upstream. Would you like me to create another fork with the same commits so that we could avoid littering the commit history?

Thank you.
I apologize again for wasting your time with such a simple issue. I should have forked again.

diegogangl:
Hi,
No problem. I learned how to rebase by destroying branches, you're already doing better :) 
I gave it a quick try and it seems to be throwing conflicts on the last commit ("Resolve Conflicts"). Maybe dropping it would fix the conflicts.

But if you want to be the safest you can start a new branch off master, and `git cherry-pick` each commit from this branch until you hit the issue.

zeddo123:
Hi,
I tried cherry-picking the commits but in the end I rebased the recurring-gui branch against master (which was in sync with upstream). Now we could expect it to have no conflicts.

diegogangl:
Yes, no more conflicts \o/

Some notes:

- The repeat checkbox is always visible but if you haven't set a recurring term you can't enable it. This is kinda confusing, I think it would be better if it was hidden initially, or it had some kind of default (toggle the checkbox without a set term and it auto-sets daily for instance).
- The icon in the task browser looks kind of squished
Screenshot from 2020-10-07 20-03-53

- If I set a day for week/year/month I get the full date (eg, "Every 18 september") but if I use the "On this day" button I only get "weekly/monthly/yearly". Is there a special reason for that?
- The yearly text should be something like "September 18", that way we can avoid ordinals in english

GTG/gtk/editor/recurring_menu.py
```diff
@@ -0,0 +1,88 @@
...
+                    formated_term = datetime.strptime(f"{formated_term[:2:]}-{formated_term[2::]}", '%m-%d').strftime('%d %B')
+            self.title.show()
+            self.title_separator.show()
+            self.title.set_markup(f"{RecurringMenu.PREFIX}<b>{formated_term}</b>")
+        else:
+            self.title.hide()
+            self.title_separator.hide()
+
+    
+    def reset_stack(self):
```
diegogangl, line 85: A docstring here would be useful

GTG/gtk/editor/recurring_menu.py
```diff
@@ -0,0 +1,88 @@
...
+        else:
+            self.icon_style.remove_class('recurring-active')
+
+    def is_term_set(self):
+        return self.selected_recurring_term is not None
+
+    def set_selected_term(self, string):
+        self.selected_recurring_term = string
+
+    def update_tick(self):
```
diegogangl, line 40: Does tick refer to the checkbox? I'd rather use the same terms Gtk uses, like `update_checkbox()`

GTG/gtk/editor/recurring_menu.py
```diff
@@ -0,0 +1,88 @@
+from datetime import datetime
+class RecurringMenu():
+    """ RecurringMenu provides a simple layer of abstraction
+    for the menu where the user enables a task to be repeating
+    """
+    PREFIX = "Every "
```
diegogangl, line 6: Mark this as translatable

GTG/gtk/editor/recurring_menu.py
```diff
@@ -0,0 +1,88 @@
...
+            done = True
+        return done
+            
+    def update_header(self):
+        """ Updates the header anytime a term is selected """
+        formated_term = self.selected_recurring_term
+        if self.is_term_set():
+            if formated_term.isdigit():
+                if len(formated_term) <= 2 :
+                    formated_term = f"{formated_term} of the Month"
```
diegogangl, line 74: Use format() for this string and make it translatable. F-strings can't be translated, see:
https://github.com/getting-things-gnome/gtg/pull/471/files

GTG/gtk/data/style.css
```diff
@@ -32,3 +32,7 @@
 .dark .prefs_list row {
     border-bottom-color: alpha(rgb(0, 0, 0), 0.2);
 }
+
+.recurring-active {
+		color: rgb(52, 130, 224);
+}
```
diegogangl, line 38: nitpick: match the indent of the other lines

diegogangl (changes requested):

GTG/core/task.py
```diff
@@ -344,53 +350,90 @@ def recursive_sync(self):
...
+                # If a start date is already set,
+                # we should calculate the next date from that day.
+                if self.start_date == Date.no_date():
+                    start_from = Date(convert_datetime_to_date(date.today()))
+                else:
+                    start_from = self.start_date
+
+                newdate = start_from.parse_from_date(recurring_term, newtask)
+                return (True, newdate)
+            except Exception as e:
```
diegogangl, line 388: Do we know which exception might be thrown here?

diegogangl:
Looking good, some more notes:

- Setting a task with subtasks to repeat propagates the repeat to all subtasks, but disabling (with the checkbox) doesn't disable it in the subtasks
- Lets try to keep one change per commit so we have a cleaner history and it's easier to make changelogs later (and rolling back changes, bisecting, etc.)
- Also try to keep the summary of commit messages (the first line) at 50 chars and the body (lines after that) at 72. Most GUIs and terminal applications, including github expect this format. You can read more about it here: https://chris.beams.io/posts/git-commit/#limit-50

GTG/core/task.py
zeddo123: Yeah, that would be a `ValueError` from the raised from the date object when parsing the date in the a String format.
Should I explicitly put it in the exception?

zeddo123:
> Setting a task with subtasks to repeat propagates the repeat to all subtasks, but disabling (with the checkbox) doesn't disable it in the subtasks

In this case, it is tricky. Could a 'normal' task have a repeating task? in this case what should we do? assume that when a task goes from repeating to non-repeating we disable all its children or we let it be a normal task with repeating task inside it?

Either way it should be simple to modify(although in the first case, it would not be pretty--code wise).

GTG/core/task.py
diegogangl: Yeah, it would be better to catch a explicit exception

zeddo123:
> Would this be too complicated code wise?

No, no problem actually. I had to remove the call for `inherit_recursion` for `set_recurring`. (I would not consider to be very clear as I wanted a 'centralized' way to propagate recurrence.)

Anyway, it should work as expect for now on. 

on thing though, if we mark a non-repeating task to done and it has a recurring child, the child will duplicated either way. The only way to fix this in my opinion is to add a flag to the `set_status` method to know if the task is being marked as done due to propagation or not. As currently we cannot know why the task is marked as done and we need this functionality to know whether we duplicate ( next occurrence) or we mark it as done definitely.

I hope I was clear, if not I'll try to make an example.

diegogangl:
> one thing though, if we mark a non-repeating task as done and it has a recurring child, the child will be duplicated either way. The only way to fix this in my opinion is to add a flag to the set_status method to know if the task is being marked as done due to propagation or not. As currently we cannot know why the task is marked as done and we need this functionality to know whether we duplicate ( next occurrence) or we mark it as done definitely.

> I hope I was clear, if not I'll try to make an example.

Yeah, I could see it while testing. Kind of sucks to add more complexity on `set_status` but this seems like the only way for now. Please go ahead. I haven't noted anything else to fix/change, so this would be the last thing

diegogangl:
And it's merged! Thank you guys so much for all this work :grinning:

Issue #277, opened 2020-04-08
Fix Dbus wrapper


diegogangl:
GTG used to expose some of its functionality to dbus using some custom code in the dbus wrapper class (`dbus.py`). Nowadays, GtkApplication registers itself with the same name as the app id which conflicts with the old way, since both end up with the same bus name.

My **"** solution **"** was to change the well known name in the dbus wrapper. So now GTG has two dbus interfaces. It would be a shame to throw away the old one since it was nicer and more specific to gtg. A real solution would be to add those functions to the well know name exposed by GtkApplication.

It would be great if someone with dbus knowledge could help out here!

nekohayo:
Mart suggested that we should eliminate the dbus-python dependency by using glib/gio GDBus stuff via introspection to achieve the same thing. I guess this should be done as part of this ticket, so renaming the title (in case anyone out there is looking at this ticket and wants to tackle it).

diegogangl:
I managed to port the timer to gio dbus (which is the only place we use dbus ATM). Unfortunately I can't get it to attach a callback to the proper signal. It seems to be a problem with the system bus, I can attach callbacks to signals in the session bus just fine but when it comes to the system bus it just... doesn't do it. There are no errors or anything.

Haven't had any luck asking in #gtk, #python and Gnome Discourse. Leaving this code here in case someone knows the proper incantation:

```python

        system_bus = Gio.bus_get_sync(Gio.BusType.SYSTEM, None)
        system_bus.signal_subscribe('org.freedesktop.login1',
                                    'org.freedesktop.login1.Manager',
                                    'PrepareForSleep',
                                    '/org/freedesktop/login1',
                                    None,
                                    Gio.DBusSignalFlags.NONE,
                                    on_prepare_for_sleep,
                                    None)
```

The callback:

```python

def on_prepare_for_sleep(connection, sender_name, object_path,
                         interface_name, signal_name, parameters,
                         *user_data):
    """Handle dbus prepare for sleep signal."""

    print('Called')
```

Neui:
@diegogangl 
Adding the following code after calling `signal_subscribe` I receive the signals:

```python
        from gi.repository import GLib$
        GLib.MainLoop().run()$
```
However, as you can see, this will stop application flow, in other words GTG won't continue executing.
[...]

PR #479, opened 2020-10-02
Migrate Timer from dbus-python to Gio & drop dbus-python

Neui:
For #277

Yes, I figured the timer system bus signal not being executed thingy out. Finally!
See the commit message for more details:

> Initially, it didn't work. The reason was that system_dbus object is created on that line and when it goes out of scope, it gets destroyed and the signals get disconnected.
> 
> I assumed that the connection would stay open as part of Gio and not be created on first use (althrough it makes sense when implementing it). Also, I assumed that the signal would make it stay connected.
> 
> This took days to figure out, I even wrote a more "minimal" python version, and even a C version which let me find out the issue. See https://gitlab.gnome.org/-/snippets/1288
> 
> Thus the fix is to just keep a reference to it in the timer object itself.

### DIFF
--- a/GTG/core/plugins/engine.py
+++ b/GTG/core/plugins/engine.py
@@ -19,8 +19,6 @@ import imp
 import os
 import configparser
 
-import dbus
-
 from GTG.core.dirs import PLUGIN_DIRS
 from GTG.core.borg import Borg
 from GTG.core.logger import log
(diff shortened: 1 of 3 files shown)


diegogangl:
Ha! I knew it had to be something damn simple. Thanks and kudos on cracking this :D

Issue #85, opened 2015-01-02
Replace the limited set of tag icons by emojis


tkdchen:
It would be a nice feature to allow to specify a custom icon for a tag.

vrutkovs:
This is already supported in 0.3.x - see 'Click to set icon' button
![tag editor](https://lh5.googleusercontent.com/-U9XmCzB--0Q/UJmV9wPkxII/AAAAAAAAAFM/byDjXKqhcvc/w800-h800/gtg-0.3-tag-editor.png)

tkdchen:
@vrutkovs  cannot select other icons from file system except the fixed set of icon listed by "Click To Set Icon"

vrutkovs:
@tkdchen ah, yes, that would be a nice feature. At least updating a current list of available icons would be nice

izidormatusov:
This is outside of the scope of 0.4 milestone that we would release after summer.

nekohayo:
I would like to see the oldschool icons support replaced by GTK3's emojis support. It would be much nicer-looking, more portable, and reduce the amount of stuff we need to ship with the app. Patches welcome!

nekohayo:
A tiny nitpick for @diegogangl in commit 18cfd501b:
* The 10% opaque default label icon is very hard to see in a light theme?
* The "Remove icon" button label is not marked as translatable, I think?

PR #480, opened 2020-10-03
Don't print errors for the same icon multiple times

Neui:
Previously, when getting an icon error while rendering tasks (such as when the user changes icon themes), it won't find the icon and prints an error.

This is fine, but it spams the terminal/console/logs with essential useless stuff, thus it is made sure only to print it once (per renderer).

I know about that possibly being dropped (see #85), but while it happens, I'd like to make it less spammy for me.


diegogangl:
Merged! Yeah, we will move away from it eventually but at this point I'm not sure when.

PR #481, opened 2020-10-03
Move from application constructor to starup signal

### LINKED ISSUE #183: GTG is not DBusActivable
https://wiki.gnome.org/HowDoI/DBusApplicationLaunching

Neui:
Previously, a bit of stuff was done in the constructor, which (likely) hurts being D-Bus activate-able, since the "remote" instance (aka when trying to execute GTG again which then would hand off to the existing instance) would re-initialize only to not use that stuff.

Also changed from using an list-comprehension `[x for y in z]` to an "normal" for-loop, since it is two lines anyway and the result isn't used.

~~One step closer to being DBus-Activatable (#183).~~

Edit: Now also fixes #183

### DIFF
--- a/GTG/gtk/application.py
+++ b/GTG/gtk/application.py
@@ -80,11 +80,19 @@ class Application(Gtk.Application):
         super().__init__(application_id=app_id,
                          flags=Gio.ApplicationFlags.HANDLES_OPEN)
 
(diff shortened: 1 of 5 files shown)


Neui:
Also added commit that makes GTG be dbus-activatable, so I don't have to create an separate PR just for one commit.

PR #483, opened 2020-10-08
set application id as logo icon in about window

### LINKED ISSUE #389: App logo icon is not shown in the about dialog
I previously thought this was just a quirky bug from the fact that I was using launch.sh in the git version, but this problem also happens when running a flatpak package, so it seems like there's an actual bug here:

Screenshot from 2020-07-04 22-33-46

flavin:
fix #389


flavin:
hello can review this PR @diegogangl  @nekohayo ?
is to use the the applicationid for example in flatpack will use org.gnome.gtg

diegogangl:
LGTM, thanks!

PR #484, opened 2020-10-09
Reintroduce GTG tasks DBus interface with Gio

### LINKED ISSUE #277: Fix Dbus wrapper

Neui:
This is **Work In Progress**! Fixes #277. View the commits for some more information.

Some stuff is missing:
- [ ] Properly implement `ModifyTasks`, ``
- [ ] Better error handling
- [ ] Check about performance with many tasks, possibly outsource it to an separate thread
- [ ] Test it
- [ ] Document the interface introspection XML `data/dbus/org.gnome.GTG.Tasks.xml`

Decisions:
* Possibly offer pagination so client don't pull thousands of tasks at once?
* Possibly use an tuple for more common fields (title, text, dates, status, and the dictionary for future stuff)

### DIFF
--- a/GTG/core/dbus/dbus.py
+++ b/GTG/core/dbus/dbus.py
@@ -1,0 +1,149 @@
+from gi.repository import GLib, Gio
+from GTG.core.logger import log
+import os
+import traceback
+
+
+def _get_installed_xml_from_interface_name(interface_name):
+    """Return an DBus interface XML from system installed folders"""
+    raw_paths = GLib.get_system_data_dirs()
+    raw_paths.append(GLib.get_user_data_dir())
+    for path in [os.path.join(p, 'dbus-1', 'interfaces') for p in raw_paths]:
+        full_path = os.path.join(path, interface_name + ".xml")
+        try:
+            with open(full_path, 'rt') as f:
+                log.debug(f"Found file {full_path}")
+                return f.read()
+        except OSError:
+            pass
+    return None
+
+def _get_internal_xml_from_interface_name(interface_name):
+    """Return an DBus interface XML from internal data folders"""
+    log.debug("TODO: Implement this later")
+    pass
+
+def get_xml_from_interface_name(interface_name, use_internal=True, use_system=True):
+    """Return an DBus interface XML from either provided by the system (distro)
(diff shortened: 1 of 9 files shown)


diegogangl:
Can I leave a wish here? It'd be nice to have a dbus interface to add tasks with the quick add syntax (just passing a string basically). That way we can implement a mini-app to quickly add tasks from a global shortcut (I could swear there was an issue about that but I can't find it now)

gwutz:
In order to reactivate the GNote plugin i think its necessary to have a dedicated Bus method to add a single Task withouth knowing the resulting ID. I would love to reactivate this interaction as it fits perfectly the flow "Taking Notes -> abstracting tasks from these notes into GTG"

khurshid-alam:
@diegogangl 
I was looking for to use gtg plugin for kupfer launcher with newer gtg where tasks can be seached, opened and created. It used to work with older version. But found out debus interface doesn't exists any more. This is how it was it looked. 

image

Is this the correct branch?  Any way it can be revived with old methods ?

nekohayo:
Is this directly related / a prerequisite to #676?

In any case, needs to be rebased for 0.7.

Neui:
> Is this directly related / a prerequisite to #676?

No.

pentux-GitHub:
Reviewing dormant PRs while 0.7 wraps up. The demand in this thread is clearly still alive (GNote plugin, kupfer launcher, quick-add wish above). The branch predates the new core though, so it needs a port to the new stores rather than a rebase, and 0.7 is feature-frozen at this point. Would it make sense to attach this to the 0.8 milestone so the need is not lost? No pressure on anyone to pick it up now.

PR #487, opened 2020-10-18
using another unicode for recurring tasks.

flavin:
this is super small change, but in my ubuntu I haven't installed or changing the default fonts, 

and using `\u2B6E` I'm seeing this:
image

so I try another similar unicode code `\u21BB` and it works for me
image

what do you think @diegogangl @zeddo123

### DIFF
--- a/GTG/gtk/browser/treeview_factory.py
+++ b/GTG/gtk/browser/treeview_factory.py
@@ -95,7 +95,7 @@ class TreeviewFactory():
         str_format = "%s"
 
         # We add the indicator when task is repeating
-        INDICATOR = "\u2B6E "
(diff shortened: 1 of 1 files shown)


zeddo123:
Looks good to me. I'll verify the new unicode on Manjaro to be sure.

diegogangl:
> this is super small change, but in my ubuntu I haven't installed or changing the default fonts,

Yup, but Ubuntu sets their own font by default instead of Cantarell. 

Tested this change with Cantarell (Fedora/Vanilla Gnome) and Open Sans + Droid Sans (Elementary). 
Should be enough until we can move to listboxes and draw proper icons

Thanks!

PR #489, opened 2020-10-19
Fix wrong offsets for recurring tasks

zeddo123:
I was looking through the code and found that the offsets where not correct for `every other day` and `every month`.


diegogangl:
Merged, thanks!

PR #493, opened 2020-10-19
Better recurring popover titles

Neui:
I don't know where they are used, but they appear in the translations, so better give them proper titles.


diegogangl:
These aren't user visible (AFAIK) but it's an improvement anyways :+1:

PR #494, opened 2020-10-19
Update translation doc (with plugins and additional option)

Neui:
See commit messages

The plugin workaround is already known, see #457, not entirely sure why this is happening. (I would think it uses the symlink last updated which doesn't update when the file it points to updates)


diegogangl:
Thanks!

Issue #492, opened 2020-10-19
Plugin dialog: Hamster unclear "dbus" dependency (& stacks)


Neui:
The [Hamster plugin only uses the `dbus-dependencies` key](https://github.com/getting-things-gnome/gtg/blob/144814c16723fa9d00e17e047df5d79ab443fc5f/GTG/plugins/hamster.gtg-plugin):
https://github.com/getting-things-gnome/gtg/blob/144814c16723fa9d00e17e047df5d79ab443fc5f/GTG/plugins/hamster.gtg-plugin#L8

However, it'll just show that it needs the "dbus" python module:
![2 "dbus" dependencies](https://user-images.githubusercontent.com/1196130/96504104-830aad80-1254-11eb-8149-7c51ee925b74.png)

Bonus: Upon opening the plugin window multiple times, it'll stack:
![3 "dbus" dependencies](https://user-images.githubusercontent.com/1196130/96504137-93228d00-1254-11eb-88ae-e6595d818393.png)

flavin:
Hi @Neui & @diegogangl 
I'm not sure what is the proper way to solve this.

to avoid the duplication I suspect we need just to add the `dependency` of dbus in the config file
this will avoid to enter in the else here and appending a new one.
https://github.com/getting-things-gnome/gtg/blob/master/GTG/core/plugins/engine.py#L112-L116

and why probably the plugin didn't depend on dbus, because it was a project module removed here:
https://github.com/getting-things-gnome/gtg/commit/bedcae9acc6466e527fc6909d764a9e01b722f5c#diff-75d233079e760bf9aa1474d733dd045a2d996d45efcfb4538d79cd9b7698bbec

also dbus missing dependencies is not being used, not sure what should we do with `dbus-dependencies` 
https://github.com/getting-things-gnome/gtg/commit/931427b1d9f1d00c3b7ee6210802b7f9b01349c1

flavin:
@Neui can you check this PR https://github.com/getting-things-gnome/gtg/pull/497 to avoid appending once and once again dbus

PR #497, opened 2020-10-25
adding dbus as dependency to hamster plugin.

flavin:
ref #492

### DIFF
--- a/GTG/plugins/hamster.gtg-plugin
+++ b/GTG/plugins/hamster.gtg-plugin
@@ -5,5 +5,6 @@ short-description=Track time of GTG task with Hamster applet.
 description=Adds the ability to send a task to the Hamster time tracking applet
 authors=Kevin Mehall <km@kevinmehall.net>, Francisco Lavin <fcolavin@gmail.com>
 version=0.3
+dependencies=dbus
 dbus-dependencies=org.gnome.Hamster:/org/gnome/Hamster
 enabled=False


Neui:
AFAIK @diegogangl wanted to drop dbus-python in favour of the Gio(/GLib) DBus implementation, like the [project page suggests](https://dbus.freedesktop.org/doc/dbus-python/).
Since this is an plugin, this might not be too "critical", but maybe keep in mind that the plugin maybe could be migrated to Gio.

> @Neui can you check this PR #497 to avoid appending once and once again dbus

I can confirm. This would fix #492.
(But I can't activate it (checkbox just doesn't check, and there is an user-unfriendly traceback about not being able to find the hamster dbus thing (reasonable since I don't use it)) and trying to access the plugin settings also throws an traceback; But I would consider out of scope for this PR)

diegogangl:
Hi @flavin, thanks for jumping on this. I'm sorta in favor of merging this. Honestly the best solution is to port to Gio.dbus and get rid of pydbus altogether. But if it's too difficult, we could merge this and release 0.5 as it is.

flavin:
@Neui @diegogangl  it makes all the sense migrate plugins too to Gio, and apart of migrate plugins to use it gio.dbus, we also will need something like this without pydbus https://github.com/getting-things-gnome/gtg/commit/931427b1d9f1d00c3b7ee6210802b7f9b01349c1 or give the responsability to plugin to handle their dbus dependency

diegogangl:
Yeah, we could bring that method back using Gio. We just need to remember to poke holes in the sandbox for each dbus interface we need to access, otherwise that method will always fail under flatpak. 
(which reminds me I need to poke a hole for Hamster)

diegogangl:
Let's go ahead and merge this PR as it is, so we can get 0.5 out soon. We can do the port afterwards :+1:

Issue #490, opened 2020-10-19
Can't translate current recurring state


Neui:
After setting a task to be recurring, the current state can't be translated (and is "ugly" in the sense that it says "other-day" instead of like "other day").

To allow flexibility for other languages, maybe use more specific strings such as "Every other day", "Every {day}. of the Month", "Every {weekday}", ...)

![It says "every other-day"](https://user-images.githubusercontent.com/1196130/96499693-083e9400-124e-11eb-8ff3-4ee60c362097.png) ![It says "Every 09 of the Month"](https://user-images.githubusercontent.com/1196130/96500092-a3d00480-124e-11eb-985f-6f7891b0465e.png)

diegogangl:
Seems like we need to move the "Every" prefix into the strings themselves, and build a special string for other-day. What do you think @zeddo123 ?

zeddo123:
> (and is "ugly" in the sense that it says "other-day" instead of like "other day")

Yeah, I know. It is not even easy to write but I had to make it that way as it would not be parsed correctly in the quick add entry otherwise.

>To allow flexibility for other languages, maybe use more specific strings such as "Every other day", "Every {day}. of the Month", "Every {weekday}", ...)

We are currently doing similar formatting here. Could you point me to the problem? (I don't know how translating works).
https://github.com/getting-things-gnome/gtg/blob/024029129c267fbaa82a74660e1ed7cb2d7b98ba/GTG/gtk/editor/recurring_menu.py#L93
We could make the method broader for every keyword, such as 'other day', because know we are only using it for the week, month and year.

> Seems like we need to move the "Every" prefix into the strings themselves

Sorry, but I don't understand how that would help. Is having the prefix and the suffix separate makes it harder and not flexible to translate?
Wouldn't that make the parsing 'harder' in `dates.py`?

Neui:
> > (and is "ugly" in the sense that it says "other-day" instead of like "other day")
> 
> Yeah, I know. It is not even easy to write but I had to make it that way as it would not be parsed correctly in the quick add entry otherwise.

Except it is used when displaying it like in the screenshots above, it shouldn't affect parsing code.

> > To allow flexibility for other languages, maybe use more specific strings such as "Every other day", "Every {day}. of the Month", "Every {weekday}", ...)
> 
> We are currently doing similar formatting here. Could you point me to the problem? (I don't know how translating works).
> 
[...]

PR #498, opened 2020-10-25
Fix current recurring state translation issue

Maistrox9:
ref #490

### DIFF
--- a/GTG/gtk/editor/recurring_menu.py
+++ b/GTG/gtk/editor/recurring_menu.py
@@ -24,8 +24,6 @@ class RecurringMenu():
        for the menu where the user enables a task to be repeating
     """
 
-    PREFIX = _('Every ')
-
     def __init__(self, requester, tid, builder):
         # General attributes
         self.task = requester.get_task(tid)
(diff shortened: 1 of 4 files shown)


GTG/gtk/editor/recurring_menu.py
```diff
@@ -92,25 +90,26 @@ def update_task(self, enable=True):
...
-            elif formated_term == 'month':
-                formated_term = _('{day} of the Month')
-                formated_term = formated_term.format(day=datetime.today().strftime('%d'))
-            elif formated_term == 'year':
-                formated_term = datetime.today().strftime('%B %d')
+            if self.selected_recurring_term.isdigit():
+                if len(self.selected_recurring_term) <= 2 : # Recurring monthly from selected date
+                    self.title.set_markup(_('Every <b>{month_day} of the month</b>').format(month_day=datetime.strptime(f'{self.selected_recurring_term}', '%d').strftime('%d')))
+                else: # Recurring yearly from selected date
+                    self.title.set_markup(_('Every <b>{year_date}</b>').format(year_date=datetime.strptime(f'{self.selected_recurring_term[:2:]}-{self.selected_recurring_term[2::]}', '%m-%d').strftime('%B %d')))
```
Neui, line 98: I would do something like `_('Every <b>{month} {day}</b>')` since in german, we usually write the date in "day month year" rather than "month day year", and the current solution doesn't allow to change the order.

GTG/gtk/editor/recurring_menu.py
```diff
@@ -92,25 +90,26 @@ def update_task(self, enable=True):
...
+            elif self.selected_recurring_term == 'other-day': # Recurring every other day
+                self.title.set_markup(_('Every <b>other day</b>'))
+            elif self.selected_recurring_term == 'week': # Recurring weekly from today
+                self.title.set_markup(_('Every <b>{week_day}</b>').format(week_day=datetime.today().strftime('%A')))
+            elif self.selected_recurring_term == 'month': # Recurring monthly from today
+                self.title.set_markup(_('Every <b>{month_day} of the month</b>').format(month_day=datetime.today().strftime('%d')))
+            elif self.selected_recurring_term == 'year': # Recurring yearly from today
+                self.title.set_markup(_('Every <b>{year_date}</b>').format(year_date=datetime.today().strftime('%B %d')))
+            else: # Recurring weekly from selected week day
+                self.title.set_markup(_('Every <b>{week_day}</b>').format(week_day=self.selected_recurring_term))
```
Neui, line 110: From my testing, if you set a week (e. g. "Friday") and then switch to german (with my currently not released updated translation), it'll then say "Jeden Friday" instead of "Jeden Freitag", because it seems it stores an once localized value rather an language neutral value and translating it later.
This doesn't seem to be good practice in my opinion.
An workaround could be `_(self.selected_recurring_term)`, but doesn't fix when switching from non-english language to an another language.

Neui (changes requested):

GTG/gtk/editor/recurring_menu.py
diegogangl: Shouldn't day and month names be also translated? Like `_(datetime.today().strftime('%A'))`. 

---

For self.selected_recurring_term, here's what we can do:
1. Set a widget name for each week day and month. The widget name should be the english word. So for instance, the widget that has an id "mon_button" would have a name "Monday". The input for the widget name can be found in glade, in the _Common_ tab.

2. In the callback function (`set_recurring_term_week_day(self, widget)`) you get the widget. So you can do: `widget.name' to the recurring term to store. 

What do you think?
This way we make sure recurring term is always saved in english, and we can do `_(self.selected_recurring_term)`

Neui: > Shouldn't day and month names be also translated? Like `_(datetime.today().strftime('%A'))`.

No, [from the strftime documentation](https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior):
> | Directive | Meaning | Example |
> | -- | -- | -- |
> | `%A` | Weekday as locale’s full name. | Sunday, Monday, …, Saturday (en_US); Sonntag, Montag, …, Samstag (de_DE) |

> ---
> For self.selected_recurring_term, here's what we can do:
> 
> 1. Set a widget name for each week day and month. The widget name should be the english word. So for instance, the widget that has an id "mon_button" would have a name "Monday". The input for the widget name can be found in glade, in the _Common_ tab.
> 2. In the callback function (`set_recurring_term_week_day(self, widget)`) you get the widget. So you can do: `widget.name' to the recurring term to store.
> 
> 
> What do you think?
> This way we make sure recurring term is always saved in english, and we can do `_(self.selected_recurring_term)`

I assume this is just for the week days. I also didn't really examined how it determines on how to set it, but it sounds an possible solution, it seems you don't "translate" the widget name.
The only worry I would have is that the (de-)serialization code would make sure that for example it doesn't get set to the lowercase version (like "monday" instead of "Monday") or something.

Maistrox9:
Hi @Neui and @diegogangl, I hope this last commit solved the issue.

Neui (approved):
Works for me now

diegogangl:
LGTM too, thanks!

PR #501, opened 2020-10-28
Some micro optimizations

Neui:
A while ago in September I tried to optimize the startup with many tasks currently unsuccessfully. In the meanwhile I did those micro optimizations, they shouldn't change much. See the commits for more info.

I haven't really tested them since them, but they are relatively simple enough.

BTW, in `dates.py`, at the file level, `FUNCS` is initialized with some `datetime` objects with being relative to "today", but AFAIK the value isn't updated after a new day. Not sure if this is significant (or where this is used), but I'll just point this out here.
```python
FUNCS = {
    NOW: datetime.date.today(),
    SOON: datetime.date.today() + datetime.timedelta(15),
    SOMEDAY: datetime.date.max,
    NODATE: datetime.date.max - datetime.timedelta(1),
}
```

### DIFF
--- a/GTG/gtk/browser/treeview_factory.py
+++ b/GTG/gtk/browser/treeview_factory.py
@@ -152,16 +152,31 @@ class TreeviewFactory():
             return ""
 
     def sort_by_startdate(self, task1, task2, order):
-        sort = self.__date_comp(task1, task2, 'start', order)
-        return sort
+        t1 = task1.get_start_date()
+        t2 = task2.get_start_date()
+        return self.__date_comp_continue(task1, task2, order, t1, t2)
 
     def sort_by_duedate(self, task1, task2, order):
-        sort = self.__date_comp(task1, task2, 'due', order)
-        return sort
+        t1 = task1.get_urgent_date()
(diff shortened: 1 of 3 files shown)


GTG/gtk/browser/treeview_factory.py
```diff
@@ -152,16 +152,33 @@ def get_task_closeddate_column_string(self, node):
             return ""
 
     def sort_by_startdate(self, task1, task2, order):
+        t1 = task1.get_start_date()
+        t2 = task2.get_start_date()
+        return self.__date_comp_continue(task1, task2, order, t1, t2)
         sort = self.__date_comp(task1, task2, 'start', order)
         return sort
```
diegogangl, line 159: Left an inaccessible return here

diegogangl (changes requested):
Nice, seems to have made some things more simple too

GTG/gtk/browser/treeview_factory.py
Neui: Fixed

diegogangl (approved):

diegogangl:
Landed!

PR #502, opened 2020-10-29
Add translator notes primarily around parsing code

Neui:
To better signify what those strings and their restrictions are.

In `GTG/core/task.py` I had outsourced 2 lists to an new variable so it isn't that crammed in there to make "proper" comments.

Shouldn't make any functionality changes.


diegogangl:
Merged, thanks!

PR #504, opened 2020-10-29
Update meson to 0.51.0

Neui:
Because I was getting the following warning:

    WARNING: Project targeting '>= 0.50.0' but tried to use feature introduced in '0.51.0': args arg in i18n.merge_file

Because I used the args argument in #457 to set what keys gettext should translate.


diegogangl:
Thanks for taking care of this :+1:

Issue #441, opened 2020-07-25
"3 per day" visual progress indication plugin


nekohayo:
An idea I've been entertaining for a little while, a bit related to #435 but simpler and more "visual": a plugin that counts the number of tasks accomplished in the current day, and shows a thin visual progress indicator just under the headerbar (above the treeviews?). Obviously if you use GtkProgressbar this would only work properly with themes like Adwaita that have super-thin GTK progressbar widgets... but there's another widget that might be better-suited for this: [GTK3's LevelBar](https://developer.gnome.org/gtk3/stable/GtkLevelBar.html)! It has the added bonus that it changes colors depending on the value. 

The target amount of tasks to accomplish per day would be a user-defined target (whether it's 3 tasks, or 5 tasks, for example, if you're a follower of the [rule of 3](https://www.artofmanliness.com/articles/work-deliberately-instead-reactively-rule-3/) or some other similar trick), maybe even with potentially a super sophisticated system that distinguishes between "big tasks"—a.k.a. "[eat the frog](https://medium.com/@alltopstartups/eat-the-frogs-first-thing-in-the-morning-and-other-better-work-habits-7070f9e79822)!"—identified by a particular tag, for those who want to combine that with the 3-5 per day rule, into for example the "[1-3-5 rule](https://www.themuse.com/advice/a-better-todo-list-the-135-rule)").

I'm not sure it would make a _huge_ difference in productivity for "everybody", but I suspect that the adrenaline kick of seeing that progressbar/gauge "fill up" like a combo bar, and to be reminded constantly that you are "making progress" in your day's objectives, could help at least some users keep the "momentum" and trick themselves into doing "just one more task...!" ; it also kinda lets you see visually if you're ahead or behind your expected average amount of goals for the day.

Issue #435, opened 2020-07-19
Gameification (motivational) features and virtual assistant


nekohayo:
I haven't got a completely clear idea of what I'd want here, but this is a ticket so that people can help brainstorm (and more importantly, implement) a potential "[gameification](https://en.wikipedia.org/wiki/Gamification)" (aka "motivational" or "achievements") plugin. It would also depend on #424. It could also be somewhat related to the "Untouched tasks" plugin and "Hamster" plugin, too.

The idea is simple: GTG could encourage/reward the user for accomplishing tasks, by congratulating them and unlocking badges of honor (a.k.a. "achievements" or "ranks" or "titles") or something, like you find in many games today. Maybe with standard (but low-priority/ephemeral) freedesktop **notifications** (and if we want to be extra obnoxious/funny, an optional sound to accompany it), and maybe with a dashboard UI where you can see your achievements/titles/trophies. Heck, it could even use humor in the notifications for some particular corner cases, for example "Double-kill", "Multi-kill", "M-M-M-MONSTER KILL" (for oldschool Unreal Tournament fans, when accomplishing a bunch of tasks at once or in rapid succession within seconds/minutes) or "C-C-C-C-C-COMBO BREAKER" (for oldschool K.I. fans, when dismissing or deleting a bunch of tasks at once).

Of course that would totally be plugin territory, because it's not a critical feature to have, and maybe even some/many users (boring adults) would find that ridiculous :)

I guess the plugin could be looking at:

* Certain numbers of closed tasks "of all time"
* Numbers of closed tasks in particular time periods (in a day, in a week, in a month), aka velocity. Normally, most people can only accomplish 1-3 things per day, unless they're really small things (see [this](https://www.borisgloger.com/blog/2016/04/14/one-task-per-day-lets-start-an-experiment/) and [that](https://www.artofmanliness.com/articles/work-deliberately-instead-reactively-rule-3/) for example)
* Accomplishing a very long-standing task or project (creation vs closed date)
* Accomplishing a task or project that you've procrastinated too much on (i.e. that you used the "Do it tomorrow" button too often on!); but that would require saving/storing new data in the data format, a counter of how many times a task has been deferred...
* Maintaining or increasing velocity compared to the last week or month or something
* Accomplishing a big project (i.e. a task with _lots_ of subtasks); see also #245
[...]

PR #510, opened 2020-11-10
Gamify, a plugin to gamify your GTG workflow

zeddo123:
Hi,
I have implemented the three main feature that I think are most important for this kind of plugins.

### Features
1. Count the number of tasks done per day.
2. Count the number of days in which the user has completed the goal
3. Added a score and level.
* message and emoji are displayed depending on the number of tasks done.
* The scores are calculated depending on a tag
  * if the task has `@hard` as a tag, the user receives +3 points.
  * if the task has `@meduim` as a tag, the user receives +2 points.
  * if the task has `@easy` as a tag, the user receives +1 point.
* Levels
  * 100 Beginner
  * 1000 Novice
  * 2000 professional
  * 4000 Expert
  * 9000 Master
  * 13000 Master II
  * 19000 Grand Master
  * 25000 Productivity Lord

### Preferences
For the moment, the user can only change one thing which is the goal number of tasks per day.

### Plugin API
I had to add a couple of method to the plugin API so that I get the headerbar.

### Some screenshots
[No task marked as done](https://user-images.githubusercontent.com/34837535/98710585-234f8000-2384-11eb-8bce-9c57c1710fdd.png)
[one or more tasks are done](https://user-images.githubusercontent.com/34837535/98710733-598cff80-2384-11eb-8e33-d5a0f82af4aa.png)
[achieved daily goal](https://user-images.githubusercontent.com/34837535/98710929-91944280-2384-11eb-96ca-2827e05bd688.png)

### Update
* Added the GtkLevelBar to the plugin, as well as the possibility to activate one or the other.
* Ability to add custom tag mappings from the settings of the plugin

Hope you like it.
ref #435 #441

### DIFF
--- a/GTG/plugins/gamify/gamify.py
+++ b/GTG/plugins/gamify/gamify.py
@@ -156,9 +156,6 @@ class Gamify:
             return 3
         return 1
 
-    def OnTaskOpened(self, plugin_api):
-        pass
-
     def is_full(self):
         """Return True if ui type is FULL"""
         return self.preferences['ui_type'] == 'FULL'
--- a/GTG/plugins/gamify/gamify.py
+++ b/GTG/plugins/gamify/gamify.py
@@ -61,34 +61,21 @@ class Gamify:
         }
         self.builder.connect_signals(SIGNALS) 
 
-    def add_submenu(self):
-        self.menu = self.plugin_api.get_menu()
-        self.stack = self.menu.get_child()
-        self.submenu = self.builder.get_object('submenu-popover')
+    def add_headerbar_button(self):
         self.headerbar_button = self.builder.get_object('gamify-headerbar')
 
-        self.stack.add_named(self.submenu, 'gamify')
         self.headerbar = self.plugin_api.get_header()
         self.headerbar.add(self.headerbar_button)
 
-    def remove_submenu(self):
-        self.stack.remove(self.submenu)
+    def remove_headerbar_button(self):
         self.headerbar.remove(self.headerbar_button)
-
-    def add_menu_enty(self):
-        self.menu_item = self.builder.get_object('gamify-entry')
-        self.plugin_api.add_menu_item(self.menu_item)
-
-    def remove_menu_entry(self):
-        self.plugin_api.remove_menu_item(self.menu_item)
 
     def activate(self, plugin_api):
         self.plugin_api = plugin_api
         self.browser = plugin_api.get_browser()
 
         # Settings up the menu 
-        self.add_submenu()
-        self.add_menu_enty()
+        self.add_headerbar_button()
 
         # Init the preference dialog
         try:
@@ -106,7 +93,6 @@ class Gamify:
         self.update_streak()
         self.analitics_save()
         self.update_widget()
-        print(self.data)
 
     def on_marked_as_done(self, sender, task_id):
         log.debug('a task has been marked as done')
@@ -123,7 +109,6 @@ class Gamify:
         self.data['score'] += self.get_points_for_task(task_id)
         self.analitics_save()
         self.update_widget()
-        print(self.data)
 
     def get_points_for_task(self, task_id):
         """Returns the number of point for doing a perticular task
@@ -213,6 +198,9 @@ class Gamify:
     def get_number_of_tasks(self):
         return self.data['last_task_number']
 
+    def get_streak(self):
+        return self.data['streak']
+
     def update_score(self):
         score_label = self.builder.get_object('score_label')
         score_label.set_markup(_("<b>{current_level}</b>").format(current_level=self.get_current_level()))
@@ -236,7 +224,7 @@ class Gamify:
             emoji = ["\U0001F60E", "\U0001F920", "\U0001F640"]
             headerbar_label.set_markup(random.choice(emoji))
             headerbar_msg.set_markup(_("Good Job!\nYou have achieved your goal."))
-        elif tasks_done > 1:
+        elif tasks_done >= 1:
             emoji = ["\U0001F600", "\U0001F60C"]
             headerbar_label.set_markup(random.choice(emoji))
             headerbar_msg.set_markup(_("Only a few tasks to go!"))
@@ -248,9 +236,20 @@ class Gamify:
         levelbar.set_max_value(self.preferences['target'])
         levelbar.set_value(self.get_number_of_tasks())
 
+    def update_streak_widget(self):
+        streak_number = self.builder.get_object('streak_number')
+        streak_emoji = self.builder.get_object('streak_emoji')
+        if self.get_streak() > 0:
+            streak_emoji.set_markup("\U0001F525")
+        else:
+            streak_emoji.set_markup("\U0001F9CA")
+        streak_number.set_markup(_("You've completed your goal {streak} day in a row.").format(streak=self.get_streak()))
+
+
     def update_widget(self):
         self.update_score()
         self.update_goal()
+        self.update_streak_widget()
 
     def configure_dialog(self, manager_dialog):
         if not self.configureable:
--- a/GTG/plugins/gamify/gamify.py
+++ b/GTG/plugins/gamify/gamify.py
@@ -1,6 +1,7 @@
 import os
 import random
 from datetime import date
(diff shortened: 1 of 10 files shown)


diegogangl:
Hope you don't mind, I'm setting this for 0.6 for now so we can put 0.5 out soon(ish).

GTG/plugins/gamify/gamify.py
```diff
@@ -0,0 +1,361 @@
...
+        if easy:
+            return 1
+        if meduim:
+            return 2
+        if hard:
+            return 3
+        return 1
+
+    def OnTaskOpened(self, plugin_api):
+        pass
```
nekohayo, line 160: Did you want to do something here and changed your mind, or is the plugin API forcing this method to be present?

GTG/plugins/gamify/gamify.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def button_update_streak(self):
+        """Update the streak numbers in the BUTTON widget"""
+        streak_number = self.builder.get_object('streak_number')
+        streak_emoji = self.builder.get_object('streak_emoji')
+        if self.get_streak() > 0:
+            streak_emoji.set_markup("\U0001F525")
+        else:
+            streak_emoji.set_markup("\U0001F9CA")
+        streak_number.set_markup(_("You've completed your goal {streak} day in a row.").format(streak=self.get_streak()))
```
nekohayo, line 293: You might need to use ngettext (if I recall correctly) to handle plurals for the streak here, for example day vs days... unless you cheat and consider it a streak only if `get_streak() > 1`  :)

nekohayo (commented):
I didn't pay close attention to the code, but I can at least say that the plugin seems to work, from some basic testing I've done. It wouldn't shock me to merge and include it for 0.5?

Some additional remarks/questions:
* What is the "analytics" and "preferences" file format? ini? pickle? Might be better to add a filename extension for that
* The plugin doesn't take into account marking tasks as not-done (ex: un-doing/un-dismissing a task), though I don't know if it should / needs to...

zeddo123:
I've added the ability to add custom mappings from the settings, and the possibility for translation.

>What is the "analytics" and "preferences" file format? ini? pickle? Might be better to add a filename extension for that

I'm using the default way to save data of plugins from the , so it should be pickle. Which file extension should I add?

>The plugin doesn't take into account marking tasks as not-done (ex: un-doing/un-dismissing a task), though I don't know if it should / needs to...

We probably should add that feature. However, from the current implementation it won't be perfect as some tag mappings might have been deleted or altered (the deduction of points will be wrong).

>Should the difficulty/points-weighing tags be configurable (or dual-translation-capable like the dates keywords elsewhere in the app)?

That's done now. :)

leio:
A quick drive-by comment on one aspect only:
There appear to be a bunch of new files added, including python source files that have strings marked for translation, but only the plugin metadata file is added to POTFILES.in, therefore nothing else will be translatable (the translatable strings won't be extracted for translators with the i18n tooling).

zeddo123:
> A quick drive-by comment on one aspect only:
> There appear to be a bunch of new files added, including python source files that have strings marked for translation, but only the plugin metadata file is added to POTFILES.in, therefore nothing else will be translatable (the translatable strings won't be extracted for translators with the i18n tooling).

Thanks, I'll look into it :D

zeddo123:
After rebasing my branch on master, I got this message saying that I need the `GTG.core.logger` in the plugins window.
image

*I will verify on my laptop, it could be a problem in my machine*

diegogangl:
The logger module got nuked a while back:
https://github.com/getting-things-gnome/gtg/pull/539

zeddo123:
Forgot about that. okay I'll switch to the new logger.

diegogangl:
I've finally sat down to review this! Code looks good, though it could use some enters to separete lines more (and make it more readable). Specially after docstrings.

Also, congrats on being the first to use `:=` in GTG o/

Something I noted is that the score doesn't update when tasks are marked as not done. A cheater could use this to increase his score without actually doing the tasks.

## Popup

I would add some more padding between the large emoji and text of the first section. Don't be afraid to play with font size to improve the hierarchy of information.

"your current level" can be simplified to "level:" and points can be put below without any other text. Like this:

---

Level: **Beginner**
5 Points

---

The streak can be simplified to a single line like this (emoji and text in the same line):

---

🔥️ **2 day** streak
👻️ First day completing your goals (couldn't find the ice cube emoji for some reason)

---


## Preferences

I dig the fancy title at the top!

- I would move "Select UI Version" to below "Target Tasks per day", and make a listbox to hold both settings so they look like the main preferences. Look at `preferences.ui` for how that's done in the preferences dialog.

- You can set the "Target Tasks per day" value to zero and hit your goal every day forever :)

- Same thing for the tags section. Look at the `preferences.ui` file as well as the `style.css` file (I had to add some classes). Instead of the `Add` and `Delete` buttons, add a new row at the end with a + to add a new tag, and a button to each row to delete that row. Look at the "Region & Language" section of GNOME's settings for an example.

- Watch out for the @ in the text input, so the user doesn't end up adding `@@tag`.

- "Select UI Version" can be simplified to "UI Mode". I also think a [Combo Box](https://lazka.github.io/pgi-docs/index.html#Gtk-3.0/classes/ComboBox.html#example) would be a better fit than the radio buttons.

## Future ideas

- The difficulty tags are cool, but in the future there could be some heuristics to calculate score like number of subtasks, time a task has been open, number of swear words, etc.
- It would be cool to include some messages that encourage people to relax once they hit their goal (for workaholics/people stressed out of their minds)

zeddo123:
I've been working on your proposed enhancement. Just one thing though.

>Something I noted is that the score doesn't update when tasks are marked as not done. A cheater could use this to increase his score without actually doing the tasks.

This brings the problem of negative number of tasks done. Let say, today I only marked a task as not done. if I implement this feature the easy way, you would have a negative number of points. What would you think would be the best thing to do here?

We could either:
* Do it the easy way (we enable points to be negative)
* Only deduce points when the task has been marked as done the same day.

diegogangl:
> This brings the problem of negative number of tasks done. Let say, today I only marked a task as not done. if I implement this feature the easy way, you would have a negative number of points. @diegogangl What would you think would be the best thing to do here?
> 
> We could either:
> 
>     Do it the easy way (we enable points to be negative) or limit it to drop only to 0.
>     Only deduce points when the task has been marked as done the same day.

I think the simplest would be to clamp the score to 0.

zeddo123:
Hey @diegogangl. I've made the necessary changes. Whenever you have time, take a look.

diegogangl:
Hey o/
The popup looks great. I like the new icons you added (specially the star!).
I still have some changes for the preferences:

- Drop the sidebar. Sorry I wasn't clear on this, I meant the general preferences page. We're also going to drop that sidebar in the apps preferences in the future (once we get to libhandy/libadwaita)
- Use `user-trash-symbolic` as the icon for _Delete Tag_ button
- Add the `.prefs_list` css class to the listboxes so they get borders and better padding
- Remove the "close" button, it duplicates the X button
- Push the apply button to the other side of the window

zeddo123:
Yeah it looks much better now! Should I squash all commits into one?

diegogangl:
Looking good!
I would squash the last three commits into one, and leave the rest of the history.

There's still one a couple of things left:
- Make the labels left-aligned
- There's no need to repeat UI in the menu when it's already in the label. I would change "Button UI" to "Button only" and "Full/Both" to "Button and Progress Bar", and "Discrete" to "Progress bar"
- The text in the tags labels should be centered vertically

Last changes! I would also squash this last commit with the other 3

zeddo123:
>The text in the tags labels should be centred vertically

Do you want them to be in the centre of the area, and not left-aligned?

diegogangl:
> > The text in the tags labels should be centred vertically
> 
> Do you want them to be in the centre of the area, and not left-aligned?

Left-aligned horizontally, centered vertically. (I think it's alignment and padding in Glade, 0 for horizontal and 0.5 for vertical)

zeddo123:
done :smile:

diegogangl:
It's been a long road, thanks for the work!

PR #517, opened 2020-11-20
Ability to "Add a parent task" to a task

### LINKED ISSUE #516: Ability to "Add a parent task" to an existing child task by right-clicking in the Task Browser window
Oftentimes I think in reverse order when determining the chain of actions I must take for a project, or change my mind as I go, and so I have:

* Project foo
  * Task A
    * Task B
    * Task C
        * Task D

...I would like to be able to right-click "Task C" and say "Add a parent task", which would then create "Task E" and insert it in-between Task C and its parent. So it would become:

* Project foo
  * Task A
    * Task B
    * **Task E**
      * Task C
          * Task D

zeddo123:
This PR implements the feature discussed in #516 by @nekohayo.

Anytime the user uses the shortcut or the context menu, add new parent task will be added to the selected task _(switch parents if necessary)_ and a task editor will be opened.

I have also chosen the keyboard shortcut of `<ctrl><shift>P`.
close #516

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -990,6 +990,27 @@ class MainWindow(Gtk.ApplicationWindow):
 
             self.app.open_task(task.get_id(), new=True)
 
+    def on_add_parent(self, widget=None):
+        uid = self.get_selected_task()
+        selected_task = self.req.get_task(uid)
+        if uid:
+            parents = selected_task.get_parents()
+            if parents:
+                # Switch parents
+                for p_tid in parents:
+                    par = self.req.get_task(p_tid)
+                    if (par.is_loaded() and par.get_status() in
+                        (Task.STA_ACTIVE)):
+                        new_parent = par.new_subtask()
+                        par.remove_child(uid)
+                        new_parent.add_child(uid)
+            else:
+                # If the tasks has no parent already, no need to switch parents
+                new_parent = self.req.new_task(newtask=True)
+                new_parent.add_child(uid)
+
+            self.app.open_task(new_parent.get_id(), new=True)
+
     def on_edit_active_task(self, widget=None, row=None, col=None):
         tid = self.get_selected_task()
         if tid:
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -928,6 +928,9 @@ class MainWindow(Gtk.ApplicationWindow):
                 else:
                     treeview.set_cursor(path, col, 0)
                 treeview.grab_focus()
+                self.app.action_enabled_changed('add_parent', True)
+                if not self.have_same_parent():
+                    self.app.action_enabled_changed('add_parent', False)
                 self.open_menu.popup_at_pointer(event)
 
             return True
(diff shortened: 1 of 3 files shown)


GTG/gtk/browser/main_window.py
```diff
@@ -990,6 +990,27 @@ def on_add_subtask(self, widget=None):
...
+        uid = self.get_selected_task()
+        selected_task = self.req.get_task(uid)
+        if uid:
+            parents = selected_task.get_parents()
+            if parents:
+                # Switch parents
+                for p_tid in parents:
+                    par = self.req.get_task(p_tid)
+                    if (par.is_loaded() and par.get_status() in
+                        (Task.STA_ACTIVE)):
```
Neui, line 1003: Why not `par.get_status() == Task.STA_ACTIVE`?
Also, from what I understand, it just checks whenever the status is contained in `Task.STA_ACTIVE`, not equal to it (`(Task.STA_ACTIVE) == Task.STA_ACTIVE`, you probably thought of `(Task.STA_ACTIVE,)`, which is a tuple)

zeddo123: Yeah, that should be `par.get_status() == Task.STA_ACTIVE`.

>  you probably thought of (Task.STA_ACTIVE,)

yep
However, I don't know why there are these ifs in the code base (I was just mimicking them)
https://github.com/getting-things-gnome/gtg/blob/d7bbcd5b2a9d7d54b2b28278e5bbf6fd264b302b/GTG/core/task.py#L213
wouldn't be best to have `c.get_status() == self.STA_ACTIVE`

GTG/gtk/browser/main_window.py
```diff
@@ -990,6 +990,26 @@ def on_add_subtask(self, widget=None):
 
             self.app.open_task(task.get_id(), new=True)
 
+    def on_add_parent(self, widget=None):
+        uid = self.get_selected_task()
```
flavin, line 994: @zeddo123  what should do if you select multiple task? 
maybe here we can loop on selected and add those multiple task the same new parent, what do you think?

zeddo123: Yeah, I agree. Having a common parent for all the selected task would be better.

zeddo123: Now that I had time to think about it. Doing it this way is complicated when the selected tasks are not on the same level.

* Task A
   * Task A.1
   * Task A.2 [selected]
   * Task A.3 [selected]
* Task B
   * Task B.1 [selected]
* Task C

In this case, how should the feature behave? It's not clear where the new common parent task will be put.

Neui: I think you could check whenever all the selected tasks have the same parent, so you could just select all Subtasks of A, which would make it more obvious what the result does. Maybe check that when you right click to disable the menu entry as needed.

diegogangl: I think @Neui's idea is the best way to handle it. Code and functionality look good

zeddo123: Ok, I'll work on it!

zeddo123:
I was able to disable the action if the tasks selected did not have the same parent. However, I couldn't disable the keyboard shortcut with `action_enabled_changed` so I had to call `have_same_parent` in `add_parent`. Is there a way to disable them with the action?

https://github.com/zeddo123/gtg/blob/21c20db6dcb50dea37cce92ff8cf73a04252029f/GTG/gtk/application.py#L269

diegogangl:
AFAIK you can't disable keyboard accelerators. The "workaround" is to remove and re-add them. I think it's easier to just do the checks in the action.

diegogangl:
Forgot that this was basically done, thanks!

PR #521, opened 2020-12-05
Developer Console

diegogangl:
The developer console plugin. ~Ripped off~ _Based on_ the Developer Console plugin from Pitivi.

To test this:

1. Enable the plugin
2. Press F12 or open the window from the main menu
3. Type away and mess with it

Here's an example:

```python

>>> t = req.new_task()
>>> t.set_title('Task created from code')
True
```


nekohayo:
Cool stuff! Haven't tried it yet, but if it already works for you in the dev version and in a flatpak, since it's a plugin, why not just merge and include it right away? :)

PR #526, opened 2020-12-11
Fix task date sorting order in certain situations

Neui:


GTG/gtk/browser/treeview_factory.py
```diff
@@ -189,15 +189,13 @@ def sort_by_title(self, task1, task2, order):
...
-        if sort != 0:
-            return sort
-
-        # Group tasks with the same tag together for visual cleanness
-        t1_tags = task1.get_tags_name()
-        t1_tags.sort()
-        t2_tags = task2.get_tags_name()
-        t1_tags.sort()
-        sort = (t1_tags > t2_tags) - (t1_tags < t2_tags)
+        if sort == 0:
```
nekohayo, line 192: Huh, isn't this the same if condition as further down below?

Overall, basing myself purely on the diff here and not truly understanding the context, it's a bit tricky to know what the code is doing, I wonder if it could benefit from some explanatory comments in there, especially if it's sensitive code?

Neui: Best to have the original version available: https://github.com/getting-things-gnome/gtg/blob/2a213f0163ad693fe351188af3d141834efdf146/GTG/gtk/browser/treeview_factory.py#L174-L234 and the "modified" version: https://github.com/getting-things-gnome/gtg/commit/34a7acdd105ac5d510a8255178019995a6f790c2#diff-86f91a5bcfc419848d764a80fa0fb02319825b3bd4abaae6a3323be378daf5fa
(Tip: At the top right (or in the top middle inside a ⚙ in some "views") you can switch between split and unified mode, which makes comparing easier)

Basically, I did an optimization to return early, but that didn't take descending sorting into account (like in the original version). Also, the condition is the same as further down, since each if-block sets the `sort` variable and only do additional "sub-sorting" when they are equal (for example, equal times? Then check by tags. Also equal? Check by title. Then possibly negate it when sorting descending).

I'll look into making a more descriptive commit and add comments

Neui: Okay, added an comment and modified the commit message.

Neui:
Actually, I am noticing that the original code also doesn't check whenever it should "reverse" the sorting the dates only (`reverse_if_descending` is only called when resolving ties). I'll look into it further.

Neui:
Actually, I reverted the changes which looks better now, but instead fixed a small other bug (and added a comment about the sorting order, for now so others like me don't get confused again)

diegogangl (approved):

diegogangl:
Thanks man!

Issue #402, opened 2020-07-08
With a big dataset, switching between tags in workview (actionable) mode is slower than it used to be in 0.3.x


nekohayo:
It is a long-standing fact that GTG has always been rather slow when displaying large amounts of tasks, particularly when you have multiple hundreds of them and you are clicking "All tasks" (instead of a specific tag) when not in the "Actionable" view mode (a.k.a. workview).

Up until GTG 0.3.1, this was very painful for startup times (and for exiting the workview). This is issue #109.

However, with GTG 0.4 / git, the performance behavior pattern is different from 0.3.x:
* In 0.3.x, if you were inside the workview, and loaded only the tasks from a particular tag, instead of trying to display everything, then it was very fast.
* In 0.4/git, switching between tags is slow even if you are in "Actionable" view mode, even if it ends up displaying only a couple of tasks in the end.
* Interestingly, one area where 0.4/git is _faster_ than 0.3, is toggling between Open and Actionable (workview) modes. In 0.3.x, it was _excruciatingly_ slow, and in 0.4, it's instantaneous.

So, from a user's perspective, you lose something and you gain something. The performance problem kinda swapped around between 0.3.1 and 0.4: switching tags was fast and is now slow, switching view modes was slow and is now fast.

You can use `./launch.sh -s bryce` to use the heavy (~1MB) "bryce" dataset for testing performance issues more easily.

While the port to LXML _might_ fix the symptoms of issue #109 and of this new ticket here, I think it's important to report that there is some sort of underlying performance issue before the problem gets "hidden under the carpet" by LXML's superior performance:
* Normally, displaying small numbers of tasks should be near instantaneous, or at least much faster than displaying "all tasks". Right now, when a tag has over 100 tasks in it, it feels equally slow to switch to that tag whether you are in Open, Actionable or Closed tasks view mode.
[...]

PR #530, opened 2020-12-20
Optim/no refresh on hidden pane

jaesivsm:
Depends on https://github.com/getting-things-gnome/liblarch/pull/22 won't `self.refilter` on non-displayed tree, cutting by 3 the display time when selecting a tag.

### Context

When selecting a tag with around 150 tasks, the process time gets up to 1s5. Mainly I found that a lot of callbacks are called in `liblarch` but as I don't fully understand that library, I didn't do much on that side. 

I did witness though, that for all three panes the `apply_filter` function, each times, calls on `refilter`. That last one browse the tree in _lots_ of ways. 
Although, and that's where I have doubts, all three `refilter` aren't necessary, and it's only needed for the displayed pane.
The fix is to only trigger the refresh for the active pane.

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -75,21 +75,18 @@ class MainWindow(Gtk.ApplicationWindow):
 
         # Active Tasks
         self.activetree = self.req.get_tasks_tree(name='active', refresh=False)
-        self.activetree.apply_filter('active', refresh=False)
         self.vtree_panes['active'] = \
             self.tv_factory.active_tasks_treeview(self.activetree)
 
         # Workview Tasks
         self.workview_tree = \
             self.req.get_tasks_tree(name='workview', refresh=False)
-        self.workview_tree.apply_filter('workview', refresh=False)
         self.vtree_panes['workview'] = \
             self.tv_factory.active_tasks_treeview(self.workview_tree)
 
         # Closed Tasks
         self.closedtree = \
             self.req.get_tasks_tree(name='closed', refresh=False)
-        self.closedtree.apply_filter('closed', refresh=False)
         self.vtree_panes['closed'] = \
             self.tv_factory.closed_tasks_treeview(self.closedtree)
 
@@ -131,7 +128,7 @@ class MainWindow(Gtk.ApplicationWindow):
 
         self.restore_state_from_conf()
 
-        self.on_select_tag()
+        self._reapply_filter()
         self._set_defer_days()
         self.browser_shown = False
 
@@ -408,8 +405,6 @@ class MainWindow(Gtk.ApplicationWindow):
         clsd_tsk_key_prs = self.on_closed_task_treeview_key_press_event
         self.vtree_panes['closed'].connect('key-press-event', clsd_tsk_key_prs)
         self.vtree_panes['closed'].connect('cursor-changed', self.on_cursor_changed)
-
-        self.closedtree.apply_filter(self.get_selected_tags()[0], refresh=True)
 
         b_signals = BackendSignals()
         b_signals.connect(b_signals.BACKEND_FAILED, self.on_backend_failed)
@@ -1282,7 +1277,9 @@ class MainWindow(Gtk.ApplicationWindow):
                 task.set_status(Task.STA_DISMISSED)
                 self.close_all_task_editors(uid)
 
-    def _reapply_filter(self, current_pane):
+    def _reapply_filter(self, current_pane: str = None):
+        if current_pane is None:
+            current_pane = self.get_selected_pane()
         filters = self.get_selected_tags()
         filters.append(current_pane)
         vtree = self.req.get_tasks_tree(name=current_pane, refresh=False)
@@ -1320,7 +1317,7 @@ class MainWindow(Gtk.ApplicationWindow):
                 self.quickadd_entry.set_text(tag.get_attribute("query"))
                 break
 
-        self._reapply_filter(self.get_selected_pane())
+        self._reapply_filter()
 
     def on_pane_switch(self, obj, pspec):
         """ Callback for pane switching.
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -569,14 +569,9 @@ class MainWindow(Gtk.ApplicationWindow):
             GObject.idle_add(open_task, self.req, t)
 
     def refresh_all_views(self, timer):
-        active_tree = self.req.get_tasks_tree(name='active', refresh=False)
-        active_tree.refresh_all()
-
-        workview_tree = self.req.get_tasks_tree(name='workview', refresh=False)
-        workview_tree.refresh_all()
-
-        closed_tree = self.req.get_tasks_tree(name='closed', refresh=False)
-        closed_tree.refresh_all()
+        for pane in 'active', 'workview', 'closed':
+            self.req.get_tasks_tree(pane, False).reset_filters(refresh=False)
+        self._reapply_filter()
 
     def find_value_in_treestore(self, store, treeiter, value):
         """Search for value in tree store recursively."""
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -68,7 +68,6 @@ class MainWindow(Gtk.ApplicationWindow):
         self.app = app
         self.config = self.req.get_config('browser')
         self.tag_active = False
-        self.applied_tags = []
 
         # Treeviews handlers
         self.vtree_panes = {}
@@ -139,7 +138,7 @@ class MainWindow(Gtk.ApplicationWindow):
         app.timer.connect('refresh', self.refresh_all_views)
         app.timer.connect('refresh', self._set_defer_days)
 
-        self.stack_switcher.get_stack().connect('notify::visible-child', self._save_view_pane)
+        self.stack_switcher.get_stack().connect('notify::visible-child', self.on_pane_switch)
 
         # This needs to be called again after setting everything up,
         # so the buttons start disabled
@@ -427,27 +426,28 @@ class MainWindow(Gtk.ApplicationWindow):
         self.on_search_toggled()
 
     def on_search_toggled(self, widget=None):
-
         if self.searchbar.get_search_mode():
             self.search_button.set_active(False)
             self.searchbar.set_search_mode(False)
             self.search_entry.set_text('')
-            self.unapply_filter_on_panes(SEARCH_TAG, refresh=True)
+            self.get_selected_tree().unapply_filter(SEARCH_TAG)
         else:
             self.search_button.set_active(True)
             self.searchbar.set_search_mode(True)
             self.search_entry.grab_focus()
 
-    def on_search(self, data):
-        query = self.search_entry.get_text()
-        log.debug(f"Searching for '{query}'")
-
+    def _try_filter_by_query(self, query, refresh: bool = True):
+        log.debug("Searching for %r", query)
+        vtree = self.get_selected_tree()
         try:
-            parsed_query = parse_search_query(query)
-            self.apply_filter_on_panes(SEARCH_TAG, parameters=parsed_query)
-        except InvalidQuery as e:
-            self.unapply_filter_on_panes(SEARCH_TAG, refresh=True)
-            log.debug(f"Invalid query '{query}' : '{e}'")
+            vtree.apply_filter(SEARCH_TAG, parse_search_query(query),
+                               refresh=refresh)
+        except InvalidQuery as error:
+            log.debug("Invalid query %r: %r", query, error)
+            vtree.unapply_filter(SEARCH_TAG)
+
+    def on_search(self, data):
+        self._try_filter_by_query(self.search_entry.get_text())
 
     def on_save_search(self, action, param):
         query = self.search_entry.get_text()
@@ -459,7 +459,7 @@ class MainWindow(Gtk.ApplicationWindow):
         if self.tagtreeview is not None:
             self.select_search_tag(tag_id)
         else:
-            self.apply_filter_on_panes(tag_id)
+            self.get_selected_tree().apply_filter(tag_id)
 
     def select_search_tag(self, tag_id):
         tag = self.req.get_tag(tag_id)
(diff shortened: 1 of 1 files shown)


GTG/gtk/browser/main_window.py
```diff
@@ -1285,9 +1285,12 @@ def on_dismiss_task(self, widget=None):
     def apply_filter_on_panes(self, filter_name, refresh=True, parameters=None):
         """ Apply filters for every pane: active tasks, closed tasks """
```
nekohayo, line 1287: Isn't this docstring a lie now? (and the one in unapply_filter_on_panes too)

GTG/gtk/browser/main_window.py
```diff
@@ -1285,9 +1285,12 @@ def on_dismiss_task(self, widget=None):
...
 
+        current_pane = self.get_selected_pane()
         for pane in self.vtree_panes:
             vtree = self.req.get_tasks_tree(name=pane, refresh=False)
-            vtree.apply_filter(filter_name, refresh=refresh, parameters=parameters)
+            vtree.apply_filter(filter_name,
+                               refresh=current_pane == pane,
```
nekohayo: Given how important this seems to be to maintain somewhat decent performance, I think it might be good to add some comment(s) in the code to explain the gotchas / warn people to only refresh the currently shown pane to cut down the performance impact massively, avoiding potential future naïve code mistakes?

nekohayo:
Wow so... if I understand correctly, you're saying the app was asking liblarch to refresh "All tasks" + "Actionable" + "Closed" everytime you click a tag in the sidebar? Interesting finding, that explains why toggling the Workview in 0.3.x was slow but it isn't in 0.4+;

Silly question though: what happens now when you switch between panes, it will have to refilter the view, which means switching between panes will no longer be instantaneous? We might be bringing back an old problem I guess, unless there's a way for the filtering to happen first in the current view, and then "asynchronously, in the background" do it for the 2 other views as well...

Neui:
> Wow so... if I understand correctly, you're saying the app was asking liblarch to refresh "All tasks" + "Actionable" + "Closed" everytime you click a tag in the sidebar?

From what I've seen while trying to check the performance, yes. I thought that was intentional ("switching tabs is fast now"), and also thought of doing something similar to this PR (but with your async suggestion), but I ultimately wanted to rather fix the "underlying" issue first (current unsuccessfully).

GTG/gtk/browser/main_window.py
jaesivsm: Well yeah, I'll adjust the docstring, but indeed, by default with the `force_refresh=False` all pane _won't_ be applied filters upon.

jaesivsm:
> Wow so... if I understand correctly, you're saying the app was asking liblarch to refresh "All tasks" + "Actionable" + "Closed" everytime you click a tag in the sidebar?

Yup.

> I thought that was intentional

Same here, hence the PR and the open question. Although, in the call to liblarch, the `refresh` parameter was actually ignored. So i wasn't sure about it entirely. 

> Silly question though: what happens now when you switch between panes, it will have to refilter the view, which means switching between panes will no longer be instantaneous?

I didn't see issues while testing around but that's mainly what I was worried about indeed.
However if some more refiltering is needed I guess it would be more appropriate for it to be only "just on time" for the current panel. So I might need adding a call to `apply_filter_on_panes` when switching pane...

nekohayo:
I tried this out with your liblarch branch, and I saw a 2x improvement with various operations when running the "bryce" sample data set (`./launch.sh -s "bryce"`), which has 827 tasks. However, I encountered a bug: it breaks the Actionable view on that data set.

# Git master
* startup time: 23.0 seconds
* switching to "Actionable" view: 0 secs
* switching back to "Open" tasks view: 0 secs
* switching to "work" tag: 6.2 secs
* switching back to all tasks: 13.6 secs

# With your optimizations branch
* startup time: 12.7 secs
* switching to "Actionable" view: **BROKEN**, no tasks are displayed at all!
* switching back to "Open" tasks view: 0 secs
* switching to "work" tag: 3.6 secs
* switching back to all tasks: 6.6 secs

Note: performance gains on the "bryce" dataset are 2x rather than the potential 3x, because that dataset doesn't have more than 1 closed tasks in it, so from a data perspective it's almost as if it only had two panes in practice.

jaesivsm:
Ok so this time it should be good
Also I've redone the whole logic. (thanks again to the `-s bryce` dataset !)

I've dropped the previous logic of _unapplying_ one by one every filter. We now reset the whole bunch since there is no performance gain in not doing so (no `refilter()` is called between those _unapplication_). Plus, _unapplying_ implied keeping a separate list of selected tags that comes as a duplicate of the filters list that `liblarch` is keeping.
We now reset the filter list and trigger `refilter()` if needed on every tag / pane switch but only for the current pane.

GTG/gtk/browser/main_window.py
```diff
@@ -1282,45 +1282,50 @@ def on_dismiss_task(self, widget=None):
...
+        filters = self.get_selected_tags()
+        filters.append(current_pane)
+        vtree = self.req.get_tasks_tree(name=current_pane, refresh=False)
+        # Re-applying search if some search is specified
+        search = self.search_entry.get_text()
+        if search:
+            filters.append(SEARCH_TAG)
+        # only reseting filters if applieds filters are different from current
+        # ones, leaving a chance for liblarch to make the good call on wether
+        # to refilter or not
```
Neui, line 1295: Not that important but somewhat annoys me:  
reseting → resetting  
applieds → the applied  
wether → whether

Neui (commented):
Looks good to me, but I want to test it for myself.

GTG/gtk/browser/main_window.py
```diff
@@ -427,27 +426,28 @@ def toggle_search(self, action, param):
...
         try:
-            parsed_query = parse_search_query(query)
-            self.apply_filter_on_panes(SEARCH_TAG, parameters=parsed_query)
-        except InvalidQuery as e:
-            self.unapply_filter_on_panes(SEARCH_TAG, refresh=True)
-            log.debug(f"Invalid query '{query}' : '{e}'")
+            vtree.apply_filter(SEARCH_TAG, parse_search_query(query),
+                               refresh=refresh)
+        except InvalidQuery:
+            log.debug("Invalid query %r", query)
```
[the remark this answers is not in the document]
jaesivsm: Yeah, got dropped in the heat of the rewrite, i'll add it back again.

GTG/gtk/browser/main_window.py
jaesivsm: Comments are important :) Fixed !

GTG/gtk/browser/main_window.py
```diff
@@ -1282,45 +1282,50 @@ def on_dismiss_task(self, widget=None):
...
+        # to refilter or not
+        if sorted(filters) != sorted(vtree.list_applied_filters()):
+            vtree.reset_filters(refresh=False)
+        # browsing and applying filters
+        for filter_ in filters:
+            is_last = filter_ == filters[-1]
+            if filter_ == SEARCH_TAG:
+                self._try_filter_by_query(search, refresh=is_last)
+            else:
+                vtree.apply_filter(filter_, refresh=is_last)
```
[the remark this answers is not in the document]
jaesivsm: Well, `filter` is a python reserved keyword, and it's not recommended as a rule to override reserved names, as for `filter_` especially, it's the [PEP8 recommended behavior.](https://legacy.python.org/dev/peps/pep-0008/#function-and-method-arguments). But I can find a synonym. 

I'm not a big fan of the `is_last` thingy either, and I'd rather prefer explicitly call for `refilter()`. Better yet, a new `liblarch` method allowing to apply multiple filters at once. But since I didn't want to alter too much of the `liblarch` API and still wanted to limit calls I thought of this. But yeat, a comment is indeed needed.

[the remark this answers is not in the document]
Neui: > Well, filter is a python reserved keyword, and it's not recommended as a rule to override reserved names, as for filter_ especially, it's the PEP8 recommended behavior.. But I can find a synonym.

But `filter` is a [built-in function](https://docs.python.org/3/library/functions.html#filter) (or class as the python3 interpreter says), not a reserved [keyword](https://docs.python.org/3/reference/lexical_analysis.html#keywords) from what I can see. I can redefine it at least in the python REPL, so I'm sure it isn't a keyword.

GTG/gtk/browser/main_window.py
```diff
@@ -1295,7 +1295,8 @@ def _reapply_filter(self, current_pane):
         # whether to refilter or not
         if sorted(filters) != sorted(vtree.list_applied_filters()):
             vtree.reset_filters(refresh=False)
-        # browsing and applying filters
+        # Browsing and applying filters. For performance optimization, only
+        # allowing liblarch to trigger a refresh on last item.
```
Neui, line 1299: Doesn't really say why you couldn't just do `refilter`. Maybe append "This way, liblarch can prevent refreshing in certain situations rather than refilter all the time."

jaesivsm: Ok I'll edit that. But truth is, I can't call on `ViewTree.refilter()` because there is no such method. `ViewTree` doesn't proxy that method to `FilteredTree` and I don't know why at all. This is what I meant in the second part of that [comment](https://github.com/getting-things-gnome/gtg/pull/530#discussion_r554457225).

GTG/gtk/browser/main_window.py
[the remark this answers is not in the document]
jaesivsm: Yes indeed, keyword and not built-in function, sorry about the mix up. 
But to me it's still overriding a declared value. From experience it may lead to some complicated debugging. So yeah I usually follow the PEP8 recommendation about keywords on that too. 
But since it's important I'll use the name from the `ViewTree.apply_filter` function.

Neui:
Looks good to me, switching seems to work as intended. However, it takes longer to start, around 50s instead of 37s on current master (when the "Open" view is selected), and seems to be rather consistent. On top of longer startup time, it'll refresh when switching views/tabs directly after startup, so it looks like it force refreshes or something.
(Side note: On a new day, which `refresh_all_views` is called, took 3:10m, but it doesn't use any of the modified methods, so maybe don't really worry for this PR.)

jaesivsm:
Hum I'm not sure I get how you obtain those figures. Is there a defined way to measure these ? Especially the startup time, all I could come up with was to measure the time for `do_startup` and `do_activate` method execution. 
I had to write a decorator to count time passed in a function and here's what I got : 
_(everything is executed right after a `git clean -fdx`, because I have no idea of the caching that actually takes place and with the `-s bryce` dataset)_

**Note :** for the patch to actually improve perf you have to have `liblarch` on the branch mentioned in the first message of this PR. If not, the refresh won't be avoided and it might indeed take more time. 

### On `master` branch :

Startup gives up : 
```
| do_startup: 1594.282ms
| | init_style: 0.776ms

| do_activate: 9532.568ms
| | init_browser: 9484.415ms
| | | __init__: 9484.368ms
| | | | on_select_tag: 6108.745ms
| | init_actions: 1.070ms
| | init_plugin_engine: 0.090ms
```

Selecting the ubuntu tag : 
```
| on_select_tag: 2525.094ms
```

### With the patch : 

Startup :
```
| do_startup: 1621.952ms
| | init_style: 0.757ms

| do_activate: 6719.164ms
| | init_browser: 6679.087ms
| | | __init__: 6679.042ms
| | | | on_select_tag: 3412.005ms
| | init_actions: 0.934ms
| | init_plugin_engine: 0.077ms
```
Selecting the ubuntu tag : 
```
| on_select_tag: 771.174ms
```
Of course, and that's the matter of the patch, selecting a different pane takes time the first time. 
Here, with ubuntu tag selected, swithing to the "closed" pane: 
```
| on_pane_switch: 1187.721ms
```

jaesivsm:
After the previous comment from @Neui I got curious about startup time (wasn't really the subject of the PR at first), and I saw that `liblarch.FilteredTree.refilter()` was called during window initialization a bunch of time.
I could scrap one of those : 
```diff
@@ -409,8 +406,6 @@ class MainWindow(Gtk.ApplicationWindow):
         self.vtree_panes['closed'].connect('key-press-event', clsd_tsk_key_prs)
         self.vtree_panes['closed'].connect('cursor-changed', self.on_cursor_changed)
 
-        self.closedtree.apply_filter(self.get_selected_tags()[0], refresh=True)
-
         b_signals = BackendSignals()
         b_signals.connect(b_signals.BACKEND_FAILED, self.on_backend_failed)
         b_signals.connect(b_signals.BACKEND_STATE_TOGGLED, self.remove_backend_infobar)
```
salvaging another second on my test (9s on master, 6s previously, roughly 5s here). And since the method `on_select_tag` (now `_reapply_filter`) was called afterwards the call was useless.
I also noticed that during `do_startup` the registering backend phase, was actually triggering a __lot__ of `liblarch.FilteredTree.__update_node` call (without getting through a `refilter()`) which adds up to some incomprehensible 1s5. 
That might be another optim for another PR right there.

Neui:
> Hum I'm not sure I get how you obtain those figures. Is there a defined way to measure these ? Especially the startup time, all I could come up with was to measure the time for do_startup and do_activate method execution.

Sorry for not being more specific. In this case I just ran `./launch.sh` with my own dataset (not bryce, and I won't provide it, sorry), and started counting with the help of the clock after it install/seeing a warning, until the main window appeared.

> Note : for the patch to actually improve perf you have to have liblarch on the branch mentioned in the first message of this PR. If not, the refresh won't be avoided and it might indeed take more time.

Makes sense, I forgot to apply it. I'll try it with that patch later/this evening.

Neui:
> > Note : for the patch to actually improve perf you have to have liblarch on the branch mentioned in the first message of this PR. If not, the refresh won't be avoided and it might indeed take more time.
> 
> Makes sense, I forgot to apply it. I'll try it with that patch later/this evening (GMT).

Well I missed it, so I did it now.
With the liblarch PR applied, it takes around 27s for the main window to appear on master, and around 12s with this PR applied.

GTG refreshes when waking up/new day occurs. I took an bit different route and enabled the developer console (in the extensions), opened it (from the menu top right), `app.timer.emit_refresh()`, wait until you get a return value. It took around 2m30s on master, and 5s on this PR. It also "invalidates" the other tabs (so it needs to wait when changing views), but I didn't test whenever it correctly changes the dates.

One thing I noticed was that when switching to the "closed" tab (that is big since I disabled task cleanup), it takes around 18s (from clicking until responding/changing the view visually). Since master already "filters" the tab, it seems a bit odd to me that this PR takes a bit longer (12s startup + 18s switch = 30s). Not too bad though.

I'll review the new code changes soon.

Neui (approved):
Looks good to me and works.

nekohayo:
I've tested this again and the issue I had seen before seems gone. Performance is a net improvement (see my updated benchmark results in https://github.com/getting-things-gnome/gtg/issues/402#issuecomment-760660680 if you're curious), I'm excited and looking forward to further optimizations you may come up with! So I will now merge this as-is, and hope @diegogangl won't encounter too many (if any) conflicts with his big WIP branches ;)

PR #539, opened 2021-01-12
Using lazy evaluated string for logging

jaesivsm:
Avoiding casting variables into ignored logging messages (very light performance improvements). Declaring logger with module name to allow future per-module logger organization.


diegogangl:
Is %r faster than using F-strings?

jaesivsm:
Not exactly, what's faster is that the string composing is done by the logging library only if the log level is sufficient. 
On a prod log level (say WARNING or something similar) all the debug message not only would not be displayed, but the string would not get formatted. The key is in letting the logging lib making the composition.

TLDR: `logging.debug(f'{stuff}')` is just as fast as `logging.debug('%r' % stuff)` but all of that might be slower than `logging.debug('%r', stuff)`

diegogangl:
> Not exactly, what's faster is that the string composing is done by the logging library only if the log level is sufficient.

Gotcha.
Looks like I caused some more conflicts with the latest merges, sorry!

diegogangl:
Looks good, but any special reason to change `log` to `logger`?

jaesivsm:
Well I didn't put much thoughts into it but I mainly name it like that because `logger = logging.getLogger()` makes sense. It's kind of an habit I took from the [python doc](https://docs.python.org/3/library/logging.html#logging.Logger.debug). Thinking about it I'd say it marks the changes between the old import, and the new object instantiation. 

But of course it doesn't has to be this way. I can revert to `log` if you prefer :)

diegogangl:
Sorry to make you do more work, but I think log was better. Less characters == less chances of having to split the line :)

jaesivsm:
Done :)

diegogangl:
Looks good, Thanks!

PR #547, opened 2021-01-21
Fixing switch pane optimization

jaesivsm:
While working on pane switching optimization, forgot one instance of `apply_filter_on_panes` getting called in the tag editor. Renaming `_reapply_filter` to `apply_filter` (since called outside of base class, can't be a "private" method anymore).


diegogangl:
LGTM thanks!

PR #554, opened 2021-01-30
meson: Require xgettext to build plugins

Neui:
Otherwise it would just silently don't copy the plugin files, so no plugins would be available when using GTG.
Found by nekohayo, where the test machine didn't had gettext.
Couldn't really test whenever it works so please test if you have such environment.


diegogangl:
LGTM

PR #555, opened 2021-01-30
meson: Always build plugin files for translation

Neui:
Because otherwise when updating translations, those new translations aren't applied until forced a rebuild.

This would also mean you don't need the plugin workaround anymore when I introduced it in #494.


diegogangl:
LGTM thanks!

Issue #549, opened 2021-01-25
Turn on G_ENABLE_DIAGNOSTIC in debug.sh/launch.sh for deprecation warnings by default


nekohayo:
The `G_ENABLE_DIAGNOSTIC` env variable, if set to a non-zero value, "enables diagnostic messages, like deprecation messages for GObject properties and signals."

So `G_ENABLE_DIAGNOSTIC=1 ./launch.sh` 

We should probably make that the default behavior of our launch/debug script, so that we can catch deprecations early rather than nearly 9 years later like in 541 ;)

Neui:
This is from https://developer.gnome.org/glib/2.64/glib-running.html

Does it also work for methods? It says it is only for properties and signals, but in the linked issue `Gdk.cairo_set_source_color` is a method. Running with it doesn't reveal that method is still being used (I confirmed it that it runs via a debugger).

nekohayo:
Oooh, this is what you meant, I hadn't really understood by the "does it work for methods" question you had... so if this doesn't work (from what you've seen), is there some checking tool that we could use to warn us about this?

Neui:
In PyGObject I found an [`deprecated`](https://gitlab.gnome.org/GNOME/pygobject/-/blob/0a16082c3e092fdccf5499172a4b9dec07aa383d/gi/overrides/__init__.py#L210-217) annotation that uses [`warnings.warn`](https://docs.python.org/3/library/warnings.html) to warn. You can use [`PYTHONWARNINGS`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONWARNINGS) to enable them. With `PYTHONWARNINGS=once` I got it:

```
gtg/lib/python3.8/site-packages/gi/overrides/__init__.py:333: DeprecationWarning: Gdk.color_parse is deprecated
  ret = method(*args, **kwargs)
gtg/.local_build/install/lib/python3.8/site-packages/GTG/gtk/browser/main_window.py:487: DeprecationWarning: Gtk.Widget.get_state is deprecated
  is_maximized = widget.get_window().get_state() & mask == mask
gtg/.local_build/install/lib/python3.8/site-packages/GTG/gtk/browser/cell_renderer_tags.py:171: DeprecationWarning: Gdk.cairo_set_source_color is deprecated
  Gdk.cairo_set_source_color(gdkcontext, my_color)
```

There is also `PYTHONWARNINGS=default` we could use ("Warn once per call location").

PR #561, opened 2021-02-14
debug.sh: Introduce -p to enable python dev mode

Neui:
From https://docs.python.org/3/library/devmode.html
Introduced in Python 3.7

Basically enables warnings and other stuff that could be useful to look into, including deprecation warnings.

Also see #549; Does not really fix it since it doesn't set `G_ENABLE_DIAGNOSTIC=1`.

### DIFF
--- a/scripts/debug.sh
+++ b/scripts/debug.sh
@@ -9,20 +9,22 @@ fi
 args=""
 dataset="default"
 norun=0
+pydebug=0
 title=""
 
 # Create execution-time data directory if needed
 mkdir -p tmp
 
 # Interpret arguments. The ":" following the letter indicates that the opstring (optarg) needs a parameter specified. See also: https://stackoverflow.com/questions/18414054/rreading-optarg-for-optional-flags
-while getopts "bdns:t:" o;
+while getopts "bdwns:t:" o;
 do  case "$o" in
     b)   args="$args --boot-test";;
     d)   args="$args -d";;
+    w)   pydebug=1;;
     n)   norun=1;;
     s)   dataset="$OPTARG";;
     t)   title="$OPTARG";;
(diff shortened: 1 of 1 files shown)


diegogangl:
Why don't we include `PYTHONWARNINGS=default` in this one? Also maybe we should use `-w` (for warnings)

Neui:
I primarily oriented on `-d` which is to disable by default, but I guess I could export `PYTHONWARNINGS=default` (when not set) all the time.
Note that the dev mode already also does that.

I used `-p` for **P**ython Dev mode, but I can also see using `-w` for **w**arnings, but it also enables additional features (but that we might not be interested in it, such as allocation debug or asyncio).

diegogangl:
I meant adding python warnings to `-p`, and renaming to `-w` for all warnings

Neui:
> I meant adding python warnings to `-p`

The Python Development Mode already does that:

> Effects of the Python Development Mode:
> 
> * Add [`default` warning filter](https://docs.python.org/3/library/warnings.html#describing-warning-filters). ...

Or do you mean you want it explicitly?

> and renaming to `-w` for all warnings

Done.

diegogangl:
Sorry I meant `PYTHONWARNINGS=default` (from Pygobject)

Neui:
> Sorry I meant `PYTHONWARNINGS=default` (from Pygobject)

PyGObject uses Python warnings which are enabled automatically by using Python Development Mode (which is activated by using `-w` now), as I said above. Or should I explicitly enable it when using a lower python version?

diegogangl:
Sorry I hadn't read the link thoroughly -_-
This is awesome btw! Running `-dw` gets a lot of info now

PR #566, opened 2021-02-24
 Correct paths in import script for the Flatpak version

morhe-blip:
Correcting directory names to match upper case


diegogangl:
Nice catch, we missed this when we changed the appid. Thanks!

PR #577, opened 2021-03-11
Updated Dutch translation

Vistaus:
🙂


diegogangl:
Thanks for the work! I'm sorry, it seems I caused some conflicts after updating the PO files, could you rebase?

Vistaus:
Done! 🙂

Neui:
Unfortunately according to GitHub you just deleted the file, so something went wrong during the rebase... 

If you still have the original file (or the commit hash/id, then do `git checkout the-commit-id`), then you could replace it with your version and [follow the guide to re-generate your translation file](https://github.com/getting-things-gnome/gtg/blob/master/docs/contributors/translating.md) with the new strings (that you can then translate). You can then just edit and commit `po/nl.po`, ignoring the other changed files.

Vistaus:
I'll just do a new PR instead - that's easier for me.

PR #579, opened 2021-03-12
Fix bug: Recurring status of tasks is lost after GTG restart #578

### LINKED ISSUE #578: Recurring status of tasks is lost after GTG restart
Tasks that are marked as recurring lose this property after GTG restart.

Reproduce:
Create new task, set as recurring, restart GTG. Task has lost Recurring status.

Happens with current master
commit 60be39b6d6f0f9acf7d94c0f6e129883f2c3eaab (HEAD -> master, origin/master, origin/HEAD)

But I am quite sure it also happened before.

zeddo123:
I don't know why but `recurring_term = element.find('term').text` was always returning `None` even though the term was properly set which made the task lose its recurring status. So I've changed it to `recurring.getchildren()[0].text`

close #578

### DIFF
--- a/GTG/core/xml.py
+++ b/GTG/core/xml.py
@@ -77,7 +77,7 @@ def task_from_element(task, element: etree.Element):
     recurring_enabled = recurring.get('enabled')
 
     try:
-        recurring_term = element.find('term').text
+        recurring_term = recurring.getchildren()[0].text
         task.set_recurring(recurring_enabled == 'true',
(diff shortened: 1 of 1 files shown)


diegogangl:
My bad, I mixed the lxml elements when porting. The proper solution is:

```
        recurring_term = recurring.find('term').text
```

diegogangl:
I went ahead and fixed this in master, no point bothering you with it :D

Issue #580, opened 2021-03-13
Include all files that can be translated (POTFILES.in)


Neui:
Since now we're close to release, I suggested to be a good time to update the translation files, which was done in 60be39b6d6f0f9acf7d94c0f6e129883f2c3eaab.

However, I noticed that not all strings are available for translation.
As of now I found that the developer console plugin isn't marked to be translatable, but can be fixed by adding the following lines to `po/POTFILES.in`:

```
GTG/plugins/dev_console.gtg-plugin.desktop
GTG/plugins/dev_console/console.py
GTG/plugins/dev_console/utils.py
```

But that doesn't make sure that every translatable string is being considered.
Either we should just include every python file, or have some script to update that file.

(Note that for some reason if I regenerate the PO(T) files I get entries about `@APP_ID@` isn't in current master; Is this being removed manually or..?)

diegogangl:
Maybe we should include every python file in `gtk/` and `plugins/`. I don't think the dev console has any translatable strings tho

Neui:
> I don't think the dev console has any translatable strings tho

Actually it does: https://github.com/getting-things-gnome/gtg/blob/41d55257c55a84aab81bf0dbf095d5b16d03bfce/GTG/plugins/dev_console/console.py#L117-L126

diegogangl:
ohh the welcome message, right. The whole thing should be translatable actually, I'll get that fixed

PR #581, opened 2021-03-13
Update german translation (2020-03-13)

Neui:
~~I'd rather wait for #580 before reviewing and merging just in case there are some strings missing.~~

Possible interesting decisions:
- Treeview Items → Teilaufgaben (subtasks) because Treeview Items is too technical and the treeview items are the (sub)tasks.
- ... in Task Browser → (none) because it is already in a section about the task browser, no need to be redundant.
- shortcuts → Variable (variable) because it fits better (also at first I thought of keyboard shortcuts but they are just quick-access variables)
- The variable descriptions (the one in parenthesis) can't be translated. Not too bad (since it is a developer tool), but maybe it could be considered to make it translatable.

Asking @renehsz for a review (like last time)


po/de.po
```diff
@@ -2194,15 +2197,18 @@ msgstr "@untouched Schlagwort hinzufügen"
 #: GTG/plugins/dev_console.gtg-plugin.desktop:4
 #: GTG/plugins/dev_console/console.py:71 GTG/plugins/dev_console/console.py:100
 msgid "Developer Console"
-msgstr ""
+msgstr "Entwickerkonsole"
```
renehsz, line 2200: Typo here, should be "Entwicklerkonsole"

po/de.po
```diff
@@ -2216,15 +2222,31 @@ msgid ""
 "Type \"help (<command>)\" for more information.\n"
 "\n"
 msgstr ""
+"Willkommen im GTGs Entwickerkonsole\n"
```
renehsz, line 2225: Typo again, should be "in GTGs Entwicklerkonsole"

renehsz (changes requested):
I found a couple of typos that need to be fixed. Otherwise it looks good to me. Thanks again for the work you do!

po/de.po
Neui: How did I missed that? Fixed.

po/de.po
Neui: Thanks, fixed.

renehsz (approved):
Now everything looks great!

diegogangl:
Merged, thanks!

Issue #247, opened 2020-02-24
Fix the traceback catcher code/GUI, figure out what to do with Apport crasher bug reporting support


nekohayo:
In /GTG/gtk/crashhandler.py:
1. at the very end of the file, there is a mention (and URL) of LaunchPad for bug reporting. For general bugs, it should point to GitHub (or eventually GitLab). I was thinking of simply going ahead and rewriting that translatable string and pushing a fix right here and now (and take the opportunity to remove the URL from the translatable part, replace it by another variable for security so that the URL can't change based on translation files; by the way there is a way to have more descriptive gettext variables than "%s" but I don't quite remember what it was...), however...
2. the code has support for Ubuntu's [Apport](https://wiki.ubuntu.com/Apport) bug reporting tool, and that translatable string mentions it. I'm presuming it works well, but I'm wondering if this could be plugged into GitHub or Gitlab, somehow, or if it remains strictly a LaunchPad/Ubuntu-specific tool, like ABRT is to Fedora's bugzilla... presumably it is harmless to leave it there, however the problem is that crasher reports would go to LaunchPad, which we have little interest in looking at (unless we make it a policy/exception to look at crashers there).

diegogangl:
Does the crash handler actually work? I've been trying to trigger it with `os.abort()` and I don't get any reaction.

liberforce:
> Remove LaunchPad mention as being the "official" bug tracker

There's also `README.md` that references it.

diegogangl:
Thanks for spotting that :+1: I nuked both references to launchpad and also the old site

diegogangl:
I've removed the crash handler since I couldn't trigger it and looking at the code it doesn't seem to work with GtkApplication at all.

I also updated the info in the info module, though I found this and don't know what to put instead:

```
EMAIL = "gtg-contributors@lists.launchpad.net"
```

liberforce:
I think at this point, if you're switching to GitHub and don't plan to have a mailing list (people just tend to ask question in the bug tracker anyway), you could just remove it, and ask to close the existing mailing list. The difference is that people would need to have a GitHub account to ask for help, which might be a problem for some people.

nekohayo:
[...]

PR #582, opened 2021-03-13
Error popup on uncaught exceptions

Neui:
Did a while ago, but currently paused but planned to continue somewhat soon. Just in case: This is still Work In Progress (WIP)!

Basically when an uncaught exceptions is being caught (either explicit via a decorator or implicit via a global hook), it'll then display an dialog window with the info and ask the user to report it.

Also, would be a good place to also have an popup when something fails during initialization (which previously would just be silent end error out later due to uninitialized attributes).

Some ideas thrown in:
* Keep an action log of the last x actions to be included
* Maybe also include x lines of logs

Related: #247 but unlikely to have (explicit) integration with such crash reporters, for now.

First shared in the (Matrix) chat, here some (now outdated) screenshots:

first

second

### DIFF
--- a/GTG/gtg.in
+++ b/GTG/gtg.in
@@ -83,6 +83,7 @@ if _LOCAL:
 
 from GTG.core import info
 from GTG.gtk.application import Application
+from GTG.gtk.errorhandler import replace_excepthook
 
 
 def handle_local_options(application, options):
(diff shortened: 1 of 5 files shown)


Neui:
I'll un-draft it so it doesn't get stuck on "making it more perfect". It seems to work, so I guess it can be reviewed. The ideas can be sometime implemented later (better create an issue to keep track of it)

diegogangl:
Hey, this is awesome! Looks good, just some commented code and a conflict in `application.py`

Neui:
Fixed. I also changed so that it doesn't print the exception twice in the console, and added special check for `Gtk.ResponseType.DELETE_EVENT` (to prevent an "unknown response" log message)

diegogangl:
Awesome, thanks!

PR #586, opened 2021-03-13
fix nonworkview/nonactionable attribute loading

eladyn:
In the current development version, the state of the nonworkview attribute of a tag was correctly saved in the data file(s). After a restart of the application, this attribute was reset / not read properly by the loader.

Apart from the renaming in the code, I also added the `nonworkview` field in the format documentation.

However, I am not sure, if the attribute name shouldn't rather be `nonactionable` for everything. If you'd rather name it like that, I can change it accordingly in all other places.

### DIFF
--- a/GTG/core/datastore.py
+++ b/GTG/core/datastore.py
@@ -232,7 +232,7 @@ class DataStore():
             color = element.get('color')
             icon = element.get('icon')
             parent = element.get('parent')
-            nonworkview = element.get('nonworkview')
+            nonactionable = element.get('nonactionable')
 
             tag_attrs = {}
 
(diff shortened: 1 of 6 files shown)


diegogangl:
Hey @eladyn, thanks for the fix! The name should be nonactionable everywhere though. Work view is from 0.3x days and it's not used in either the code or the ui (or at least shouldn't be).

eladyn:
Thanks for the answer! In https://github.com/getting-things-gnome/gtg/blob/16080f722efba2695bd822c4b9edda54f882183c/GTG/core/versioning.py#L121 there is another attributue `notactionable`.

Since I'm not very familiar with the codebase, I am not too sure, if this should be changed as well to `nonactionable` or if this has nothing to do with this issue?

diegogangl:
Keep `nonworkview` for `tag_attrs`, but use `nonactionable` everywhere else. You can sort of see what I mean in that versioning code:

https://github.com/getting-things-gnome/gtg/blob/16080f722efba2695bd822c4b9edda54f882183c/GTG/core/versioning.py#L99

eladyn:
I'm not entirely sure, if I understood you correctly. Does this look reasonable?

diegogangl:
My suggestion was to avoid changes in the old code, what you did is better :D 
Thanks!

PR #587, opened 2021-03-15
Use a timer for live search

### LINKED ISSUE #281: Performance: Searchbar searches live on every character being typed
Observation: 0.4's searchbar searches "live" everytime you type a character in it.

~~Theoretical supposition (to be verified):~~ performance will be crap, as soon as you have more than a handful of tasks. ~~Note that I have not bench-tested this (yet), but I _presume_ performance will be horrible, unless the search is only something like a Gtk TreeModel filter and doesn't touch the storage backend, but I doubt that's the case...~~

Ways to avoid triggering performance issues:
* If keeping a (nearly) "live" filtering feature:
  * Do not run a search if 0 < search_query_string_length < 3 (but if 0 characters, that counts as "clear/reset the search"). There is no reason to search for something less than 3 characters, no significant words or acronyms are made of 2 characters and it just returns a ton of results that impair your search performance (if your search benefits from "filtering existing results" in its algorithm). This alone won't be enough, but it probably will help a little.
  * Introduce a timer-based activation algorithm like what I'm proposing for parsing tags in the task editor in #270
* The nuclear option: "don't do live search at all". In this case, put a "Search" button between the search field and the "Save search" button, and also respond to the Enter/Return keyboard shortcut, but don't trigger the search until the user explicitly says so. This is less "sexy" than semi-live filtering, but is simultaneously the best-performing and easiest-to-implement approach.

diegogangl:
Use a timer for live search (taken from taskview)

fixes #281

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -22,7 +22,7 @@ import threading
 import datetime
 import logging
 
-from gi.repository import GObject, Gtk, Gdk, Gio
+from gi.repository import GObject, Gtk, Gdk, Gio, GLib
 
 from GTG.core import info
 from GTG.backends.backend_signals import BackendSignals
@@ -69,6 +69,9 @@ class MainWindow(Gtk.ApplicationWindow):
         self.app = app
         self.config = self.req.get_config('browser')
         self.tag_active = False
+
+        # Timeout handler for search
+        self.search_timeout = None
 
         # Treeviews handlers
         self.vtree_panes = {}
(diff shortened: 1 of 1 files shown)


nekohayo:
This code is GOLD. I've been trying out various searches on before/after versions, typed fast vs semi-slowly, and your patch just brought down my search wait time from 17 seconds down to 3-5 seconds (depending on typing speed), and the UI is infinitely smoother when typing. I'd say we merge this for 0.5 for sure!

nekohayo (approved):

diegogangl:
Now you got another thing to blog about \o/

PR #594, opened 2021-03-21
Updates to manual for adding parent task to existing task (#517)

vansia43:
I think this was the only little thing that I missed since the last docs PR. I went through our closed PRs and items since then, and I think this is all up-to-date at this point. Any other items that would affect the user manual look to be slated for 0.6, so I think I am done for this release :) .


diegogangl:
Merged, thanks!

PR #598, opened 2021-03-28
debug.sh: Add -p prefix-prog for better control

Neui:
This would allow to easily run custom python modules such as profiling or even an debugger.
I needed to `cd` into the build directory, otherwise it wouldn't find the modules, since at least with `cProfile` and `profile` I get different module path configuration.

Also, since I am not quoting `$prefix_proc`, it could have strange behaviour when the user passes `;` or other "shell characters", but I couldn't find a better way to make an single string to be something more "safe".

### DIFF
--- a/scripts/debug.sh
+++ b/scripts/debug.sh
@@ -12,24 +12,28 @@ norun=0
 pydebug=0
 title=""
 explicit_title=0
+prefix_prog=""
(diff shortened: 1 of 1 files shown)


diegogangl:
Merged, thanks!

PR #606, opened 2021-04-01
scripts/debug.sh: Add install dir while running

Neui:
This makes the icon in the about dialog and probably some other stuff working in the `launch.sh`-script (so *not* when installed, where it works correctly, just when run locally).

Before:
image

After:
image


diegogangl:
Thanks!

PR #607, opened 2021-04-01
debug.sh: Add default title when it is explicitly empty

Neui:
Not specifying `-t` explicitly still uses the default generated development title, but now doing `-t ''` will not add the `--title` argument to GTG and just use the application default one.

Primarily made to being able to create screenshot(s) in the about window without having to manually figure out what the real "title" is without removing the `-t`-argument in the debug script itself.


diegogangl:
Thanks! This script is getting a little complex tho

PR #610, opened 2021-04-02
Fix recurring term not displayed correctly in Editor

zeddo123:
While testing recurring tasks, I experienced a bug that way related to the recurring term in the editor; the title (aka `Every: <term>`) was being generate from `datetime.today()` which means that if you set the task to repeat every week you would have `Every: Friday` (correct) but when you open GTG tomorrow you would have `Every: Saturday` (incorrect) displayed.

To fix it, I added a timestamp for recurring tasks that gets updated whenever the recurring term is updated.

### DIFF
--- a/GTG/core/task.py
+++ b/GTG/core/task.py
@@ -81,6 +81,7 @@ class Task(TreeNode):
 
         # Setting the attributes related to repeating tasks.
         self.recurring_term = None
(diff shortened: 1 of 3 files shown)


diegogangl:
Subtle bug, thanks!

PR #619, opened 2021-04-09
Update the Flatpak runtime version to 40

picsel2:
The previous runtime "is no longer supported as of February 13, 2021."


diegogangl:
> The previous runtime "is no longer supported as of February 13, 2021."


Did we really take that long to release the new version? :)
Thanks!

Issue #618, opened 2021-04-09
Can't open GTG web page from "Getting Started with GTG"


hfiguiere:
I cleared all the data from GTG. (Flatpak version)

When I open GTG there is a default "Getting started task".
I open it. I click on the link at the end to get to the web page.

The console show the following error:
`Failed to open '/home/hub/ https:/wiki.gnome.org/Apps/GTG': No such file or directory`

Nothing else happens.

Note: it tries to open an invalid URL (that doesn't match the text).

diegogangl:
Looks like `webbrowser.open()` is mucking the url for some reason. We should probably use `xdg-open` or something like that from Gio,  we only support linux anyways.

nekohayo:
Just for the record: for a second I thought I had screwed up the contents of the notes to have an invalid URL, but nope, something in the code is failing to launch a perfectly fine URL, as you pointed out above:

```
$ git grep "https://wiki" | wc -l
80 results

$ git grep "https:/wiki" | wc -l
0 results
```

And yet, when I do:

```
$ ipython3 
In [1]: import webbrowser                                                                                                                                                                       
In [2]: webbrowser.open("https://wiki.gnome.org/Apps/GTG/")                                                                                                                                     
Out[2]: True
```

...it opens the page fine in Firefox. So I don't understand how this can fail.

picsel2:
```sh
sebastian@sebastian-desktop$ flatpak run --command=bash org.gnome.GTG

[📦 org.gnome.GTG ~]$ python -m webbrowser -t "https://wiki.gnome.org/Apps/GTG/" # opens the url

[📦 org.gnome.GTG ~]$ python -m webbrowser -t " https://wiki.gnome.org/Apps/GTG/" # fails

[📦 org.gnome.GTG ~]$ Failed to open '/home/sebastian/ https:/wiki.gnome.org/Apps/GTG': No such file or directory
```

The url has a preceding space. That leads to whe webbrowser module to interpret it as a relative path from the current working directory.

picsel2:
I think the [pre part](https://github.com/getting-things-gnome/gtg/blob/ec4dded93749d3613474baa8e7e070d37c73f7e1/GTG/core/urlregex.py#L31) of the urlregex adds the extra space: https://regexr.com/5qdhg

Removing it would solve this issue but I don't understand the purpose of the pre-regex at all.

diegogangl:
Fixed in master

PR #620, opened 2021-04-09
Exclude preceeding whitespace from urls in textview

picsel2:
The urlregex to find urls in the text contained a pattern that also
matched [a space character before the url][1]. This part of the pattern
was removed completely.

[1]: #618


diegogangl:
LGTM, thanks!

PR #639, opened 2021-04-21
Squash some deprecations

Neui:
See commit messages for more info. I basically ran `./launch.sh -w` and fixed some of the DeprecationWarnings it came up. I didn't touch theme/colour related stuff yet.

I am unsure about the "Try to use .get_property instead of .get_font" commit (with the UnicodeError) if I am doing something wrong. A quick `print` tells me `font` is `Pango.FontDescription`, just like what the old `.get_font()` returns. I'd guess it is maybe a object lifetime problem?

Testing appreciated.

### DIFF
--- a/GTG/backends/backend_signals.py
+++ b/GTG/backends/backend_signals.py
@@ -16,7 +16,7 @@
 # this program.  If not, see <http://www.gnu.org/licenses/>.
 # -----------------------------------------------------------------------------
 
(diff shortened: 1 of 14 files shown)


diegogangl (commented):
My bad, this got removed in Python 3.9 but the replacement was there

diegogangl:
Merged, thanks!

PR #641, opened 2021-04-21
Add optional setproctitle to set process title

### LINKED ISSUE #638: Set the app's process name (instead of "python3")
This is purely a wishlist item, and very much a trivial one, but GTG doesn't show up with its name (and icon) in the process list, it just shows up as "python3" (and the accompanying flatpak bubblewrap processes):

foo0

... unlike some other python apps:

foo

It would be neat if it could properly identify itself, in a way that works both when launched from `launch.sh` or from a flatpak. I don't know if/how this is possible, but maybe https://stackoverflow.com/questions/564695/is-there-a-way-to-change-effective-process-name-in-python can help.

Neui:
Uses https://pypi.org/project/setproctitle/.
Made optional since it isn't necessary for the application itself and is basically a use-once thing.
Fixes #638. An icon is even displayed (although it is provided by a 3rd party theme here):
Screenshot of system monitor of the change
The only downside I can see is that it also overwrites the command line thing, so you can't see with what command line it has been started with.

### DIFF
--- a/GTG/gtg.in
+++ b/GTG/gtg.in
@@ -64,6 +64,12 @@ if __name__ == "__main__":
     gettext.bindtextdomain('gtg', LOCALE_DIR)
     gettext.textdomain('gtg')
 
+    try:
+        import setproctitle
+        setproctitle.setproctitle("gtg")
+    except ImportError:
+        pass
+
 import gi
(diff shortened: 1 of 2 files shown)


diegogangl:
Merged!

PR #642, opened 2021-04-22
Update recurring calendar to today or selected date

### LINKED ISSUE #637: Calendar picker widget to select the monthly or yearly recurrence date always default to September 2020
When I open the task editor for a task that was never set to recurring before, and want to set it to repeating **monthly**, or **yearly** (at a specific date), it always shows me the calendar picker widget defaulting to September 20, 2020:

Screenshot from 2021-04-20 10-19-16

...which makes no sense, until you see that every task in the file format defaults to having this:

```
      <recurring enabled="false">
        <term>None</term>
        <updated_date>9999-12-30T00:00:00</updated_date>
      </recurring>
```

Why does it have those lines of fake data if it doesn't even use recurrence, shouldn't those be added only when used? In any case, the date is invalid, and the code shouldn't use it.

More generally, I think the recurring task's date picker should use the month of the task's current due date (if set) or the current month as a fallback.

Neui:
Previously it'll always default to the same year and month, by removing
the attributes it defaults to being today.

In code, it is then updated to match the current recurring status when
applicable.

The monthly selection is now always using January since it has 31 days,
the max days in a month.
To make sure this detail isn't seen when going to the yearly calendar,
this is made hidden with an ugly loop, trying out the next possible
month.

Fixes #637. Yes, the code is just a mess.

### DIFF
--- a/GTG/gtk/data/task_editor.ui
+++ b/GTG/gtk/data/task_editor.ui
@@ -441,8 +441,7 @@
                   <object class="GtkCalendar" id="month_calender">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
-                    <property name="year">2020</property>
-                    <property name="month">11</property>
+                    <property name="month">0</property>
                     <property name="show_heading">False</property>
                     <property name="show_day_names">False</property>
                     <property name="no_month_change">True</property>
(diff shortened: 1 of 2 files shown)


diegogangl:
uhg, we really need a better calendar widget.
Merged tho, thanks!

PR #657, opened 2021-05-01
Use application ID as the window icon

Neui:
Because otherwise in at least KDE it will display the default one.
It uses the application id because you need one anyway for the desktop
file, and would show the development icon version appropriately.

Before: Look at after but all the Icons are "X"
After (in development mode): Screenshot of GTG with various window open and the development icon visible


diegogangl:
Thanks for fixing both of these issues!

Issue #647, opened 2021-04-26
Flatpak version failing to find custom cursor theme


santiagocezar:
Version 0.5 from Flathub doesn't launch when using a custom cursor theme installed by the user. It does work after giving the app permission to read ~/.icons, or by moving the cursor to $XDG_DATA_DIRS (but then it doesn't work in every other host application).

This is the application log

```
Gdk-Message: 11:52:13.791: Unable to load hand2 from the cursor theme
Traceback (most recent call last):
  File "/app/bin/gtg", line 79, in <module>
    from GTG.gtk.application import Application
  File "/app/lib/python3.8/site-packages/GTG/gtk/application.py", line 28, in <module>
    from GTG.gtk.browser.main_window import MainWindow
  File "/app/lib/python3.8/site-packages/GTG/gtk/browser/main_window.py", line 35, in <module>
    from GTG.gtk.browser import quick_add
  File "/app/lib/python3.8/site-packages/GTG/gtk/browser/quick_add.py", line 27, in <module>
    from GTG.gtk.editor.taskview import TAG_REGEX
  File "/app/lib/python3.8/site-packages/GTG/gtk/editor/taskview.py", line 54, in <module>
    class TaskView(Gtk.TextView):
  File "/app/lib/python3.8/site-packages/GTG/gtk/editor/taskview.py", line 78, in TaskView
    CURSOR_HAND = Gdk.Cursor.new(Gdk.CursorType.HAND2)
TypeError: constructor returned NULL
```

After building the Flatpak using latest commit (`2208e62` when writing this issue), the app does work, but it uses a fall-back cursor theme that's missing many cursor sprites. 

![shot1](https://user-images.githubusercontent.com/48573316/116162775-d7438e00-a6cc-11eb-8aca-c6014def27d4.png)

When hovering a link it displays the previous cursor sprite, instead of a 👆...
![shot2](https://user-images.githubusercontent.com/48573316/116151268-c68a1c80-a6ba-11eb-98fd-7cf63af1a7db.png)

and it throws many exceptions when moving the mouse (that are catched by a try-except block)
```
Traceback (most recent call last):
  File "/app/lib/python3.8/site-packages/GTG/gtk/editor/taskview.py", line 616, in on_mouse_move
    cursor = Gdk.Cursor.new_for_display(window.get_display(),
TypeError: constructor returned NULL
``` 

I think adding a `--filesystem=/home/$USER/.icons/:ro` in the flathub manifest could solve this issue

[...]

Issue #621, opened 2021-04-10
0.5 won't launch from app icon until xapp python3-xapp is installed


seraph:
After installing GTG 0.5 from flatpak/flathub on a fresh fedora 33 system:

1. Unable to launch the app using the app icon / gnome launcher
2. `flatpak run org.gnome.GTG -d` worked without xapps python3-xapp being installed
3. `sudo dnf install xapps python3-xapp` fixed launching via app icon

There were no error messages about xapp being missing unlike in issue #616 and it still worked with the `flatpak run -d` command which ... is strange.

Is it missing from the flatpak deps, or is it that xapps needs to be installed on the host system natively?

Thanks. Let me know if you require further details.

diegogangl:
This might be a Gnome 40 platform change, because we don't have any dependencies to python-xapp. Looking at the API there doesn't seem to be anything we're using, even indirectly

nekohayo:
This is weird, because I'm on Fedora 33 as well, I don't have python3-xapp installed, and GTG 0.5 (flatpak version) launches from my GNOME Shell's "dash-to-panel" extension dock without any problems, no matter how many times I shut it down and retry.

seraph:
> This is weird, because I'm on Fedora 33 as well, I don't have python3-xapp installed, and GTG 0.5 (flatpak version) launches from my GNOME Shell's "dash-to-panel" extension dock without any problems, no matter how many times I shut it down and retry. Could it have something to do with some plugin @seraph might have activated?

Plug-in for GTG itself, or for GNOME Shell? If you mean GTG then it was a fresh install and I've never used the app before either so there would be no cruft left from before like plugin settings etc. As for GNOME Shell I just run stuff like app indicator support, workspace indicator and a couple of other things to add items to the top right systray area.

fernandochiquito:
So I'm having the same problem.
For me when I run the command `flatpak run org.gnome.GTG -d` this is the outcome.

`GNOME keyring not found, passwords will be not stored after restarting GTG
Gdk-Message: 08:41:10.751: Unable to load hand2 from the cursor theme
Traceback (most recent call last):
  File "/app/bin/gtg", line 79, in <module>
    from GTG.gtk.application import Application
  File "/app/lib/python3.8/site-packages/GTG/gtk/application.py", line 28, in <module>
    from GTG.gtk.browser.main_window import MainWindow
[...]

PR #658, opened 2021-05-01
 Use Gdk.Cursor.new_from_name()

Neui:
Because GDK4 dropped new_for_display in favour of new_from_name, and it *might* have better cursor theme compatibility.

Also partially see #647 and the linked #621. Not sure if it fixes it.

Also fixed a small flicker in the 2nd commit, that is very small that I think is OK for that PR.

### DIFF
--- a/GTG/gtk/editor/taskview.py
+++ b/GTG/gtk/editor/taskview.py
@@ -597,8 +597,8 @@ class TaskView(Gtk.TextView):
         tags = view.get_iter_at_location(x, y)[1].get_tags()
 
         # Reset cursor and hover states
-        cursor = Gdk.Cursor.new_for_display(window.get_display(),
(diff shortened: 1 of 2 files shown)


diegogangl:
I'm getting this when running:

```
  File "/home/januz/code/gtg/.local_build/install/lib/python3.9/site-packages/GTG/gtk/general_preferences.py", line 86, in get_default_editor_font
    editor_font = font.to_string()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa0 in position 0: invalid start byte
```

diegogangl:
Oh yeah I forgot about that. Well I can't test those issues either, but this doesn't seem to produce any undesirable side effects and is more future proof so merging. Thanks!

PR #659, opened 2021-05-02
Render tag icons better on HiDPI

### LINKED ISSUE #324: Symbolic icons (and color markers) are not HiDPI (retina) friendly
If you launch `GDK_SCALE=2 ./gtg.sh`, this is what you get:

Screenshot from 2020-05-25 08-53-11

As you can see, some things are fuzzy. Not quite sure how that should be fixed, but maybe part of Diego's liblarch listbox & sidebar rework...

Neui:
Antialiasing was disabled, so the corners of the color boxed weren't as
smooth. Additionally, the correct icons are now used and rendered.
Unfortunately not GTK4-ready, since there has been some changed around
rendering icons it seems.

Fixes #324

grafik

### DIFF
--- a/GTG/gtk/browser/cell_renderer_tags.py
+++ b/GTG/gtk/browser/cell_renderer_tags.py
@@ -122,7 +122,6 @@ class CellRendererTags(Gtk.CellRenderer):
 
         # Drawing context
         gdkcontext = cr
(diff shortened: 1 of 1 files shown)


diegogangl:
Looks like we were using some old APIs there, thanks!

Issue #653, opened 2021-04-30
Quick entry keeps due dates in title but trims them sometimes


morhe-blip:
Description:
When using quick entry with a due date, the due agrument is kept in the title. Sometimes, the due argument is trimmed.
Current master(commit c53bf0ee73bfec9c87570f7f9afdbd6ea9e5ff5a)

Reproduction:
1. Create a task "Test tags:test due:now" via quick entry and it creates a task "Test due:now".
2. Create a task "Test tags:test due:soon" via quick entry and it creates a task "Test due:soon".
3. Create a task "Test tags:test due:tomorrow" via quick entry and it creates a task "Test due:tomor".
4. Create a task "Test tags:test due:2021-05-13" via quick entry and it creates a task "Test due:2021".

Issue #652, opened 2021-04-30
Quick entry does not recognize translated string for "due" (german translation)


morhe-blip:
I found some strange behaviour of the quick entry field (commit c53bf0ee73bfec9c87570f7f9afdbd6ea9e5ff5a). I will open a separate issue for each observation. If this is wrong, feel free to put them into one issue.

Description:
Using the quick entry (with german translation) the keyword "fällig" does not create a task with this mentioned due date. However if using "due", it works.

Reproduction:
Create a task "Test tags:test fällig:2021-04-30" via quick entry, it creates a task "Test fällig:20" with the tag test (which I inserted for finding the tasks afterwards).
The trimming of the date is another issue mentioned separately.

If this is just a stupid issue because "due" is not intended to be used translated in quick entry (which kind of is the same with "tags" I think), close it.

Neui:
> I found some strange behaviour of the quick entry field (commit c53bf0e).

Do you mean since that commit or was that the commit you just tested on?

> If this is just a stupid issue because "due" is not intended to be used translated in quick entry (which kind of is the same with "tags" I think), close it.

Showing a german interface but then not accepting german input is in my opinion at least wrong.
It is supposed to be translated (when I looked at the code last time), and it worked before when I tested the german translation, so this is a valid regression.

morhe-blip:
It is the commit I tested this on. I think, that I saw this behaviour before, but I did not test it at that time.

PR #660, opened 2021-05-04
don't use invalid ranges in quick add

eladyn:
The `re.finditer()` function yields matches of the expression on the original passed string. When this string is modified and truncated later, the subsequent matches return invalid ranges.

This should fix bugs like #653 and #652.

I decided to implement this with this `while True` loop with the break condition. If you think that there is a better way, let me known. I'm always happy to learn :grinning:.

### DIFF
--- a/GTG/gtk/browser/quick_add.py
+++ b/GTG/gtk/browser/quick_add.py
@@ -69,7 +69,10 @@ def parse(text: str) -> Dict:
         data = match.group(0)
         result['tags'].add(data[1:])
 
-    for match in re.finditer(TOKEN_REGEX, text):
+    while True:
+        match = re.search(TOKEN_REGEX, text)
+        if match is None:
+            break
         token = match.group(2)
         data = match.group(3)
 


GTG/gtk/browser/quick_add.py
```diff
@@ -69,7 +69,10 @@ def parse(text: str) -> Dict:
         data = match.group(0)
         result['tags'].add(data[1:])
 
-    for match in re.finditer(TOKEN_REGEX, text):
+    while True:
+        match = re.search(TOKEN_REGEX, text)
+        if match is None:
+            break
```
Neui, line 75: I believe you can use

```python3
while match := re.search(TOKEN_REGEX, text):
```

Neui:
In the current state, it fixes #653 but *not* #652. The latter issue is about not being able to use the translated version of the "tags" (the `_(...)` above the code, to which I think I just found the cause).

eladyn:
Thanks for the correction. Would be a bit strange if the same person had reported the same bug twice :grin:. I'll change the description accordingly

GTG/gtk/browser/quick_add.py
eladyn: Oh, cool! I didn't know this one.

Neui (approved):
Looks good to me and fixes the issue.

diegogangl:
That's quite a subtle bug, thanks for squashing it!

PR #661, opened 2021-05-04
Add quick_add to POTFILES so translated parsing works

### LINKED ISSUE #652: Quick entry does not recognize translated string for "due" (german translation)

Neui:
I've noticed that `_("due")` doesn't return the translated version, and looking into the language files it was commented out (by https://github.com/getting-things-gnome/gtg/commit/60be39b6d). I found out that `quick_add.py` wasn't in the `POTFILES.in`, so I added it and updates the PO files in the 2nd commit.

Fixes #652


diegogangl:
Thanks!

PR #662, opened 2021-05-04
Make sidebar in menu a toggle

Neui:
Because other applications like nautilus do something similar.
Also makes the code a bit cleaner because the synchronisation is done by
GObject rather than doing it ourselves.

The F9 shortcut still works. Note that the video shows a bit of an older version, the newest has put the checkbox to the right like other GTK applications.
Demonstration


diegogangl:
Man, I wanted to do this when I started messing with the GTK3 port ages ago but couldn't figure it out at that time. Looks like the other PRs caused some conflicts though.

Neui:
I guessed that'll happen. Rebased and fixed the conflicts, so it should merge/rebase cleanly.

diegogangl:
Merged, thanks!

PR #663, opened 2021-05-05
ESC now closes calendar picker window

### LINKED ISSUE #382: ESC doesn't close the dedicated date selector window
1. Right click on an open task
2. Set start/due date → select date...
3. Press ESC

Expected: Window will close, no change is being made

Actual: Window doesn't close

I am wondering if maybe the popup, when you click onto one of the date fields when editing a task, could be reused rather than opening an window.

Neui:
Copied the code used in the tag editor.

Fixes #382

Unfortunately not GTK4-ready since handling of shortcuts etc. has been changed in GTK4.


diegogangl:
Thanks!

PR #664, opened 2021-05-05
Add F10 shortcut to open the main menu button

### LINKED ISSUE #532: Bind F10 keyboard shortcut to the menubuttons
It seems like F10 is the de-facto standard among CSD-using applications to toggle the main menu button of a window (we have that in the main GTG window, but should we also bind that shortcut for individual task windows? that's probably a bit more complicated to manage cleanly, and I'm not sure if it's worth it for the "Dismiss" and "Delete" actions that have their own shortcuts anyway).

I don't know if every application hardcodes F10 or if there is a way to get that info from the system in a standardized way...

Neui:
Copying other GNOME applications behaviour.

Fixes #532. Seems to be hardcoded to F10.


diegogangl:
Works as advertised, thanks!

PR #665, opened 2021-05-05
Fix certain main menu entries not being selectable via keyboard

Neui:
Before while pressing the down and up arrow keys:

https://user-images.githubusercontent.com/1196130/117217958-93563480-ae02-11eb-8104-69c85d242f7d.mp4

After: What you would expect how it should behave.

### DIFF
--- a/GTG/gtk/data/main_window.ui
+++ b/GTG/gtk/data/main_window.ui
@@ -213,7 +213,6 @@
         <child>
           <object class="GtkModelButton" id="toggle_sidebar_button">
             <property name="visible">True</property>
-            <property name="can_focus">False</property>
             <property name="receives_default">True</property>
             <property name="action_name">win.toggle_sidebar</property>
             <property name="text" translatable="yes">Toggle Sidebar</property>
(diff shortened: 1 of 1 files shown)


diegogangl:
I wonder why it had different focus properties :thinking: , I guess that's why GNOME devs don't recommend Glade

Neui:
> I wonder why it had different focus properties :thinking:

It's an checkbox in glade, likely someone accidentally checked it but saw no visible change (mouse still works), so they probably left it as-is.

> I guess that's why GNOME devs don't recommend Glade

Glade unfortunately doesn't support everything (like I looked at Gtk.FlowBox, which Glade just didn't support showing/interacting with it), and it rewrites the entire file, loosing information it doesn't "support" (like [action-target](https://gitlab.gnome.org/GNOME/glade/-/issues/330) or [properties bindings](https://gitlab.gnome.org/GNOME/glade/-/issues/340), although it seems the latter has a merged PR to fix that)

diegogangl:
Yup, I want to comb through the ui files by hand one day (maybe for the Gtk4 port). Or using that specialized editor I saw on planet gnome once

PR #666, opened 2021-05-07
New tag editor window

### LINKED ISSUE #276: Tag editor redesign
The tag editor needs some lovin'

- It saves data "live", it would be better if it had headerbar buttons to cancel and save
- The generate color function should have a button in this dialog. Let's just call it "Random Color"
- Make it look less GTK2-ish :)

### LINKED ISSUE #306: Stability + performance: only apply changes from the tag editor dialog when actually OKing the changes
When you click "Edit tag" in the sidebar and proceed to renaming a tag with the dialog that shows up, the renaming process is very slow and expensive, particularly when you have hundreds of tasks containing that tag. It seems like it tries to apply changes everytime you type a character, which is stupid:

* It destroys performance
* It is an application stability and data integrity risk. For example, what happens if you just delete the whole text string? Or replace it too quickly? It happened to me while testing (and previously in the past) that renaming a tag actually duplicated it into a new tag because the app couldn't keep up with my typing.

This is almost identical in spirit to #270, except that in this case (the tag editor dialog) I see no reason why there should be live applying of the changes at all, not even with a "timer". That dialog can afford to simply have standard "Cancel / Apply" buttons and to _only_ apply the changes once you definitely click the _Apply_ button (maybe with a progress bar indication, if applying the changes for multi-hundred items is still is expensive even after a switch to lxml).

Neui:
Based on a design I saw that diegogangl made.
I made some change (like a remove button), but I am not too happy.
Additionally, you now require to apply the changes manually rather than
it being done automatically to prevent some possibly weird bugs.
Maybe this can be changed later, but not now.

Rather than constructing it in code, it is now done via a .ui-file,
initially generated by glade.
Thus a lot of the tag editor window code was essentially remade. This
time, I used GObject properties to be able to connect them with the
widget appropriately.

Fixes #306
~~Based on #276, but not marking as being fixed for a possibly better design. However, this PR already lays the foundation, so hacking on it will be easier.~~
Fixes #276 since I've got feedback and iterations

I've noticed that renaming tags is buggy (won't always work, once it even created an duplicate tag that caused GTG to error while reading the tasks xml file, but this is outside of this PR).

https://user-images.githubusercontent.com/1196130/117505864-0687b400-af85-11eb-972e-fca28f4d07d7.mp4

### DIFF
--- a/GTG/gtk/data/tag_editor.ui
+++ b/GTG/gtk/data/tag_editor.ui
@@ -243,7 +243,6 @@
                     <property name="can_focus">True</property>
                     <property name="receives_default">True</property>
                     <property name="tooltip_text" translatable="yes">Change the color of the tag</property>
-                    <property name="halign">end</property>
                     <signal name="color-set" handler="activate_color" swapped="no"/>
                   </object>
                   <packing>
--- a/GTG/gtk/data/tag_editor.ui
+++ b/GTG/gtk/data/tag_editor.ui
@@ -1,0 +1,265 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.22.2 -->
+<interface>
+  <requires lib="gtk+" version="3.20"/>
+  <object class="GtkHeaderBar" id="headerbar">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <property name="title" translatable="yes" comments="Edit tag windwo title">Edit %s</property>
+    <property name="has_subtitle">False</property>
+    <child>
+      <object class="GtkButton" id="cancel">
+        <property name="label" translatable="yes">Cancel</property>
+        <property name="visible">True</property>
+        <property name="can_focus">True</property>
+        <property name="receives_default">True</property>
+        <signal name="clicked" handler="cancel" swapped="no"/>
+      </object>
+    </child>
+    <child>
+      <object class="GtkButton" id="apply">
+        <property name="label" translatable="yes">Apply</property>
+        <property name="visible">True</property>
+        <property name="can_focus">True</property>
+        <property name="can_default">True</property>
+        <property name="receives_default">True</property>
+        <signal name="clicked" handler="apply" swapped="no"/>
+        <style>
+          <class name="suggested-action"/>
+        </style>
+      </object>
+      <packing>
+        <property name="pack_type">end</property>
+        <property name="position">1</property>
+      </packing>
+    </child>
+  </object>
+  <object class="GtkBox" id="main">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <property name="margin_left">18</property>
+    <property name="margin_right">18</property>
+    <property name="margin_top">18</property>
+    <property name="margin_bottom">18</property>
+    <child>
+      <object class="GtkBox">
+        <property name="visible">True</property>
+        <property name="can_focus">False</property>
+        <property name="orientation">vertical</property>
+        <child>
+          <object class="GtkBox">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="halign">center</property>
+            <child>
+              <object class="GtkButton" id="icon-button">
+                <property name="label" translatable="yes">🏷️</property>
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="receives_default">True</property>
+                <property name="tooltip_text" translatable="yes">Click here to set an icon for this tag</property>
+                <property name="halign">center</property>
+                <property name="valign">start</property>
+                <property name="always_show_image">True</property>
+                <signal name="clicked" handler="set_icon" swapped="no"/>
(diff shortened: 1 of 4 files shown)


diegogangl:
You've been on fire lately :D
I thought about using a template for this window, maybe it's something we can hack later? I think it would simplify the code quite a bit.

I'm getting ```(gtg:126336): Gtk-CRITICAL **: 21:02:06.665: gtk_window_add_accel_group: assertion 'GTK_IS_WINDOW (window)' failed``` when opening.

Nitpicks:

- Add the `.linked` class to the color and random buttons.
```
                    <style>
                      <class name="linked"/>
                    </style>
```
- The window should start a little larger so the emoji popup is fully visible. Otherwise users would have to resize the window every time they want to change the icon. This is a Gtk limitation on X11 that is fixed in Gtk4  (IIRC).
- Since you added the "color" switcher, why not drop the "remove icon" button (that looks sort of out of place) and have an "icon" switcher? 
- In that case, we could move the color buttons below the icon selector so those kinds of options are grouped
- The icon button could be a bit larger to accomodate the color buttons below
- If the "color" switcher is disabled, then the color buttons should be disabled too
- Add some padding [following the HIG](https://developer.gnome.org/hig/stable/visual-layout.html.en). 18px between name an actionable, 12px between the icon selector and the other column, 6px between switchers

Code LGTM, I hadn't seen that `GObject.property` thing before. Nice one!

diegogangl:
Also, you got the devil's PR number :metal:

Neui:
Ok I've implemented the changes as far as I can see (+ `linked` class, just forgot to mention in the commit).

What I don't really like is that the colour buttons get disabled when colour is disabled, since now you need an additional step to set an colour rather than just pressing the button.

I did that for the more prominent icon button (clicking on it automatically checks the icon), which I think is better UX.

https://user-images.githubusercontent.com/1196130/117555496-fe507700-b05f-11eb-9688-876dc869d3ef.mp4

(A bit laggy because of wifi latency)

diegogangl:
I found the glade file I used to make that mockup: https://gist.github.com/diegogangl/20ef13dc3ee32eedd098967704e8b7f1

You need to set `linked` to the container of the buttons, sorry.
Let's try an alternative (if you want), keep the color buttons as they were in the mockup but add a button next to random with the `user-trash-symbolic` icon. Pressing that button would remove the color (and set the swatch to white).
We could do the same with the icon, place a button with the trash icon centered below it.
I think that would be more explicit in the sense that the user would expect the color/icon to be removed. 

Right now, the switches make it seem like you can switch them on/off without losing your previous selection. The design would also look a bit more balanced.

Neui:
Seems better now: (mis-aligned color button is fixed, just not in the following video)

https://user-images.githubusercontent.com/1196130/117595087-097ad400-b140-11eb-9cd6-c3c1a2582959.mp4

diegogangl:
Yes! Looking good now :)
Two final questions:
- Does the color button row still have the 8px padding? Seems too close to the actionable switcher
- Would it be possible to veritcally align the name input, actionable switcher and color buttons row?

Neui:
> Does the color button row still have the 8px padding? Seems too close to the actionable switcher

You mean 6px? [According to the properties](
https://github.com/Neui/gtg/blob/d3347564849f998b646861b1c71a3add3bac3bdb/GTG/gtk/data/tag_editor.ui#L193), the row spacing is 6px and seems to be correct when measuring the pixels in an image editing application:

```xml
            <property name="row_spacing">6</property>
```

> Would it be possible to veritcally align the name input, actionable switcher and color buttons row?

Yes, glade wasn't really showing the margins correctly, so it went unnoticed. Also double-checked and fixed some other padding issues.

Tag editor window

(Yes, between the icon button and the Actionable label is indeed 12 pixels)

diegogangl:
Seems like `margin-left` and `margin-right` were changed to `margin-start` and `margin-end` at some point, which is why the icon selector looks so close to the other column. Everything else looks good and ready to go in :+1:

GTG/gtk/data/tag_editor.ui
```diff
@@ -0,0 +1,320 @@
...
+    <property name="margin_bottom">18</property>
+    <property name="spacing">12</property>
+    <child>
+      <object class="GtkBox">
+        <property name="visible">True</property>
+        <property name="can_focus">False</property>
+        <property name="halign">center</property>
+        <property name="valign">start</property>
+        <property name="orientation">vertical</property>
+        <property name="spacing">6</property>
```
diegogangl, line 52:
```suggestion
        <property name="spacing">6</property>
        <property name="margin-end">18</property>
```

diegogangl (commented):
See the suggested change. I probably explained it incorrectly, I was talking about the icon selection column being too close to the column that has all the switchers. 
Here's a screen with that change:

Screenshot from 2021-05-12 22-20-39

diegogangl:
Awesome, thanks for the work! Also I noticed the window starts small and gets larger when you open the emojis selector, that's pretty clever :)

PR #671, opened 2021-05-10
Fix cut-off when you expand the columns too much

### LINKED ISSUE #669: Columns can get hidden outside the bounds of the window
I was using GTG on a 1080p output, and the Due column was at the far right. When I moved the window to a "768p" output, the column was completely hidden, and scrolling horizontally is not possible.

Possible fixes:
- adding a horizontal scrollbar to this view;
- constraining columns' position by automatically resizing to fit inside the window bounds.

Neui:
Scrollbar policy "Never" means that it'll let the widget expand, not
sure why it was inserted there.
The default value works as expected.

Fixes #669. Note that the columns aren't re-adjusted when the window is resized, it just allows you to scroll.

https://user-images.githubusercontent.com/1196130/117730588-db070280-b1ec-11eb-87e1-bcc1231266ec.mp4

### DIFF
--- a/GTG/gtk/data/main_window.ui
+++ b/GTG/gtk/data/main_window.ui
@@ -503,7 +503,6 @@
                       <object class="GtkScrolledWindow" id="open_pane">
                         <property name="visible">True</property>
                         <property name="can_focus">True</property>
-                        <property name="hscrollbar_policy">never</property>
                         <child>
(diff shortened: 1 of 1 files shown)


diegogangl:
Ah ok :+1: , merged then!

PR #672, opened 2021-05-10
Fix gamfiy bugs

### LINKED ISSUE #654: Gamify does not count subtasks that were marked as done by using the checkbox inside parent task
Description and reproduction:

Create a task with one or more subtasks. Tick the checkbox inside the parent tasks description. The task is marked as done, however the counter is not increased.

If the subtasks are marked as done via right-click or Ctrl-D, it works.

### LINKED ISSUE #655: Gamify does not count subtasks when closing the parent task
Description and reproduction:

Create a task with with n different subtasks. Mark the parent task as done with n open subtasks. The counter is increased by one instead of n (or n+1 don't know how this is counted)

zeddo123:
Fixes #654 
Fixes #655

### DIFF
--- a/GTG/core/requester.py
+++ b/GTG/core/requester.py
@@ -36,6 +36,7 @@ class Requester(GObject.GObject):
     Multiple L{Requester}s can exist on the same datastore, so they should
     never have state of their own.
     """
+    __gsignals__ = {'status-changed':(GObject.SignalFlags.RUN_FIRST, None, (str, str,))}
 
     def __init__(self, datastore, global_conf):
         """Construct a L{Requester}."""
--- a/GTG/plugins/gamify/gamify.py
+++ b/GTG/plugins/gamify/gamify.py
@@ -10,6 +10,8 @@ from gi.repository import Gtk
 from gettext import gettext as _
 from gettext import ngettext
 
+from GTG.core.task import Task
+
 log = logging.getLogger(__name__)
 class Gamify:
     PLUGIN_PATH = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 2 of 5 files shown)


GTG/core/requester.py
```diff
@@ -36,6 +36,10 @@ class Requester(GObject.GObject):
     Multiple L{Requester}s can exist on the same datastore, so they should
     never have state of their own.
     """
+    __string_signal__ = (GObject.SignalFlags.RUN_FIRST, None, (str, str,))
+    __none_signal__ = (GObject.SignalFlags.RUN_FIRST, None, tuple())
```
Neui, line 40: Never used?

GTG/core/requester.py
```diff
@@ -36,6 +36,10 @@ class Requester(GObject.GObject):
     Multiple L{Requester}s can exist on the same datastore, so they should
     never have state of their own.
     """
+    __string_signal__ = (GObject.SignalFlags.RUN_FIRST, None, (str, str,))
```
Neui, line 39: Why not inline it below?

GTG/plugins/gamify/gamify.py
```diff
@@ -190,7 +188,13 @@ def get_number_of_tasks(self):
     def get_streak(self):
         return self.data['streak']
 
-    def on_marked_as_done(self, sender, task_id):
+    def on_status_changed(self, sender, task_id, status):
+        if status == "Done":
```
Neui, line 192: I'd use `STA_DONE` since this is what other code uses

Neui (commented):
Not tested the change, but the modifications look ok

GTG/core/requester.py
Neui: I mean something like

```python3
    __gsignals__ = {'status-changed': (GObject.SignalFlags.RUN_FIRST, None, (str, str,))}
```

diegogangl:
LGTM, thanks! (and thanks @Neui for the review)

PR #674, opened 2021-05-13
TagEditor: Use Gtk.Template for less boilerplate

Neui:
From: https://github.com/getting-things-gnome/gtg/pull/666#issuecomment-834890560

> I thought about using a template for this window, maybe it's something
> we can hack later? I think it would simplify the code quite a bit.

It does, diegogangl.

Note that most of the diff in the UI-file is some indentation change
since instead of separate objects, they are directly children of the
template window.


diegogangl:
Thought so :D 
Thanks!

PR #676, opened 2021-05-14
Migrate hamster plugin to use GIO dbus

Neui:
~~Marking this as a draft since this is 'WIP'.~~

~~Primarily, it needs testing whenever it still works (and whenever catching the exception that way is correct). Ugh~~

Ready for review now. I believe this is the last instance the old dbus python package is being used, so I think we can remove that dependency after merging this PR.

### DIFF
--- a/GTG/plugins/hamster/hamster.py
+++ b/GTG/plugins/hamster/hamster.py
@@ -22,8 +22,7 @@ import time
 from calendar import timegm
 from gettext import gettext as _
 
-import dbus
-from gi.repository import Gtk
+from gi.repository import GLib, Gtk, Gio
 
 from GTG.core.task import Task
 from GTG.plugins.hamster.helper import FactBuilder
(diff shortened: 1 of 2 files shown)


Neui:
Ok I managed to get myself Hamster installed and tested (and fix) various functionalities I could find.

CC @flavin for review because you last updated the plugin (#465).

flavin (approved):
@Neui  LGTM

nekohayo:
Hi @Neui, please rebase this for 0.7 :)

pentux-GitHub:
This is still relevant, and it's close. A quick status to help it move:

The Hamster plugin was just ported to GTK4 and the new core in #1287, but that PR deliberately left the D-Bus layer alone. Your change here is still the missing piece: hamster.py and helper.py currently keep 6 references to the old `dbus-python` package, and this PR is exactly what removes that last dependency.

@Neui, if you have the bandwidth to rebase on current master (nekohayo asked back in 2024), I think this could finally land cleanly on top of the port. It'd be good to close the loop on `dbus-python`.

PR #685, opened 2021-06-22
Fixing backend loading

jaesivsm:
Backend loading imports all the files in `GTG/backends/` and, if the import succeed, consider the module to be one that contains an actual Backend implementation. It's false at least for `backend_signals`. 
I try to fix that approach by only listing as an actual backend module the one that contains a subclass of `GenericBackend` to some extent.

### DIFF
--- a/GTG/backends/__init__.py
+++ b/GTG/backends/__init__.py
@@ -46,8 +46,8 @@ class BackendFactory(Borg):
     BACKEND_PREFIX = "backend_"
 
     def __init__(self):
-        """
-         Creates a dictionary of the currently available backend modules
+        """ Creates a dictionary of the modules containing a GenericBackend
+        subclass (which are all available backends)
         """
         super().__init__()
         if hasattr(self, "backend_modules"):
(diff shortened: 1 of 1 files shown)


nekohayo:
I think this is a bit too fancy/abstract for me, probably @Neui or @diegogangl would be better suited to review this :)

Neui (commented):
I am not entirely sure what "problem" it wants to solve. Also, you might want to write the description into the commit message itself.

jaesivsm:
This traceback when loading backend list in the synchronization pop up : 

```
  File "/home/jaes/devel/gtg/.local_build/install/lib/python3.9/site-packages/GTG/backends/__init__.py", line 66, in __init__
    sys.modules[extended_module_name].Backend
AttributeError: module 'GTG.backends.backend_signals' has no attribute 'Backend'
```

jaesivsm:
Addendum : the error above is raised because the line `sys.modules[extended_module_name].Backend` was added to force the backend loading mechanism to explicitly fail on loading non-backend module (which is the error I'm fixing with this PR) in the branch I cloned to work on caldav integration. You won't see it without caldav, but basically you won't see it unless you try loading backends.

To sum it up, previous behavior is :
* backend loading mechanism list files in `GTG/backends/` and `__import__` them 
* if loading is successfull, it assumes it's a backend module and adds it to the list of loaded backends
* because of this `backend_signals.py`, which isn't a backend module (doesn't not inherit from `GenericBackend`), is considered as such and various error ensues

New behavior : 
* backend loading mechanism list files in `GTG/backends/` and `__import__` them
* once it's done, all subclasses of `GenericBackend` with a declared `_general_description` value are registered as available backend

diegogangl:
This is a useful improvement, but it's not easy to tell what's going on without the explanation in the previous comment.  My suggestion would be:

- Paste that comment in the commit message (so we have a log of why it was added)
- Add a docstring explaining what the new method is doing

diegogangl:
Merged, thanks!

zippy-zebu:
@jaesivsm Hi, Can you fix the backend loading for 0.5 version? I am on Debian stable and I can't update now. A patch would be sufficient. Would this pull be enough ? Thanks

zippy-zebu:
@jaesivsm I still get backend selection disabled in backend editor window with this.

Issue #550, opened 2021-01-26
Improve consistency between the "Start Tomorrow" headerbar deferment menu and the right-click task deferment menu


nekohayo:
A spiritual successor to #244, with some usability nitpicks that have been bugging me ever since, but that I never found to be critical enough to consider it a big problem, hence the low-priority tag on this ticket. 244's last comments also hints at some technological blockers that may be a problem for a more advanced implementation, but I guess this ticket here is more about harmonizing the labels and the ordering of the items, so probably would work no matter what widgets we use.

What's been bothering me a lot in daily use so far has been the inconsistency between the headerbar's deferment menubutton:

![Screenshot from 2020-03-28 20-36-07](https://user-images.githubusercontent.com/264829/77836319-fd96a300-7133-11ea-9528-7ebc383bb26e.png)

...and the contextual (right-click) deferment menu:

![foo](https://user-images.githubusercontent.com/479401/105856398-35ba7c80-5fb7-11eb-99b8-472967214233.png)

The options present, are not identical from one to another. I think mostly it boils down to:
* One mentions week day names, the other mentions a number of days. I'd like to have both (or if I can only have one, "week day names"), labelled identically in both places.
* One has "Today", the other doesn't. I still think "Today" makes little sense when you have "Clear Date" and would suggest removing it (and the separator that follows it) for consistency.
* There should be a separator before "Pick a Date..." 

I wonder if those menu item labels allow having span styling on text, i.e. something where we could make a portion of a text label only 70% opaque, so that you could have for example "Friday (in 2 days)" with the parenthesis part faded out a little bit? This is probably a little bit nightmarish from a translatable plurals strings perspective (i.e. how do you assemble/mangle the strings if at all) though. But it could open the opportunity to have the best of both worlds, including being able to have more precision for the context-menu-specific actions by having semi-transparent things in parentheses, like for example "Next Week (in 7 days)", "Next Month (Feb 26th)", "Next Year (Jan 26th)"...

Out of curiosity, I went to check how the popular "Todoist" web app presents it, and this is how they lay things out in their popover widget:

![todoist scheduling popover](https://user-images.githubusercontent.com/479401/105858347-36541280-5fb9-11eb-9c37-6df4df2396a9.png)

[...]

PR #686, opened 2021-06-30
Make setting due date today an option in context menu

Smittyvb:
This is useful for when you want to reschedule future tasks to today, or bump forward overdue tasks to today.

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -188,6 +188,7 @@ class MainWindow(Gtk.ApplicationWindow):
             ('start_next_year', self.on_start_for_next_year, None),
             ('start_custom', self.on_start_for_specific_date, None),
             ('start_clear', self.on_start_clear, None),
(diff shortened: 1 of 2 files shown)


nekohayo:
Hi, had you seen issue #550? There's some nuances in there, where I thought, regarding the start dates:

>  "Today" [as a start date] makes little sense when you have "Clear Date" and would suggest removing it (and the separator that follows it) [from the popover menu from the headerbar] for consistency [with the right-click menus].

For the due dates however, I agree that "Today" is missing from the right-click menu. I didn't think it would be a huge deal when I saw it being removed last year, but I can definitely say that in practice it happens more often than not that I want to quickly retarget the due date of a task to today...

Smittyvb:
@nekohayo I've updated the PR to just change the due date menu

Neui (approved):
Looks ok code-wise and works.

diegogangl:
Merged, thanks!

PR #689, opened 2021-07-18
Cleanup the Cell Renderer Codebase

andersonjwan:
Hello, this is my first ever contribution to an open-source project, and the prospect of GTG seems very promising. I thought I'd start small with some code formatting cleanup and better self-documentation for readability.

If you have any suggestions, please let me know. I will be more than willing to adjust as necessary. Here's to my first contribution, hopefully! 🎉

The following errors and warnings according to PyCodeStyle have been fixed:

- E402 (import order)
- E261 (comment spacing)
- E303 (too many blank lines)
- W504 (line break after a binary operator)
- E222 (too many spaces after operator)
- E305 (expected 2 blank lines after class declaration)

### DIFF
--- a/GTG/gtk/browser/cell_renderer_tags.py
+++ b/GTG/gtk/browser/cell_renderer_tags.py
@@ -49,9 +49,7 @@ class CellRendererTags(Gtk.CellRenderer):
         #  G       D
         #   F  *  E
 
-        context.move_to(
-            x + r, y
-        )  # Move to A
+        context.move_to(x + r, y)          # Move to A
         context.line_to(x + w - r, y)      # Line to B
 
         context.curve_to(
--- a/GTG/gtk/browser/cell_renderer_tags.py
+++ b/GTG/gtk/browser/cell_renderer_tags.py
@@ -228,4 +228,5 @@ class CellRendererTags(Gtk.CellRenderer):
         else:
             return (self.xpad, self.ypad, self.xpad * 2, self.ypad * 2)
 
+
 GObject.type_register(CellRendererTags)
--- a/GTG/gtk/browser/cell_renderer_tags.py
+++ b/GTG/gtk/browser/cell_renderer_tags.py
@@ -109,7 +109,6 @@ class CellRendererTags(Gtk.CellRenderer):
         self.config = config
         self._ignore_icon_error_for = set()
 
-
     def do_set_property(self, pspec, value):
         if pspec.name == "tag-list":
             self.tag_list = value
--- a/GTG/gtk/browser/cell_renderer_tags.py
+++ b/GTG/gtk/browser/cell_renderer_tags.py
@@ -51,35 +51,35 @@ class CellRendererTags(Gtk.CellRenderer):
 
         context.move_to(
             x + r, y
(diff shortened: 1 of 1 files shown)


GTG/gtk/browser/cell_renderer_tags.py
```diff
@@ -42,24 +43,43 @@ class CellRendererTags(Gtk.CellRenderer):
...
-        context.line_to(x + w, y + h - r)                      # Move to D
+        #   A  *  BQ
+        #  H       C
+        #  *       *
+        #  G       D
+        #   F  *  E
+
+        context.move_to(
+            x + r, y
+        )  # Move to A
```
Neui, line 54: I think you wanted to format like the following `curve_to` calls, but it doesn't have many parameters and maybe could easily led to make one think it is a `curve_to` call. It's not too bad tho.

Neui (approved):
Looks OK and still works (locally tested).

I personally would prefer if such style commits were squashed together into one, since they are usually uninteresting and having many style-commits doesn't really help trying to find behaviour differences.

Also, we have a [`pylint.rc`](https://github.com/getting-things-gnome/gtg/blob/master/pylint.rc) configuration file. I'm not into the ecosystem, but I wonder about the "compatibility" with PyCodeStyle, like does it respect that file? Doesn't seems so [according to the documentation](https://pycodestyle.pycqa.org/en/latest/intro.html#configuration). We should decide on what we should use.
(Also, I need to check if my editor respects it or whatever linter my editor uses...)

andersonjwan:
> I personally would prefer if such style commits were squashed together into one, since they are usually uninteresting and having many style-commits doesn't really help trying to find behaviour differences.

Understood. I wanted to be as clear as possible with my intentions, but from here forward, for more stylistic and formatting changes, I will group into a single commit that covers each error/warning fixed to simplify the review process.

> Also, we have a [`pylint.rc`](https://github.com/getting-things-gnome/gtg/blob/master/pylint.rc) configuration file. I'm not into the ecosystem, but I wonder about the "compatibility" with PyCodeStyle, like does it respect that file? Doesn't seems so [according to the documentation](https://pycodestyle.pycqa.org/en/latest/intro.html#configuration). We should decide on what we should use.

I was originally using the command below (found [here](https://github.com/getting-things-gnome/gtg/blob/master/docs/contributors/coding%20best%20practices.md#checking-the-coding-style) in the docs) to determine the set of formatting errors. I also just went ahead and installed pylint and setup the proper linting to be based off of the `pylint.rc` file, and I do not receive any formatting errors - only argument/variable usage warnings.

```sh
find . -name '*.py' -print0 | xargs -0 pycodestyle --repeat --max-line-length=100 --ignore=E128
```

If there are any further checkers/configurations that I should verify, please let me know. Thank you!

GTG/gtk/browser/cell_renderer_tags.py
andersonjwan: I agree. I've made the changes and pushed them.

diegogangl:
LGTM, thanks! Congrats on your first contribution :beers:

PR #690, opened 2021-07-21
Demote default editor font log error to a warning

Neui:
Caused some confusion to think it could have something to do with some issue (like https://github.com/getting-things-gnome/gtg/pull/658#issuecomment-834830045), but it doesn't.
Thus demoted to a simple warning, which isn't shown by default (without `-d`).


diegogangl:
Merged, thanks!

PR #691, opened 2021-07-23
Fix PEP8 errors

odoood:
No functionality changes, only formatting changes to clear errors reported by `make pep8`

### DIFF
--- a/GTG/backends/generic_backend.py
+++ b/GTG/backends/generic_backend.py
@@ -276,12 +276,14 @@ class GenericBackend():
             # if it's not specified, then this is the default backend
             # (for retro-compatibility with the GTG 0.2 series)
             parameters[self.KEY_DEFAULT_BACKEND] = True
+
         # default backends should get all the tasks
-        if parameters[self.KEY_DEFAULT_BACKEND] or \
-                (self.KEY_ATTACHED_TAGS not in parameters and
-                 self._general_description[self.BACKEND_TYPE] ==
-                    self.TYPE_READWRITE):
+        attach_tags = self.KEY_ATTACHED_TAGS not in parameters and \
+            self._general_description[self.BACKEND_TYPE] == self.TYPE_READWRITE
+
+        if parameters[self.KEY_DEFAULT_BACKEND] or attach_tags:
             parameters[self.KEY_ATTACHED_TAGS] = [ALLTASKS_TAG]
+
         self._parameters = parameters
         self._signal_manager = BackendSignals()
         self._is_initialized = False
@@ -311,7 +313,7 @@ class GenericBackend():
             return [ALLTASKS_TAG]
         try:
             return self._parameters[self.KEY_ATTACHED_TAGS]
-        except:
+        except Exception:
             return []
 
     def set_attached_tags(self, tags):
@@ -507,7 +509,7 @@ class GenericBackend():
         """
         try:
             return self.get_static_parameters()[param_name][self.PARAM_TYPE]
-        except:
+        except Exception:
             return None
 
     def register_datastore(self, datastore):
@@ -557,7 +559,7 @@ class GenericBackend():
 
         # saving
         with open(path, 'wb') as file:
-                pickle.dump(data, file)
+            pickle.dump(data, file)
 
     def _load_pickled_file(self, path, default_value=None):
         """
@@ -692,11 +694,11 @@ class GenericBackend():
             self.please_quit = True
             try:
                 self.to_set_timer.cancel()
-            except:
+            except Exception:
                 pass
             try:
                 self.to_set_timer.join()
-            except:
+            except Exception:
                 pass
         self.launch_setting_thread(bypass_quit_request=True)
         self.save_state()
--- a/GTG/backends/generic_backend.py
+++ b/GTG/backends/generic_backend.py
@@ -278,10 +278,10 @@ class GenericBackend():
             parameters[self.KEY_DEFAULT_BACKEND] = True
 
         # default backends should get all the tasks
-        attach_tags = self.KEY_ATTACHED_TAGS not in parameters and \
+        no_attached_tags = self.KEY_ATTACHED_TAGS not in parameters and \
             self._general_description[self.BACKEND_TYPE] == self.TYPE_READWRITE
 
-        if parameters[self.KEY_DEFAULT_BACKEND] or attach_tags:
+        if parameters[self.KEY_DEFAULT_BACKEND] or no_attached_tags:
             parameters[self.KEY_ATTACHED_TAGS] = [ALLTASKS_TAG]
 
         self._parameters = parameters
--- a/GTG/core/task.py
+++ b/GTG/core/task.py
@@ -164,6 +164,7 @@ class Task(TreeNode):
         newtask.sync()
         return newtask.tid
 
+
     def set_title(self, title):
         """Set the tasks title. Returns True if title was changed."""
 
@@ -386,6 +387,7 @@ class Task(TreeNode):
         else:
             self.set_recurring(False)
 
+
     def get_next_occurrence(self):
         """Calcutate the next occurrence of a recurring task
 
@@ -427,6 +429,7 @@ class Task(TreeNode):
                 if p.is_loaded() and p.get_status() in (self.STA_ACTIVE) and p.get_recurring():
                     return True
         return False
+
 
     # ABOUT DUE DATES
     #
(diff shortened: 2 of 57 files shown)


GTG/plugins/unmaintained/bugzilla/services.py
```diff
@@ -8,6 +8,17 @@
...
+# them at once is better.
+services = {
+    'bugzilla.gnome.org': GnomeBugzilla,
+    'bugs.freedesktop.org': FreedesktopBugzilla,
+    'bugzilla.mozilla.org': MozillaBugzilla,
+    'bugzilla.samba.org': SambaBugzilla,
+    'bugs.gentoo.org': GentooBugzilla,
+    'bugzilla.redhat.com': RedHatBugzilla,
+}
+
```
Neui, line 21: While the plugin is not maintained, this shouldn't work because the referenced classes are declared below, so python wouldn't know about it at that point and fail.

GTG/backends/generic_backend.py
```diff
@@ -276,12 +276,14 @@ def __init__(self, parameters):
...
         # default backends should get all the tasks
-        if parameters[self.KEY_DEFAULT_BACKEND] or \
-                (self.KEY_ATTACHED_TAGS not in parameters and
-                 self._general_description[self.BACKEND_TYPE] ==
-                    self.TYPE_READWRITE):
+        attach_tags = self.KEY_ATTACHED_TAGS not in parameters and \
+            self._general_description[self.BACKEND_TYPE] == self.TYPE_READWRITE
+
+        if parameters[self.KEY_DEFAULT_BACKEND] or attach_tags:
             parameters[self.KEY_ATTACHED_TAGS] = [ALLTASKS_TAG]
```
Neui, line 285: I see that you've separated the conditions, but `attach_tags` also contains whenever the backend type is read and writable. Also, it says "not in parameters", so `no_attached_tags`?

GTG/core/task.py
```diff
@@ -214,10 +213,8 @@ def set_status(self, status, donedate=None, propagation=False, init=False):
                     if self.has_parent():
                         for p_tid in self.get_parents():
                             par = self.req.get_task(p_tid)
-                            if (par.is_loaded() and par.get_status() in
-                                (self.STA_ACTIVE)):
+                            if (par.is_loaded() and par.get_status() in (self.STA_ACTIVE)):
```
Neui, line 216: Sometimes you removed the parenthesis in `if (...)`, but not here. Not sure if PEP8 likes this, but maybe we should be consistent and remove the `(` and `)` here.

GTG/core/task.py
```diff
@@ -348,8 +346,7 @@ def is_valid_term():
         if self.has_child():
             for c_tid in self.get_children():
                 child = self.req.get_task(c_tid)
-                if (child.is_loaded() and child.get_status() in
-                    (self.STA_ACTIVE)):
+                if (child.is_loaded() and child.get_status() in (self.STA_ACTIVE)):
```
Neui, line 349: Same `if (...)` parenthesis thing like I've written before.

GTG/core/task.py
```diff
@@ -412,27 +409,25 @@ def get_next_occurrence(self):
...
+            except Exception:
                 raise ValueError(f'Invalid recurring term {self.recurring_term}')
 
     def is_parent_recurring(self):
         if self.has_parent():
             for p_tid in self.get_parents():
                 p = self.req.get_task(p_tid)
-                if (p.is_loaded() and p.get_status() in
-                    (self.STA_ACTIVE) and p.get_recurring()):
+                if (p.is_loaded() and p.get_status() in (self.STA_ACTIVE) and p.get_recurring()):
```
Neui, line 427: Same `if (...)` parenthesis thing like I've written before.

GTG/gtk/backends/backendstree.py
```diff
@@ -190,8 +190,7 @@ def cell_edited_callback(self, text_cell, path, new_text):
         @param new_text: the new name of the backend
         """
         # we strip everything not permitted in backend names
-        new_text = ''.join(c for c in new_text if (c.isalnum() or
-                                                   c in [" ", "-", "_"]))
+        new_text = ''.join(c for c in new_text if (c.isalnum() or c in [" ", "-", "_"]))
```
Neui, line 193: Same `if (...)` parenthesis thing like I've written before, but I am not sure if it would cause some ambiguity at syntax level

GTG/gtk/browser/cell_renderer_tags.py
```diff
@@ -16,17 +16,20 @@
...
 
 
 class CellRendererTags(Gtk.CellRenderer):
 
+    # XXX: these 2 lines were moved within the class to keep the PEP8 checker
+    #      from reporting an E402 error due to require_version needing to be
+    #      called before the PangoCairo import
+    gi.require_version('PangoCairo', '1.0')
+    from gi.repository import PangoCairo
+
```
Neui, line 32: Couldn't we use something like `# nopep8` to bypass the error?

Neui (commented):
Looks OK, haven't tested yet (and want to, later) but seems it shouldn't change anything.

I am mostly worried about current PRs having an conflict after that, but I think they should be simple to fix.

GTG/gtk/browser/cell_renderer_tags.py
odoood: Ah thank you - I was looking for that feature in the pycodestyle docs but didn't see it

GTG/core/task.py
odoood: Agreed, will remove them

GTG/backends/generic_backend.py
odoood: Makes sense, will update it

GTG/plugins/unmaintained/bugzilla/services.py
odoood: Was not paying attention...will put this back

GTG/gtk/backends/backendstree.py
odoood: It's small, I don't think it would cause confusion to remove them, but let me know if you want to leave them

GTG/gtk/browser/cell_renderer_tags.py
Neui: I've seen it in some PR while looking. I looked at the [pycodestyle documentation](https://pycodestyle.pycqa.org/en/latest/intro.html#related-tools), and it suggests `# noqa` instead.

odoood: Ah...I thought `# noqa` was reserved for the error codes marked with the `(^)`, but it works for this - updating it

andersonjwan:
I previously used Pylint and PyCodeStyle to format the `cell_renderer_tags.py` file in #689, but after looking at your method with `make pep8`, do we have a formatting standard method that I should apply across all files?

i.e. Do we have a set of linters, formatters, and type checkers that this project formally accepts that can be consistently used across the files?

I'm familiar with mypy, pylint, and black; so, I'm wondering if maybe we can standardize some of this? I'd be happy to update documentation if we can agree on the set of tools to use.

Neui:
I guess since we have that Makefile (that I didn't noticed we had) I'd say we should use that, since it is already stored in the repository and thus may by used by others. It uses pycodestyle and pyflakes. AFAIK mypy uses the python type annotations we currently use rather sparingly, and AFAIK pygobject functions aren't covered in those type annotations, so i wouldn't make it mandatory, but I support the idea of using it. Maybe we should move this to #237 or its own issue.

odoood:
The Makefile is referenced in [docs/contributors/coding best practices.md](https://github.com/getting-things-gnome/gtg/blob/1c1496bba6fc598345a09ef0f187b5e6cc2c4d7f/docs/contributors/coding%20best%20practices.md). Made the changes in this PR based on the instructions there for the test suite & coding style checks.

That doc also needs a couple of tweaks (specifically for the `make` commands listed) - was going to address those separately.

diegogangl:
Hi, thanks for the work! Sorry for taking forever. Extremely time-constrained these days.
This LGTM , the only issue for me is removing blank lines. I can't stand PEP8's one-blank-line rule for methods, in fact I manually added many of those lines :p

I should have added a line to pylintrc to ignore those errors ages ago. Could you revert those changes? If you don't have the time I can do it in a couple months, if I'm a bit more free.

odoood:
Glad to be able to contribute--no need to apologize, we're all doing this in the time that we can spare. That is actually a relief, I don't like that rule about the single blank lines either lol. I'll revert those changes and update the pylint.rc to ignore those errors too.

odoood:
Setting the pylintrc to disable them had no effect on the pycodestyle output, so I set it to be ignored in the Makefile target command

diegogangl:
Merged, thanks!

Issue #629, opened 2021-04-13
Version 0.5 fails to parse tasks file (and refuses to launch) when a saved search's name and query matches exactly one of the tags


iaeiou:
Hello there,

I was using GTG 0.4 with Flatpak and everything was going smoothly.

Today the new version (0.5) is running after an automatic update. Unfortunately, only the test tasks are showing now:

![gtg2](https://user-images.githubusercontent.com/69427615/114487761-71590000-9c10-11eb-963e-1e90975a355c.png)

I think there's no data loss though, here is the treeview of  `~/.var/app/org.gnome.GTG`

```
.
├── cache
│   ├── fontconfig
│   │   └── 8e575bbd690cbdd5c13d70f46836611e-le64.cache-7
│   ├── mesa_shader_cache
│   │   ├── 7b
│   │   │   └── 129dd520072e893669cbd79084012ead13f847
│   │   ├── 8c
│   │   │   └── 96c84181837fef71d69bec0d80f4461654e67d
│   │   ├── 93
│   │   │   └── b39334786cf727f6b3c2fccda51462ab461af6
│   │   ├── 9d
│   │   │   └── 24b54391ecbab9c3cdda96131158b038beae83
│   │   ├── a9
│   │   │   └── aa11003a3cf6237ae869d3a085ddb9664cab72
│   │   ├── ab
│   │   │   └── f7769e64fc012c17c312fce2667f6bc479cf6b
│   │   ├── dc
│   │   │   └── a844c1628e39756e9c218b18b4538519f7a7dd
│   │   ├── de
│   │   │   └── 43b818d9980a0a4bdcc07e913b3bc467359b0c
│   │   └── index
│   ├── tmp
│   └── yelp
│       └── WebKitCache
│           └── Version 16
│               ├── Blobs
│               └── salt
├── config
│   ├── glib-2.0
│   │   └── settings
│   │       └── keyfile
│   ├── gtg
│   │   ├── backends.conf
│   │   ├── gtg.conf
│   │   ├── plugins
│   │   │   └── Urgency Color
│   │   │       └── preferences
│   │   └── tasks.conf
│   └── yelp
└── data
    ├── gtg
    │   ├── backup
    │   │   ├── gtg_data.xml.2021-04-13.bak
    │   │   ├── gtg_data.xml.bak.0
    │   │   ├── gtg_data.xml.bak.1
[...]

PR #703, opened 2021-09-05
Workaround: Separate saved searches with tags name

Neui:
Since unfortunately saved searches and tags share the same namespace, it
makes it difficult to create a saved search with the exact name as an
existing tag.

As a current workaround until we separate them, we just prefix the saved
searches with `SEARCH_TAG_PREFIX`, which an user is unlikely to use.
From what I can see, it still displays properly in the UI after some
fixes, aside from existing bugs, like the tag editor stripping spaces, `@blog !or @efficiency` having the first `@` stripped out in the sidebar, and so on.

In the XML file this prefix isn't added, because it would be redundant
and make it easier to deal with in future versions.

See for example #629. Does not fix tag inconsistencies (renaming, special symbols etc.).

### DIFF
--- a/GTG/backends/backend_localfile.py
+++ b/GTG/backends/backend_localfile.py
@@ -34,6 +34,7 @@ from gettext import gettext as _
 from GTG.core import xml
 from GTG.core import firstrun_tasks
 from GTG.core import versioning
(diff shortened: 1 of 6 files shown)


diegogangl:
Always impressed by how thorough your are with these PRs, thanks!

Issue #716, opened 2021-10-27
TypeError: Couldn't find foreign struct converter for 'cairo.Context'


mizzunet:
on ./launch.sh compilation, the programs opens but shows this error,
`TypeError: Couldn't find foreign struct converter for 'cairo.Context'`
I have installed pycairo  
```
pip3 install pycairo
Requirement already satisfied: pycairo in /usr/lib/python3/dist-packages (1.16.2)
```
Same for python3-cairo  apt package

Also, how do I install GTG to /usr after this compilation?

Debian bookworm

Thanks

Edit:

Getting Things GNOME! v0.5-115-g3eacad13
CPython 3.9.7 (default, Sep 24 2021, 09:43:00)   [GCC 10.3.0]
GTK 3.24.30, GLib 2.70.0
PyGLib 3.42.0, PyGObject 3.42.0
Linux-5.14.14-xanmod2-x86_64-with-glibc2.32

PR #717, opened 2021-10-27
README: Install python3-gi-cairo for debian

Neui:
This is required for getting GTG to work, otherwhise the following error
would occur:

TypeError: Couldn't find foreign struct converter for 'cairo.Context'

It seems that package is installed by gedit, which is a default
applictaion for GNOME, so during package testing it might be easy to
miss.

Since python3-gi-cairo depends on python3-cairo, I "replaced" that
dependency, so the line isn't getting too long.

Found out by #716
I actually haven't tested this, but according to the reporter it should work.
Also, I haven't looked at Fedora yet, and I am unsure how to call it in the textual dependency list (where we mention pycairo).


diegogangl:
Can't test it either, so we'll go by the user's report. Thanks!

PR #719, opened 2021-10-30
Fix tasks losing status at restart

### LINKED ISSUE #718: Not saving tasks 'Marked as Done' after closing
I installed gtg by
```
meson -Dprefix=/usr/local/ systembuild
ninja -C systembuild install
```
Once any task is marked as done, the task will be moved to the 'Closed' tab as normal.
But as gtg closed and reopened, the tasks marked as Done will come back under the 'Open' tab. This seems like the data does not get retained.

But creating, editing, modifying tasks are retained after reopening though.

Moreover, launching gtg from repo root dir by `.launch.sh` works well.
How do I get logs?

Info:
Debian bookworm distro
GTG version v0.5-115-g3eacad13

Neui:
The recent date thing #597 also changed the file loading code to make the
date reading code more compact, but also accidentally also removed
reading the tasks status, so all tasks were considered Open after a
restart.

Fixes #718

Also, it seems to only rewrite tasks in the XML tree when the task has been changed? Seems this is the behaviour, so unless something else was changed in the tasks other than just closing them, they should automatically be closed when this PR is applied (since the state is still in the XML file).

### DIFF
--- a/GTG/core/xml.py
+++ b/GTG/core/xml.py
@@ -37,6 +37,7 @@ def task_from_element(task, element: etree.Element):
 
     task.set_title(element.find('title').text)
     task.set_uuid(element.get('id'))
+    task.set_status(element.attrib['status'], init=True) # Done date set later
 
     # Retrieving all dates
     dates = element.find('dates')


diegogangl:
Wish I had more to review things thoroughly >_>
Thanks! Merged

PR #722, opened 2021-11-04
Add gtg://TASK-ID to the command line help

### LINKED ISSUE #721: %U parameter in desktop file
The current desktop file contains
`Exec=gtg %U`
The %U means it can accept list of urls. But according to application usage message (gtg --help), it cannot accept them. 
Need to remove %U from desktop file.

Neui:
Fixes #721

Looks like this now:
```
$ LC_ALL=C ./launch.sh -- --help
[cut build system output]
Usage:
  gtg [OPTION?] [gtg://TASK-ID?]

Help Options:
  -h, --help                 Show help options
  --help-all                 Show all help options
  --help-gapplication        Show GApplication options
  --help-gtk                 Show GTK+ Options

Application Options:
  -v, --version              Show program version
  -d, --debug                Enable debug output
  -t, --title=TITLE          Use special title for windows' title
  --display=DISPLAY          X display to use

```

For some reason, the special triple dots "…" displays as a question mark when using `LC_ALL=C`, but display correctly in my language (German UTF-8). Using `LC_ALL=C.UTF-8` just displays the German text for me.

### DIFF
--- a/GTG/gtk/application.py
+++ b/GTG/gtk/application.py
@@ -92,6 +92,7 @@ class Application(Gtk.Application):
 
         super().__init__(application_id=app_id,
                          flags=Gio.ApplicationFlags.HANDLES_OPEN)
+        self.set_option_context_parameter_string("[gtg://TASK-ID…]")
 
     # --------------------------------------------------------------------------
     # INIT


Ashark:
> For some reason, the special triple dots "…" displays as a question mark when using LC_ALL=C

The tripple dot symbol is from unicode, but it is not supported by C locale. I have recently also been [caught](https://bugs.kde.org/show_bug.cgi?id=444579#c1) by LC_ALL=C behavior in another project. Quoting the info here:
> If you run `LC_ALL=C locale charmap` I'm sure you'll find that C is in fact not capable of handling unicode because it defaults to ANSI. You need to use a .UTF-8 capable locale.

Neui:
> > For some reason, the special triple dots "…" displays as a question mark when using LC_ALL=C
> 
> The tripple dot symbol is from unicode, but it is not supported by C locale. I have recently also been [caught](https://bugs.kde.org/show_bug.cgi?id=444579#c1) by LC_ALL=C behavior in another project. Quoting the info here:
> 
> > If you run `LC_ALL=C locale charmap` I'm sure you'll find that C is in fact not capable of handling unicode because it defaults to ANSI. You need to use a .UTF-8 capable locale.

I did actually knew that, but I'd expected `LC_ALL=C.UTF-8` to work which didn't. Also I didn't manage to get `en_US.UTF-8` and a couple other working either to get the Unicode symbols. But that isn't something GTG should try to "fix" since it isn't in our problem.

diegogangl:
Merged, thanks!

Issue #533, opened 2020-12-30
Test GTG and liblarch with GTK4


nekohayo:
Maybe it will all work seamlessly with zero changes required, or maybe it will break something, I have no idea. I haven't closely followed what GTK4 is about. So if someone outside is seeing this and would like to help by testing, please report your findings here; even better if you can provide fixes if any are needed!

nekohayo:
Well, it is pretty much proven to work in `master` now, nothing actionable to do here, closing!

PR #723, opened 2021-11-05
Port to GTK4

ranchester2:
Port GTG to GTK4 while keeping things the same. Highly recommended to not look at the entire diff but go through the commits individually.

Mentioning issue: https://github.com/getting-things-gnome/gtg/issues/533

Build with liblarch from here: https://github.com/ranchester2/liblarch/tree/gtk4
Liblarch changes are very minimal, however drag and drop will need to be changed a bit when it starts working in GTK

Mostly works, existing issues:

- [x] Figure out how to recolor symbolic icons for dark mode in cell_renderer_tags NOTE: fix adds dependency on master sdk, which isn't available on Flathub, an alternative would be to bundle the latest GTK instead
- [x] ~~cell_renderer_tags is slightly blurry~~ this is also an issue in master, not a gtk4 issue
- [x] Harder to open right click menu for multi selection, you now have to shift+select and shift+right click instead of just right click
- [x] .menu-disclose
- [x] tags sidebar slightly tight spacing
- [x] plugin about dialog markup
- [ ] Figure out treeview drag and drop (EDIT: Treeview DND is completely broken in GTK4, GTK4 bug prevents this).

I think this is quite ready currently, however there is the treeview dnd issue that can't really be fixed at the moment.
And this now uses the master SDK as GTK 4.5 is needed, this prevents releasing on Flathub with this until GNOME 42 is released (or bundle GTK)

Plugins:
- [x]  Export and print
- [x] Send tasks via email
- [x] Untouched tasks
- [x] Urgency color
- [x] Hamster
- [x] Dev console
- [x] Gamify

Window positions aren't restored, GTK4 removed this feature as the api is now based on Wayland, not X11.

I added a "Today" button to the startdate calendar as you can no longer receive events about selection of the currently selected date in GtkCalendar (nor can you not have any date selected)

gtg-gtk4-screenshot


Neui:
Nice! I looked only at some commits for now, but they seemed fine. The commit messages are the cherry on top.
I'd probably not merge this for the next release because this will take a while to review and test.
Also, I wonder if some early stuff can be applied to the current GTK3 version now so some of the changes can already be used and smaller chunks are a bit easier to review. Like making the TagEditor an GtkDialog now (I don't remember why I didn't do that...)

ranchester2:
> Nice! I looked only at some commits for now, but they seemed fine. The commit messages are the cherry on top. I'd probably not merge this for the next release because this will take a while to review and test. (And this is still a draft PR for I just noticed) Also, I wonder if some early stuff can be applied to the current GTK3 version now so some of the changes can already be used and smaller chunks are a bit easier to review. Like making the TagEditor an GtkDialog now (I don't remember why I didn't do that...)

All commits before the update imports commit were for "preparing", and are for GTK3. (However the gmenu commit didn't port the hamster plugin). However I missed quite a lot of stuff there, as PyGObject seems to be really bad at printing deprecation warnings, so I fixed those things later. Later invisible changes could be backported probably, however probably not as-is

ranchester2:
The main remaining issue with this is currently treeview drag and drop.

And that can't be really fixed as GTK broke it when inventing the new drag and drop api and it currently just doesn't work:

https://gitlab.gnome.org/GNOME/gtk/-/issues/3649
https://gitlab.gnome.org/GNOME/gtk/-/issues/4206
https://gitlab.gnome.org/GNOME/gtk/-/issues/3656
https://gitlab.gnome.org/GNOME/gtk/-/issues/3642

Maybe there should be an alternative way of adding parents for tags in the sidebar or reordering in the tree? However that will definitely be worse to use.

Neui:
On a related thing, it seems [PyGObject also needs some fixes](https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/175) for gtk4 TreeView. Also I don't think you can "reoder" stuff anyway currently (minus parenting and un-parenting), but looking at your issue pages DnD looks like to be broken completely?  We could hope that when we finally merge this PR that it'll be fixed by then...

Diegogangl started a branch https://github.com/getting-things-gnome/gtg/tree/ui_listboxes to migrate away from the TreeView widget, but doesn't seem to be complete however (and would need testing etc.)

(I am also wondering how keyboard-only users do parenting and unparenting?)

ranchester2:
A possible solution since there seem to be plans to drop treeview anyway, would maybe be to never actually fix treeview drag and drop, and instead not merge this until a port to the new list widgets is done? And then reimplement drag and drop with the non-treeview widgets that aren't broken. However that will definitely require very major changes.

diegogangl:
Hey @ranchester2, thanks for working on this! It really takes a load off our backs :)
Shame treeviews are being a problem again. Like @Neui said we don't support manual re-ordering, so DnD is only for parenting. Also I'm impressed we're at 4.2 and DnD is this broken still.

The plan is to eventually move to the new Gtk4 listviews based on the new core (#668). I started a branch to port to Gtk3 listboxes because I was getting impatient, but had to stop to fix some bugs and missing stuff in the new core instead.
Also, I was wondering if neui's branch would be a better segway into Gtk4.

If DnD is the only limitation, we could merge this after we plug the new core during the 0.7 cycle. We'll have to accept a temporary regression in master until we complete the switch to listviews.

diegogangl:
Hey @ranchester2  could you also open a PR for your liblarch changes when you have some time? That way we can land those as soon as we're ready to land this one too

diegogangl:
Hey @ranchester2 , think you could open a PR for your liblarch port? Once we get the new core branches in, it would be nice to work on getting this landed. Then we can move on to newcore + list views

diegogangl:
Closing for now since this had +infinity conflicts, is based on an older version of Gtk and is superceded by the megaport branch

nekohayo:
For those wishing to help, the successor to this branch is #894

PR #725, opened 2021-11-08
Add search syntax popup

diegogangl:
Quick idea I had, I can never quite remember this. The button in the search field toggles the popup with a quick cheatsheet of the syntax. 

Screenshot from 2021-11-08 19-42-15

We could do the same with the quick add entry. Replacing the emojis button with one of this.


GTG/gtk/browser/main_window.py
```diff
@@ -260,6 +261,17 @@ def _init_widget_aliases(self):
 
         self.tagpopup = TagContextMenu(self.req, self.app)
 
+        # Setup search syntax popup button
+        self.search_entry.set_icon_from_icon_name(Gtk.EntryIconPosition.SECONDARY, 'help-faq-symbolic')
```
Neui, line 265: Couldn't we do this directly in the Builder XML?

GTG/gtk/browser/main_window.py
```diff
@@ -260,6 +261,17 @@ def _init_widget_aliases(self):
...
+        self.search_entry.connect('icon-release', self.on_icon_click)
+
+
+    def on_icon_click(self, entry, icon_pos, event):
```
Neui, line 269: I'd suggest `on_search_box_icon_click` to be less ambiguous.
You could also add typehints here.

GTG/gtk/data/main_window.ui
```diff
@@ -1,59 +1,61 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.36.0 -->
+<!-- Generated with glade 3.38.2 -->
```
Neui, line 2: Possibly "remove" glade introduced automatic changes so reviewing is a bit easier (to see what has been changed).

Neui (commented):
Idea is OK for me, seems to work but I have some things to complain about.
Also, It seems a bit inconsistent for me with `word` but later `[date]` is being used later.
And what the exact difference between `word` and `"text"` isn't very clear (maybe add spaced in the example?). Also, what about `word word`?

GTG/gtk/browser/main_window.py
diegogangl: huh, didn't check but it turns out it can be done

diegogangl:
> Also, It seems a bit inconsistent for me with word but later [date] is being used later.
> And what the exact difference between word and "text" isn't very clear (maybe add spaces in the example?). Also, what about word word?

I copy/pasted from the table in help, but I agree we could make it clearer

diegogangl:
Updated

Screenshot from 2021-11-09 18-41-00

GTG/gtk/data/main_window.ui
diegogangl: Removed a bit, but it's hard to tell what can be removed from the auto changes without breaking things

Neui:
Looks better now, but what about `[word] [word]`? I think that is useful to know

nekohayo:
My first reaction was, "Don't we have GtkTooltips for this?" but I understand that those are probably too limited in styling (especially text wrapping etc.) and that the popover approach is preferrable. Just gotta make sure it doesn't interfere with focus or keyboard use in general or something like that (I don't have specific in mind here, I don't know why I'm saying this vague theoretical concern, but I'm saying it just in case there is something we haven't thought of). I think the UI you are proposing looks pretty neat.

Is it supposed to support the `tags:foo,bar,baz` syntax too like the quickadd entry, or is that specific to the quickadd entry?

As for the quickadd entry, is there much more to document than what's already in the tooltip? If not, then it could probably remain simply as a one-line tooltip like it is now...

nekohayo:
Another theoretical concern (but probably not an actual issue): are the translated versions of the soon, now, etc. keywords accepted as input in the app in practice? I think so, but I don't remember if it worked 100%.

In any case, at first glance and very broadly looking (and without testing), it looks good. What could possibly go wrong? :)

Neui:
At the moment this is just for the search box and you manually need to click on the little icon in the search entry (not sure if it is keyboard friendly). And to my knowledge, the search box and quick add entries accept translated keywords.

nekohayo:
Oh right this is for the search box at the top, which doesn't even have a tooltip right now,  
not the quick-add entry at the bottom!

Makes a lot of sense, now that I'm looking at this again, given how many search operators there are.

@diegogangl would need to rebase this to work with `master` though, so marking this as a draft until then.

Issue #727, opened 2021-11-09
Blurred tags_cell_renderer colors on lodpi


ranchester2:
c6974bf03470132c7daf52e78fab18d2d276c603
I think this looks just worse, I agree it is better on HIDPI though. However I definetely think having a sharp border is better when you don't have lots of pixels.

![now-lodpi](https://user-images.githubusercontent.com/77527554/140898525-439c1062-d6af-43f6-827e-b62bfb38f6e4.png)
![before-lodpi](https://user-images.githubusercontent.com/77527554/140898528-4ed93196-b502-4c7f-99a4-01a6fcab921e.png)

https://github.com/getting-things-gnome/gtg/pull/728

PR #728, opened 2021-11-09
cell_renderer_tags: don't antialias on lodpi

ranchester2:
This blurs the pretty border and imo makes it look very smeared (and the shape looks weird too).

This definetely is better than having a low res rounded rectangle in
HIDPI though, so there it is still anti-aliased.

now-lodpi
before-lodpi

See https://github.com/getting-things-gnome/gtg/issues/727  (I first typod the issue number, this is why I mentioned the wrong issue)

### DIFF
--- a/GTG/gtk/browser/cell_renderer_tags.py
+++ b/GTG/gtk/browser/cell_renderer_tags.py
@@ -140,6 +140,10 @@ class CellRendererTags(Gtk.CellRenderer):
 
         # Drawing context
         gdkcontext = cr
+        scale_factor = widget.get_scale_factor()
+        # Don't blur border on lodpi
+        if scale_factor == 1:
+            gdkcontext.set_antialias(cairo.ANTIALIAS_NONE)
 
         # Coordinates of the origin point
         x_align = self.get_property("xalign")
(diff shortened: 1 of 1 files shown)


diegogangl:
Barely noticed it TBH. Change LGTM though, the cell renders will go away when we switch to listviews. What do you think @Neui ?

Neui:
I think it looks better. I had a worry about the icons being "changed", but by a screenshot comparison by myself they are identical (at least the bottom one).

However it being "blurry" means the box isn't pixel aligned, but looking at the code it looks fine, no additional fraction introduces... I am OK with the current solution, but I think this should at least be looked into further later why cairo wants to smooth it out that way.

Code LGTM.

> the cell renders will go away when we switch to listviews.

But this is about drawing the colour icons icons, which likely is something we can re-use parts of it for your new set of widgets.

diegogangl:
Merged, thanks!

> But this is about drawing the colour icons icons, which likely is something we can re-use parts of it for your new set of widgets.

There's another way, using a button (stole that from Planner :P). It's not super pretty either, but avoids cairo
https://github.com/getting-things-gnome/gtg/commit/7d10bcae26c31e240dc74b59fb332aad62861dbc#diff-16410d7ac29c0084ac86e47b216c003a553ccbab0cd9e7ad83b6b1d40cc957aeR236

PR #729, opened 2021-11-15
Revert meson plugin translation apply thing

### LINKED ISSUE #720: Meson 0.60.0 build error: i18n.merge_file got unknown keyword arguments "build_always_stale"
Building the `gtg-git` AUR package from 828c345d:

```
Build started at 2021-11-01T19:44:03.185274
Main binary: /usr/bin/python
Build Options: -Db_lto=true -Db_pie=true -Dprefix=/usr -Dlibexecdir=lib -Dsbindir=bin -Dauto_features=enabled -Dbuildtype=plain -Dwrap_mode=nodownload
Python system: Linux
The Meson build system
Version: 0.60.0
Source dir: /build/gtg-git/src/gtg
Build dir: /build/gtg-git/src/build
Build type: native build
Project name: gtg
Project version: 0.5.0
Build machine cpu family: x86_64
Build machine cpu: x86_64
Host machine cpu family: x86_64
Host machine cpu: x86_64
Target machine cpu family: x86_64
Target machine cpu: x86_64
Program python3 found: YES (/usr/bin/python3)
Configuring prefix-gtg.sh using configuration
Configuring gtg using configuration
Configuring local-gtg using configuration
Program git found: YES (/usr/bin/git)
Running command: /usr/bin/git describe --dirty
--- stdout ---
v0.5-117-g828c345d

--- stderr ---


Configuring info.py using configuration

gtg/GTG/plugins/meson.build:19:7: ERROR: i18n.merge_file got unknown keyword arguments "build_always_stale"
```

Neui:
Since meson 0.60, the `build_always_stale` option was removed:
https://github.com/mesonbuild/meson/commit/6b1a80024c79876dc3b75f2b525be24ca8fe671a

However, meson doesn't make such files depend on the translation .po
files, which is a known bug:
https://github.com/mesonbuild/meson/issues/2621

So with that option taken away, I tried passing a list to the input
which includes all .po files, since I saw it only uses the first element
for the translations, but that fails when trying to generate po files.

Since it seems you can't dynamically add dependencies after the fact,
this seems like a dead end, so I am reverting this for now to make it
compile on meson 0.60.

Reverts #555 to fix #720

### DIFF
--- a/GTG/plugins/meson.build
+++ b/GTG/plugins/meson.build
@@ -24,9 +24,7 @@ foreach plugin : gtg_plugins
     install_dir: plugin_install_dir,
     type: 'desktop',
     args: ['--keyword=name', '--keyword=short-description', '--keyword=description'],
-    build_by_default: true,
-    build_always_stale: true
-    # build always because otherwise new translations won't be applied
+    build_by_default: true
   )
   subdir(plugin.underscorify())
 endforeach
--- a/docs/contributors/translating.md
+++ b/docs/contributors/translating.md
@@ -75,5 +75,8 @@ Make sure to this AFTER COMMITING YOUR CHANGES!
 * During launching, it might complain about certain files not being found in [`po/POTFILES.in`][POTFILES.IN].
   It is safe to remove the lines from that file and re-run until it works.
   It would be useful to comment about that if you're submitting your translation, just in case.
+* Plugin related strings don't update after updating the translation.
+  The cause is unknown, but you can delete the plugin files to re-generate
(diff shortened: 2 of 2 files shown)


diegogangl:
Could we run `rm -f .local_build/GTG/plugins/*.gtg-plugin` as a step before building?

Neui:
> Could we run `rm -f .local_build/GTG/plugins/*.gtg-plugin` as a step before building?

You mean in the `scripts/debug.sh` script? We could do that.

diegogangl:
Merged, we can take a look at the script afterwards. Thanks!

PR #730, opened 2021-11-18
Use GtkSource.view for taskview and enable undo

### LINKED ISSUE #296: Implement undo & redo in the "task editor" window
In most (all?) text editors hitting ctrl+z undo the last action, like word or character typed/removed, etc...

This has never worked in GTG, old or current master version. So for example removing a task's content by mistake needs to be re-typed.

diegogangl:
This PR makes taskview sublcass GtkSource.view instead of the regular textview.
The immediate change is that now we can have undo/redo. It also opens the door for more features, like tag autocompletion.
It's a small enough change that can go into 0.6

The only limitation for now is that you can't undo a subtask once it's created. I don't think this is particularly bad though, since ctrl+z'ing a subtask would delete it without confirmation.

closes #296

### DIFF
--- a/GTG/gtk/editor/taskview.py
+++ b/GTG/gtk/editor/taskview.py
@@ -24,7 +24,7 @@ from time import time
 from uuid import uuid4
 import logging
 
-from gi.repository import Gtk, GLib, Gdk, GObject
+from gi.repository import Gtk, GLib, Gdk, GObject, GtkSource
 
 from GTG.core.requester import Requester
 import GTG.core.urlregex as url_regex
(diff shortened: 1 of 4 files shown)


Neui:
Don't forget to modify the README dependency list

README.md
```diff
@@ -50,6 +50,7 @@ git pull --rebase
   - pango
   - gdk-pixbuf
   - GTK 3
+  - GtkSource 4
```
Neui, line 53: The project name is GtkSourceView 4, might better to use that name

GTG/gtk/editor/taskview.py
```diff
@@ -188,6 +188,7 @@ def process(self) -> None:
 
         # Parse the text line by line until the end of the buffer
         while not start.is_end():
+
```
Neui, line 191: Unnecessary empty line.

Neui (approved):
Works.

diegogangl:
Thanks for reviewing man!

Issue #377, opened 2020-06-21
Unified "Reopen" action for closed tasks in the treeview


nekohayo:
Somewhat similar in spirit to #298, but not exactly.

I noticed at least one case where the translatable string "context" parameter can make a difference for translators, such as my own FR translation... in GTG/gtk/data/context_menus.ui, the closed_task_menu items have that problem where I was translating them identically in the FR version I had considered both "Mark as Not Done" and "Undismiss" as the same thing in practice, "Restaurer" (which means "Restore"), because French is verbose and I preferred to have a shorter label there.

But then when you look at the right-click menu for Closed tasks in the main window, you end up with twice the same string due to my overzealous translation simplification:

![restaurerX2](https://user-images.githubusercontent.com/479401/85230780-77030200-b3c0-11ea-8490-73f6a08733e3.png)

One way to address this could be to have  `context="right-click menu in the treeview"` as a parameter everywhere in GTG/gtk/data/context_menus.ui... but this made me think of something else.

In addition to the potential UI improvements for #298, I think that in the main window / closed tasks treeview, when you right-click closed tasks, instead of being presented with two choices, GTG should just offer you to "Reopen" the tasks, and behind the scenes it figures out for each task if it needed to be "marked as not done" or "undismissed".

Hence this bug report, because that means there'd some code logic involved instead of just flagging strings now.

Neui:
I think combining those is much more intuitive, because I was also confused at first before reading this, and I guess the distinction when reopening isn't really important.

diegogangl:
Tasks have three states:

```python
    STA_ACTIVE = "Active"
    STA_DISMISSED = "Dismiss"
    STA_DONE = "Done"
```
So in both cases they would just return to "active". There's not much point in having both options in the menu TBH.

nekohayo:
Confirming, as I'm running the English version lately: even with the distinction, it makes no sense to me to have both "Mark as Not Done" and "Undismiss" as context menu actions in the task browser, so in the case of the Closed tasks view, I'd have just a unified "Reopen" action.

PR #731, opened 2021-11-19
Unified 'Reopen' action (#377)

Sir-Thomas:
This change replaces the current "Mark as Not Done" and "Undismiss" options when right clicking a closed task with a single "Reopen" option as suggested in #377.

This is my first time contributing to an open source project, so let me know if there is anything that I need to change.

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -1315,6 +1315,23 @@ class MainWindow(Gtk.ApplicationWindow):
                 task.set_status(Task.STA_DISMISSED)
                 self.close_all_task_editors(uid)
 
+    def on_reopen_task(self, widget=None):
+        tasks_uid = [uid for uid in self.get_selected_tasks()
+                     if uid is not None]
+        tasks = [self.req.get_task(uid) for uid in tasks_uid]
+        tasks_status = [task.get_status() for task in tasks]
+        for uid, task, status in zip(tasks_uid, tasks, tasks_status):
+            if status == Task.STA_DONE:
+                task.set_status(Task.STA_ACTIVE)
+                GObject.idle_add(self.emit, "task-marked-as-not-done", task.get_id())
+                # Parents of that task must be updated - not to be shown
+                # in workview, update children count, etc.
+                for parent_id in task.get_parents():
+                    parent = self.req.get_task(parent_id)
+                    parent.modified()
+            elif status == Task.STA_DISMISSED:
+                task.set_status(Task.STA_ACTIVE)
+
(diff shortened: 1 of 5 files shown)


diegogangl:
Hey welcome and thanks!
Code looks ok, and works great. The only thing to change would be the commit message. Try using a verb (like "Add unified reopen action"). The number should be in the description part of the message (leave a blank line after the first line). Also you can use keywords to automatically close issues in Github, so you could just write  `fix #377` to autoclose that issue once this is merged in master.

More info: https://chris.beams.io/posts/git-commit/

GTG/gtk/browser/main_window.py
```diff
@@ -1315,6 +1315,26 @@ def on_dismiss_task(self, widget=None):
                 task.set_status(Task.STA_DISMISSED)
                 self.close_all_task_editors(uid)
 
+    def on_reopen_task(self, widget=None):
+        tasks_uid = [uid for uid in self.get_selected_tasks()
+                     if uid is not None]
+        if len(tasks_uid) == 0:
+            return
```
Neui, line 1322: Not needed. `tasks_uid` is an emtpy list, so the following `tasks` and `tasks_status` are empty as well, and `zip([], [], [])` is empty too, so the loop isn't entered.
I know you've copied from above, but if you're already here you can change this instance.

GTG/gtk/browser/main_window.py
```diff
@@ -1315,6 +1315,26 @@ def on_dismiss_task(self, widget=None):
...
+        tasks = [self.req.get_task(uid) for uid in tasks_uid]
+        tasks_status = [task.get_status() for task in tasks]
+        for uid, task, status in zip(tasks_uid, tasks, tasks_status):
+            if status == Task.STA_DONE:
+                # Marking as undone
```
Neui, line 1327: This is already clear from the context.  I know you've copied from above, but if you're already here you can change this instance.

Neui (approved):
LGTM and works. Suggesting removing some unnecessary lines. Maybe also put the shortcut in the shortcuts window.

Sir-Thomas:
I have made the changes you guys suggested. Do you want to take another look at it and see if I understood everything correctly?

diegogangl:
LGTM, one last thing for me:

Could you squash commit `40fa626ac6d3015d741fdd4c69aa0726caae44d7` onto `c9737dbfc0a62361b2111878c7a3a3ace2f54237` to make the history cleaner?

diegogangl:
Hi, @Sir-Thomas did you have a chance to could squash the commits? It'd be great to land this for 0.6

diegogangl:
No problem, thanks!

PR #732, opened 2021-11-20
Adaptive tomorrow button to make window smaller

### LINKED ISSUE #673: Gnome window placement and resizing via keys
Hi all

For most gnome applications, Super-Right moves the current window to the right and makes it half screen size. It's a very efficient way of getting two apps to sit size by side. Typically to use one as reference while you work in the other.

I have noticed that GTG does not react to Super-Right (or any of the super direction keys) at all and wondered why that was?

I also notice that the GTG window has a minimum width that would actually prevent it from being half a screen wide on my 1366x768 laptop.

A nice to have.

Neui:
A bug report #673 was that the minimum width of the main window is a bit too
high for some resolutions (when trying to use GTG half of it). A
proposed solution #677 would be to use an icon, but it is rather hard to find
a good system-provided one.

What I've thought is that maybe just display the "Start tomorrow" text
while the window is big enough, and only the icon when it gets smaller.

I couldn't find a good way to do that with only using GTK stuff, so I
whipped my own (and first) widget that displays an widget depending on
the (minimum) size of its children widgets, which then would be the
"Start tomorrow" label, and an Image for the icon.

This widget is rather tailored for GTG in that usecase only, which means
not everything has been implemented yet, like different size request
modes. Maybe this'll be done later.
Another thing to look out is easy GTK4 migration.

This is still WIP since this is primarily in a development stage ("how
do I create an GTK widget?"). For example the widget is manually
inserted in code and not in the Builder XML file, or that I might still
need to implement some interfaces methods to make it stable.
Or that the button doesn't size down to the "!" label widget.

Also, be careful, I've encountered accident exceptions-loops that
caused my window manager (i3wm) to "freeze". However, I could just
switch to an virtual terminal, log in, kill GTG, and restart i3wm from
there (via i3-msg, very nice).

Supposed to supersede #677 to fix #673

https://user-images.githubusercontent.com/1196130/142735164-0f2a9ca7-3b2b-4636-b9f1-84bdf20672da.mp4

### DIFF
--- a/GTG/gtk/browser/adaptive_button.py
+++ b/GTG/gtk/browser/adaptive_button.py
@@ -1,0 +1,288 @@
+import logging
+import dataclasses
+from typing import Optional, List, Tuple
+
+from gi.repository import GObject, GLib, Gtk, Gdk
+
+
+log = logging.getLogger(__name__)
+
+
+class AdaptiveFittingWidget(Gtk.Container):
+    """
+    This widget chooses the biggest but fitting children widgets and displays
+    that one. This is useful to switch out between text and an icon in an
+    button if the window gets smaller.
+    Note that this is heavily "adjusted" what GTG uses this widget for.
+    """
+
+    __gtype_name__ = "GTGAdaptiveFittingWidget"
+
+    @dataclasses.dataclass
+    class ChildItem:
+        """
+        Represents a child widget with some useful to have properties to use
+        later.
+        """
+        widget: Gtk.Widget
+        """Widget to possibly draw and request data from"""
+
+        request_mode: Gtk.SizeRequestMode = Gtk.SizeRequestMode.HEIGHT_FOR_WIDTH
+        """What request mode the widget is in"""
+
+        minimum_width: int = 0
+        """The minimum width the widget claims to have"""
+        natural_width: int = 0
+        """The natural width the widget claims to have"""
+
+        minimum_height: int = 0
+        """The minimum height the widget claims to have"""
+        natural_height: int = 0
+        """The natural height the widget claims to have"""
+
+    def __init__(self, *args, **kwargs):
+        self._children: List[ChildItem] = []
+        """Child widgets to consider together with internal bookkeeping data"""
+
+        self._child_to_draw: Optional[ChildItem] = None
+        """Current child to draw"""
+
+        self._draw_smallest_child_if_smaller: bool = True
+        """Whenever to draw the smallest child even if it doesn't fit"""
+
+        self._spammy_debug: bool = False
+        """Show more debug messages related to this widget"""
+
+        super().__init__(*args, **kwargs)
+
+    @classmethod
+    def new(cls):
+        return cls()
+
+    @GObject.Property(type=bool, default=True)
+    def draw_smallest_child_if_smaller(self) -> bool:
+        """
+        If this widget is smaller than the smallest child widget allows, just
+        draw the smallest child. Ensures that something is always being
+        displayed when at least one child exists.
+        """
+
+        return self._draw_smallest_child_if_smaller
+
+    @draw_smallest_child_if_smaller.setter
+    def draw_smallest_child_if_smaller(self, value: bool):
+        self._draw_smallest_child_if_smaller = value
+
+    @GObject.Property(type=Gtk.Widget)
+    def active_child(self) -> Optional[Gtk.Widget]:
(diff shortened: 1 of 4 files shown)


Neui:
Seeing the v0.6 release coming closer and v0.7 might become GTK 4, I continued trying to make it "snap" into places, but I couldn't figure it out. So I made the label ellipsis-able. Looks like this now:

https://user-images.githubusercontent.com/1196130/151713379-aba2cb05-c09c-40fd-addf-6076aec707bd.mp4

I'm also marking this as "done", so it can be reviewed and tested.

GTG/gtk/browser/adaptive_button.py
```diff
@@ -0,0 +1,283 @@
...
+        """The natural height the widget claims to have"""
+
+    def __init__(self, *args, **kwargs):
+        self._children: List[ChildItem] = []
+        """Child widgets to consider together with internal bookkeeping data"""
+
+        self._child_to_draw: Optional[ChildItem] = None
+        """Current child to draw"""
+
+        self._draw_smallest_child_if_smaller = True
```
diegogangl, line 50: Missed a type hint here

GTG/gtk/browser/main_window.py
```diff
@@ -147,6 +148,16 @@ def __init__(self, requester, app):
...
+        # ad.add(Gtk.Label.new("Start Tomorrow"))
+        # ad.add(Gtk.Label.new("Don't start tomorrow"))
+        # ad.add(Gtk.Label.new("Super short"))
+        # ad.add(Gtk.Label.new("short"))
+        # ad.add(Gtk.Label.new("!"))
+        # ad.show_all()
+        # self.defer_btn.remove(self.defer_btn.get_child())
+        # self.defer_btn.add(ad)
+
         # This needs to be called again after setting everything up,
```
diegogangl, line 161: Forgot some commented code here

GTG/gtk/browser/adaptive_button.py
```diff
@@ -0,0 +1,283 @@
...
+            log.debug("allocation={x=%r, y=%r, w=%r, h=%r}",
+                      allocation.x, allocation.y,
+                      allocation.width, allocation.height)
+
+        self.set_allocation(allocation)
+        for ci in self._children:
+            ci.widget.size_allocate(allocation)
+        self._determine_and_save_active_child()
+
+    # Gtk.Container implementation
```
diegogangl, line 237: This is a sort of "title comment", right? I find useful to surround these with lines, or at least one line after.

```
# -----------------------------------------------------------------------------
# Gtk.Container implementation
# -----------------------------------------------------------------------------
```

Otherwise it looks like commented code at first glance, or some regular comment about the code

GTG/gtk/browser/adaptive_button.py
```diff
@@ -0,0 +1,283 @@
...
+        and saves it in attribute _child_to_draw, so calculation doesn't need
+        to be done when drawing (which may be called often).
+        """
+        old_child = self._child_to_draw
+        self._child_to_draw = self._determine_active_child()
+        if old_child is not self._child_to_draw:
+            self.notify('active-child')
+            self.queue_draw()
+
+    # Gtk.Widget implementation
```
diegogangl, line 124: Same as the gtk.container code below

diegogangl (changes requested):
LGTM, I need to test functionality though.
Wish all the code in GTG was type hinted and nicely documented like this!

GTG/gtk/browser/adaptive_button.py
Neui: Yeah, a section title comment.. I applied the suggestion for now.

diegogangl:
Testing functionality, is this normal?

```
 2022-01-31 08:17:30,865 - WARNING - adaptive_button:do_forall:271 - Got error in for but it should've stay valid: AttributeError("'AdaptiveFittingWidget' object has no attribute '_children'")
```

Neui:
> Testing functionality, is this normal?
> 
> ```
>  2022-01-31 08:17:30,865 - WARNING - adaptive_button:do_forall:271 - Got error in for but it should've stay valid: AttributeError("'AdaptiveFittingWidget' object has no attribute '_children'")
> ```

I also encounter this (happens when closing the main window), but I have not really an idea why this happens, so I just made an special case in the code (catching `AttributeError`) so the exception won't go further.

diegogangl (approved):

PR #734, opened 2021-11-21
Adjust anonymize task script to work with new file format

### LINKED ISSUE #733: Fix/update the tasks anonymization script to match the new file format
For issues that require samples of very complex/heavy task files to trigger them, it would be nice if I could turn my current task file into an anonymized version that could become part of the sample datasets for testing alongside the "Bryce" one. I think this would probably be the only way for devs to encounter (hopefully reliably) issues like #595 and #724.

The file to fix is `scripts/anonymize_task_file.py`

Ideally it should be able to find even flatpaked data files, but in the meantime if it requires specifying a manual file path as the argument (like `./anonymize_task_file.py input.xml anonymized_output.xml` ) then that would already make it useful as a stopgap.

Currently, if you try to use that script with a file with the 0.5 file format, you get this traceback:

```
  File "scripts/./anonymize_task_file.py", line 110, in <module>
    main()
  File "scripts/./anonymize_task_file.py", line 106, in main
    anonymize(xmlfile, outputfile)
  File "scripts/./anonymize_task_file.py", line 52, in anonymize
    contentnode = contentnode[0].childNodes[0]
IndexError: list index out of range
```

Neui:
Also, flatpak detection added because I can, but untested.
And anonymizes tags by using their ID.
And adds the task ID to the title so it is easier to identify them.
And saved searches, where the query is a constant value.

Fixes #733

Testing appreciated since I whipped this in like 30 minutes. Works for neko however.

### DIFF
--- a/scripts/anonymize_task_file.py
+++ b/scripts/anonymize_task_file.py
@@ -16,12 +16,13 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
+import os.path
 import sys
+import re
 
 from xml.dom.minidom import parse
 
 from gi.repository import GLib
-
 
 def anonymize(filename, outputfile):
     try:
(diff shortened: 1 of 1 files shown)


nekohayo:
Can confirm this fix works flawlessly with my personal data file at least!
I only tested direct conversion of a file provided as an argument though, did not test flatpak's data files autodetection.

diegogangl:
It converts subtasks in content into mmm as well. 
`   {! 4d827c65-30fe-4e69-bf37-ae557a2f2718 !} ` is converted into
`   mm mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm mm`

It's not a big deal since subtasks get added at the end of the content when opening the task.
Otherwise works perfectly. Flatpak detection works too.

Neui:
> It converts subtasks in content into mmm as well.

Mainly because this is what the old script did, and since the subtasks are stored separately it shouldn't really matter when debugging relationships (but might matter for the future...)

diegogangl:
Yup, let's move on. Thanks!

Issue #715, opened 2021-10-27
Change shortcut for dismiss task


VaZark:
`Ctrl + I` is practically the universal format for italics. So anytime I reach for italics (out of habit), I end up dismissing the task by accident. It would be more intuitive to use `Ctrl + Shift + D` as _Dismiss_ since `Ctrl + D` is used already for _Done_ because they perform similar actions

koleesch:
My fix for this issue.. https://github.com/getting-things-gnome/gtg/pull/736

PR #736, opened 2021-11-26
fix for issue:715

koleesch:
I have changed the shortcut

### DIFF
--- a/GTG/gtk/backends/backendstree.py
+++ b/GTG/gtk/backends/backendstree.py
@@ -20,7 +20,7 @@ from gi.repository import Gtk
 from gi.repository import GdkPixbuf
 
 from GTG.core.tag import ALLTASKS_TAG
-from GTG.gtk.colors import get_colored_tags_markup, rgb_to_hex
+from GTG.gtk.colors import get_colored_tags_markup, rgba_to_hex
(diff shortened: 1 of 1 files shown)


GTG/gtk/data/help_overlay.ui
```diff
@@ -105,7 +105,8 @@
               <object class="GtkShortcutsShortcut">
                 <property name="visible">True</property>
                 <property name="title" translatable="yes" context="shortcut window">Dismiss Task</property>
-                <property name="accelerator">&lt;Primary&gt;I</property>
+		<!--- <property name="accelerator">&lt;Primary&gt;I</property> -->
```
diegogangl, line 108: No need to keep the commented code around

diegogangl (changes requested):
Hi, welcome to GTG! 
The `help_overlay.ui` file only controls what's shown in the shortcuts window, not the actual shortcuts. Shortcuts are controlled by the actions system in Gtk, and we set them up in the Application class in `application.py`.

As for the commit message, try to write an explanation of what is being changed in the first line. Then leave a blank line and reference the issue. If you write it like `fix #715` the issue will be autoclosed when we merge this

PR #744, opened 2021-12-25
Some GTK4 port changes we can do now

Neui:
These are primarily some (easy) cherry-picks from #723 by @ranchester2.
I've done some changes to the colour RGBA stuff, since I made (or rather copied) the RGBA-to-hex implementation before I saw this was done in a later commit. But I didn't liked the `ignore_alpha` parameter, especially since the usage in the background colour code it had was a bit harder to read (does it ignore now or not?). I opted in to make 2 separate functions.

One explicit GTK3 slipped in because GDK3 `Gdk.RGBA.parse` doesn't seem to support alpha (at least in hex syntax), but GDK4 does. To avoid too much code changes, I took the simpler route.

Since part of the GTK4 migration is moving stuff to UI templates (since it seems you need to use `@Gtk.Template` for now, the `Gtk.Builder.connect_signals` is gone), I thought of maybe also do that (in a new PR), but I am not sure how easy it would then be to combine with the other commits of the port...

I still want to test whenever any other problem occurs. (You can also test as well, or merge and test it while developing)


diegogangl:
Only issue I found is that with colored backgrounds, the tasks that have colored backgrounds end up with white backgrounds (even in dark mode).

Neui:
Thank you, fixed it. The GTK4 liblarch "fixed" this by correctly using `Gdk.RGBA` instead of `Gdk.Color`: https://github.com/ranchester2/liblarch/commit/c191c9e9950e61d44f5c53c11e8c90c12d99ad49#diff-00425c456e04723ec0d1a347e969b436d0f2d1d8144c2064d4efe81eb81db411R338-R341

Also then I noticed the colors are a bit too strong, since it blends for a white background, so I changed it to consider `bgcolor` and basically "reverted" to what GTG was using before this PR.

diegogangl (approved):

diegogangl:
LGTM, we can merge if you are ok with it

Neui:
I'm OK with it being merged now, so it gets some testing.

Issue #747, opened 2022-01-04
Make Caldav workflow more obvious


diegogangl:
This could involve redesigning the entire workflow or adding some text to the panel.
Also, documentation

diegogangl:
Some notes of the top of my head (and Matrix):

- Tasks need to be tagged with 'DAV_[CALENDAR_NAME]` to be synced

> tasks needs to have a certain tag to select the collections they are gonna be synced on, and by default the plugin list all the collection and sync them all 
> I think it's doable to add an optionnal whitelist to select the collection we actually wanto see, and some documentation about the tag behavior 

>  in my implementation, gtg list available todo collection from the server, if the server lists all the collection then all the collection are accessible
> but yeah it should be whitelistable / configurable

(from @jaesivsm in Matrix)

nekohayo:
I suppose this has been mostly addressed by @vansia43's neat additions to the user manual, or is there something else that remains to be done here?

Neui:
From what I can see the main thing was about that it would sync from all collections the server reports rather than just a few that you could select. Some Interfaces also provide a link for CalDAV to a specific collection yet GTG syncs everything rather than looking at the "link-provided" collection.

dmp1ce:
That's right. I was surprised all of my lists were synced from Nextcloud to GTG when I only expected one empty list where I could sync my GTG tasks.

nekohayo:
Right, though that's tracked in issue #751, unless we would want to document this behavior in the user manual?

diegogangl:
Eh, could have worded this better. This was mostly about the issue of having to tag tasks a certain way. I think we could find a more obvious way to link tags to calendars.

vansia43:
> Right, though that's tracked in issue #751, unless we would want to document this behavior in the user manual?

Got a note in the caldav page about this: “The CalDAV provider might give you a URL for a specific calendar, but actually send the todos from all calendars. In this case, the CalDAV backend will end up syncing all calendars.”

So should be good there 😀

Issue #740, opened 2021-12-23
Update the User Manual for 0.6


vansia43:
I am using this as a tracking issue for user manual updates. I have made the following updates on my end, but I was waiting to open a PR until it was determined if there will be documentation for CalDav (#525) as I will have to update the index and possibly resurrect some of the sync-services pages that existed before 0.4 to include that. 

So far I have identified the following changes that are needed to the user manual based on current closed issues/PRs:

- Plugins Page: Add Gamify plugin and description as this was missing
- Setting Task Start Date and Due Date (was missing from last time): Need new image of Set Dates from Task Browser to show Set Recurring
- Shortcut Keys Page: Add Keyboard focus shortcut CTRL+B Sets focus on the sidebar (#648) and F10 opens Main Menu (#664)
- Overview Page: Add info about additional new options from the hamburger menu (#662 and #563)
- Editing Tag Properties (#276): Need new screenshot and to replace the current one and need to update this whole page to match the description of the edits to this UI

vansia43:
Closed via #748 .

PR #748, opened 2022-01-06
User Manual Updates for 0.6

vansia43:
Addresses all items listed in #740. 

Per #747, it sounds like CalDav docs will go in the next release or not right now.


diegogangl (approved):

diegogangl:
Thanks for the PR! We should have some docs for caldav at release though

docs/user_manual/C/gtg-plugins.page
```diff
@@ -39,6 +40,9 @@
 <p><em>Export and Print</em></p>
 <p>Exports the tasks in the current view into a variety of formats.</p>
 
+<p><em>Gamify</em></p>
+<p>Gamify adds a game aspect to your GTG workflow.</p>
```
Neui, line 44: Maybe a short mention what "game aspect" means here? Like "(like task completion streaks)".

Neui (approved):
Changes are OK.

I am wondering whenever we should be more "distro-neutral" in the `set_date_task_browser.png` image (where you can see the background image). Additionally, if replaced with solid colour (or even transparency), then we might save some kilobytes because of better image compression (maybe run something like `optipng` on those 2 changed images, to get the most out of it?)

docs/user_manual/C/gtg-plugins.page
vansia43: Good point! Added a bit more of a description for this.

vansia43:
> Thanks for the PR! We should have some docs for caldav at release though

I made the 2 requested changes...

As for CALDAV. I hear you. I tried to grab the latest and greatest code to test it out, but I am having trouble getting it to work on my end though, so I can't really come up with any docs for it. >.<

diegogangl:
Maybe it'd be easier to have @jaesivsm write some notes, and then we can put the docs together from that

nekohayo (approved):
It looks good to me too, thank you for paying attention to these details!

diegogangl:
Merged, thanks! We can work on the caldav docs on another branch

Shahin-rmz:
Hello, I have followed #407 issue and end up being here, where I can see, the caldav sync feature is already done, and is waiting for documentation, until in next release it implemented.
There is a question though, how can we update and/or install the plugin in order to be able to use it?
Thanks

dmp1ce:
With the latest code, are you getting any errors? I was missing the caldav python module which I only noticed because of a warning on the CLI.

diegogangl:
Good point, we also need to include that in the docs (at least for now).

Shahin-rmz:
> With the latest code, are you getting any errors?
I haven't got any error, actually I don't know how to use CalDave. GTG is updated but don't know how to connect to caldav.

Neui:
> I haven't got any error, actually I don't know how to use CalDave. GTG is updated but don't know how to connect to caldav.

Also took me a while, you need to have `caldav` python packgae installed (I did it via `pip3 install caldav`), then go to the menu top left → Synchronization, then on the plus bottom left. The right side should show a "Add backend" thing with caldav automatically being selected with some options and an OK bottom right.

Shahin-rmz:
> Also took me a while...
Thanks so much for the fast answer.
I have actually installed caldav now.
image
But after that, the sync button is still deactivated.
image

Shahin-rmz:
not sure if here is the right place to talk about installation AND if I missed any dependencies.
App works fine but with sync button deactivated.

Neui:
> App works fine but with sync button deactivated.

What version are you on? 5.0 or the git version? You need the current git version to make it selectable.

Shahin-rmz:
image
I have version 0.5, not 5!
I'll check for git version.
let's hope I can find it within documentation. :)

Shahin-rmz:
image
I have installed flatpak version, pulled git version and tried to update it and that was the result.

diegogangl:
Hi @Shahin-rmz , did you run GTG with the `./launch` command? Was the flatpak version from flathub (that one is 0.5)?

Shahin-rmz:
Hi @diegogangl , you are right.
flatpack was from flathub. and I didn't run it via ./launch command.
just double click.

PR #749, opened 2022-01-08
Don't loop forever with invalid quick add date

Smittyvb:
Quick entering `Do a thing due:meh` causes GTG to completely freeze, and need to be killed. This fixes it to treat tokens with invalid dates the as other invalid tokens, ignoring them (and removing them from the input text).

This issue was caused since the the `continue` in the parsing loop didn't remove the token, causing it to loop forever on the invalid token.

### DIFF
--- a/GTG/gtk/browser/quick_add.py
+++ b/GTG/gtk/browser/quick_add.py
@@ -85,20 +85,20 @@ def parse(text: str) -> Dict:
             try:
                 result['start'] = Date.parse(data)
             except ValueError:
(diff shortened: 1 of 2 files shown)


Neui (approved):
Was able to reproduce the issue and this fixed the issue. Test seems also work (just hangs, which hopefully would indicate something is wrong).

diegogangl (approved):

diegogangl:
Merged, thanks!

PR #750, opened 2022-01-10
Save task before creating parent

### LINKED ISSUE #745: Changes to child task titles are not saved when using the button to create/open the parent
1. Create (or open) a task
2. Type some additional words in the title of that task (not the body*)
3. Click the "Open parent" (or "Add parent") button

Result: the additional words you had typed in the title are lost.

 *: somehow, the body is not affected by this bug...

Rating the severity as critical because technically it can mean data loss for users, but low-hanging-fruit because it is most likely a trivial fix...

diegogangl:
Saving makes sure the task has the correct title when being added as a
subtask to the new parent. Otherwise the title gets overwritten with the
default title.

fixes #745

### DIFF
--- a/GTG/gtk/editor/editor.py
+++ b/GTG/gtk/editor/editor.py
@@ -787,6 +787,7 @@ class TaskEditor:
         and to prevent visible content divergence when the child title changes.
         """
         parents = self.task.get_parents()
+        self.save()
 
         if not parents:
             tags = [t.get_name() for t in self.task.get_tags()]


nekohayo (changes requested):
Tested, it works for the case where you create a brand new task and then click "Add Parent", but it doesn't work in the case where you open an existing child task, change its title and immediately click the "Open Parent" button.

diegogangl:
Force pushed a fix

Neui:
Looks similar to my fix in the issue, so the title thing works better, but still has the problem that if you type too fast and click add parent fast, then it still won't get applied.

https://user-images.githubusercontent.com/1196130/149037479-65c69e0b-18e4-4558-b19c-b36640f6e9cf.mp4

Of course you have to be quite fast, so it might not be a realistic scenario, but I'd still consider a bug.

nekohayo:
I can confirm that from a normal user's standpoint, the issue is solved with the revised commit, though "in theory" still exists if you click ultra fast (as in, less than a quarter of a second delay or something like that) which I can't imagine someone encountering unless they're actively _trying_ to trigger the bug.

So if there's a way to guard against that theoretical bug without adding much more code complexity then that's ideal, as it might prevent some other unforeseen issues maybe (race conditions or domino effects, anyone?), otherwise I wouldn't be utterly shocked if the fix solves only the user-visible problem and "hopes" it won't be a real problem beyond that.

diegogangl:
A potential fix would be using `self.textview.process()` directly. The problem is that it calls light save which has a timer, so you don't get the effects of this fix. The proper solution would be to revisit taskview to use signals, and whether we really need light save (since lxml does everything in a few ms). Then we could call `process()` while disabling the signals, and then call `save()` immediately after.

diegogangl:
Let's go ahead and merge this for now. I plan on revisiting the textview and the editor in the near future anyways (once we complete the Gtk4 transition at least)

PR #752, opened 2022-01-11
Minor updates to contributors documentation

nekohayo:
This is pretty much trivial stuff I wanted to correct, that has been sitting around in my todo list for way too long ;)
Probably should not break anything or be inaccurate, but making a pull request out of it just in case. Would recommend fast-forward (rebase-type) merging (which is also probably a policy we should encourage/document somewhere...).


diegogangl (approved):

vansia43 (approved):
LGTM!

diegogangl:
Merged!

PR #754, opened 2022-01-12
Prevent errors when no task is selected

### LINKED ISSUE #753: ctrl+shift+p after startup/filter change throws uncaught exception
After startup or changing filters, nothing will be selected and ctrl+shift+p to create a new parent will throw an uncaught exception. This means the "crash" dialog comes up, but since it is not fatal you can still continue using GTG.

```
  File ".local_build/install/lib/python3.8/site-packages/GTG/gtk/application.py", line 338, in add_parent
    if self.browser.have_same_parent():
  File ".local_build/install/lib/python3.8/site-packages/GTG/gtk/browser/main_window.py", line 1385, in have_same_parent
    first_task = self.req.get_task(selected_tasks[0])
IndexError: list index out of range
```

Found out because I wanted to look if pressing that shortcut in an taskeditor window would create an parent task of that task.

diegogangl:
This can be triggered when GTG first starts and there are no selected
tasks.

fixes #753

### DIFF
--- a/GTG/gtk/application.py
+++ b/GTG/gtk/application.py
@@ -335,8 +335,14 @@ class Application(Gtk.Application):
 
     def add_parent(self, param, action):
         """Callback to add a parent to a task"""
-        if self.browser.have_same_parent():
-            self.browser.on_add_parent()
+        
+        try:
+            if self.browser.have_same_parent():
+                self.browser.on_add_parent()
+
+        # When no task has been selected
+        except IndexError:
+            return
 
     def edit_task(self, param, action):
         """Callback to edit a task."""


GTG/gtk/application.py
```diff
@@ -335,8 +335,12 @@ def new_subtask(self, param, action):
 
     def add_parent(self, param, action):
         """Callback to add a parent to a task"""
-        if self.browser.have_same_parent():
-            self.browser.on_add_parent()
+        try:
+            if self.browser.have_same_parent():
+                self.browser.on_add_parent()
+
+        except IndexError:
```
Neui, line 342: I think at least a comment on what this error causes would be nice, like `# When no task has been selected`.

Neui (approved):
Seems to work

diegogangl:
Thanks for reviewing

PR #759, opened 2022-01-20
Make Synchronization menu entry keyboard selectable

Neui:
Basically #665 but for that entry that (wrong) property was introduced in the CalDav changes it seems.


diegogangl:
Merged, thanks!

PR #764, opened 2022-02-01
Update create-dismiss-delete for Reopen action

vansia43:
After working on NEWS, I realized I forgot to update this page when I was updating the rest of the user manual. I missed this feature. Includes minor changes to page. I tested everything else on the page/updated text as needed.


diegogangl:
Thanks!

PR #766, opened 2022-02-03
fixes #765

### LINKED ISSUE #765: App hangs with recurring task that lands on Feb 30th
Found in version 0.5, flathub on Ubuntu 21.10

# How to reproduce

1. Create a task with a recurring date of the 30th of each month
2. Mark the task as done until it hits 30/01/2023
3. Mark the task as done and the App hangs and must be force quit

# Expected Behavior

1. App not to crash
2. Select a reasonable alternative?
  - 28/02 Feb
  - Or 29/02 if leap
  - Alternatively skip to 30/03 (not ideal)

zeddo123:
closes #765

### DIFF
--- a/GTG/core/dates.py
+++ b/GTG/core/dates.py
@@ -464,8 +464,7 @@ class Date:
     def _parse_only_month_day_for_recurrency(self, string, newtask=True):
         """ Parse next Xth day in month from a certain date"""
         self_date = self.dt_by_accuracy(Accuracy.date)
-        if not newtask:
-            self_date += timedelta(1)
+
         try:
             mday = int(string)
             if not 1 <= mday <= 31 or string.startswith('0'):
@@ -478,7 +477,10 @@ class Date:
         except ValueError:
             result = None
 
-        if result is None or result <= self_date:
+        # If it's not possible to recreate the date by replacing or
+        # newtask=False and the result is still less than self_date,
+        # we need to jump to the next month.
+        if result is None or (not newtask and result <= self_date):
             if self_date.month == 12:
                 next_month = 1
                 next_year = self_date.year + 1
--- a/GTG/core/dates.py
+++ b/GTG/core/dates.py
@@ -483,8 +483,12 @@ class Date:
                 next_month = 1
                 next_year = self_date.year + 1
             else:
-                next_month = self_date.month + 1
                 next_year = self_date.year
+                next_month = self_date.month + 1
+                max_mday = calendar.monthrange(next_year, next_month)[1]
+                if mday > max_mday:
+                    mday = max_mday
+
 
             try:
                 result = date(next_year, next_month, mday)


Neui:
Could you please use a more descriptive commit message? So I don't have to look up what issue 765 is in the first place when looking at the history.

zeddo123:
Sorry, yeah I'll rename the commit.

>I think we can take out that timedelta(1) increment.

I'll do some testing to sure, because I remember it being important in some usecases.

zeddo123:
I've done some testing, and the 31th of the month bug shouldn't happen anymore. @Neui if you have time, do some more testing to be sure.

GTG/core/dates.py
```diff
@@ -478,13 +477,17 @@ def _parse_only_month_day_for_recurrency(self, string, newtask=True):
         except ValueError:
             result = None
 
-        if result is None or result <= self_date:
+        if result is None or (not newtask and result <= self_date):
```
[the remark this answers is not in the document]
zeddo123: I was going to but I felt that I would right a whole paragraph to explain it -_-.

So we need `not newtask` so that when it the first time the task is being enable it does not go to the next occurrence and when it's not the first, we make sure that the new date (result) is superior to `self_date` (to avoid an infinite loop).

diegogangl:
Works for both february and 30/31 months. It'd be nice to have a comment explaining the `if` though (newtask, etc).

zeddo123:
> Works for both february and 30/31 months. It'd be nice to have a comment explaining the `if` though (newtask, etc).

done.

GTG/core/dates.py
```diff
@@ -478,13 +477,20 @@ def _parse_only_month_day_for_recurrency(self, string, newtask=True):
         except ValueError:
             result = None
 
-        if result is None or result <= self_date:
+        # If it's not possible to recreate the date by replacing or
+        # newtask=False and the result is still less than self_date,
+        # we need to jump to the next month.
```
Neui, line 482: You're just repeating the statement, you are missing the important WHY you are doing that. Maybe:

```suggestion
        # When creating a new task, make sure to use the current month instead
        # of the next month (when today_date <= mday), so that when creating a
        # recurring task the user should do the task "now".
        # (This also causes to be start_date > due_date)
```

(I also tried removing `not newtask`, and I think the behaviour is more reasonable with `result is None or result <= self_date`, so that when creating a recurring monthly task, if it is a previous day then don't do it this month, otherwise if today or later (or the start date or later) do it this month).

diegogangl:
Merged, thanks!

GTG/core/dates.py
zeddo123: >I think the behaviour is more reasonable with result is None or result < self_date

I've also tried it before, but when I tested it made other cases not work correctly.

Neui: I did noticed later that `result < self_date` is buggy in case `result == self_date` when checking the latter date would result in an infinite loop, but I still think this behaviour can be made better (with proper reasoning in the comments)

zeddo123: I've made a new PR that rewrites the function. let me know what you think. :)

PR #769, opened 2022-02-04
Cleanup _parse_only_month_day_for_recurrency

zeddo123:
I tried rewriting the whole function to make more clean and clear.

### DIFF
--- a/GTG/core/dates.py
+++ b/GTG/core/dates.py
@@ -461,6 +461,25 @@ class Date:
         else:
             raise ValueError(f"Can't parse date '{string}'")
 
+    @staticmethod
+    def date_in_the_next_month(mday, dt):
+        if dt.month == 12:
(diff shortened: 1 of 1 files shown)


diegogangl (changes requested):
Overall it looks good, and it works as well as it did before. Some unit tests might be good, but I'm not sure we should be spending so much time on this. Dateutils can do this already (https://dateutil.readthedocs.io/en/stable/relativedelta.html) and we could probably use it for other things too. 
We should take another look at this date module in the next cycle

zeddo123:
> but I'm not sure we should be spending so much time on this. 

Yeah just for now until we rewrite the repeating tasks.

diegogangl (approved):

diegogangl:
Thanks!

PR #772, opened 2022-02-07
Only remove recognized tokens in quick add

### LINKED ISSUE #771: Cannot add task with colon in the title by "Quickly create tasks here"
When I try to create new task containing colon in its title, like `some:thing`, using "Quickly create tasks here" bottom bar, my new task is created without any content, with just default "my new task" title. It does work properly with full task creation dialog, however. There is no extra information when running GTG with `--debug` switch.

I'm using GTG v0.5 as Flatpak package from Flathub. 

The problem does not occur with GTG v0.4 which I tested by downgrading to Flatpak commit 4d47040f32b109d6374f8f2fcb951e31640fd69a56216b8cd8c88cedf747f5b8

diegogangl:
Also switch to a for loop for matches, this prevents the while loop from
accidentally going infinite.

Also don't add tags if they are empty. For instance, "tags:some,other,,"
would have added two empty tags at the end (single @s).

Also use replace to remove text instead of slicing (more reliable)

fixes #771

### DIFF
--- a/GTG/gtk/browser/quick_add.py
+++ b/GTG/gtk/browser/quick_add.py
@@ -69,27 +69,33 @@ def parse(text: str) -> Dict:
         data = match.group(0)
         result['tags'].add(data[1:])
 
-    while match := re.search(TOKEN_REGEX, text):
+    for match in re.finditer(TOKEN_REGEX, text):
         token = match.group(2)
         data = match.group(3)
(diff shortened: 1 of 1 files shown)


Neui (approved):
Works.

diegogangl:
Thanks for testing!

PR #775, opened 2022-02-08
User docs: mention the possibility of using Unison to manually sync data

nekohayo:
This documents a technique I've been using since the onset of GTG, as an alternative serverless manual/batch sync trick. It doesn't go in depth about it, but presents the possibility, which may help some users.


vansia43:
LGTM. One spelling correction needed: continously > continuously

vansia43 (approved):

Neui (commented):
Looks OK

PR #743, opened 2021-12-24
Fix for issue: 742

unsupported-transceiver:
Fix typo in default tasks (dot to space)

Fixes #742

PR #776, opened 2022-02-09
Update first task about sync services to the current state

### LINKED ISSUE #742: Cosmetic bug: tutorial task text typo
The tutorial task "Learn About Synchronization Services" contains a dot

> ... to work with the new codebase.or to interface with the latest APIs ... 

which should be a space
>... to work with the new codebase or to interface with the latest APIs ...

(This is my first issue here, sorry for the Inconvenience)

Neui:
Also contains the commit of #743 (which fixes #742) but with a better commit message.

Not sure if I should remove the mention of unsupported services (like Tomboy or RTM) because they can't be used anyway, so why bother mentioning them?


diegogangl:
I would remove them. I don't think there's a lot of demand for RTM anymore. Tomboy is unmaintained and as far as I can see, tomboy-ng is a full rewrite in pascal. It's unlikely these two will come back soon.

diegogangl:
Thanks!

PR #777, opened 2022-02-10
Improve plugin error messages english

Neui:
Small text changes that bugged me while I was first updating the translation of GTG to german.


diegogangl:
Thanks!

PR #778, opened 2022-02-11
Harmonize the headerbar's task deferral popover menubutton labels with the right-click menus'

nekohayo:
This is an attempt at resolving some of the consistency problems presented in issue #550.

It doesn't solve them _100%_ because I couldn't find a way to access the individual menu items in the right-click menus like you can with GtkPopover's menu button widget IDs, so I got stumped by that. I tried some things but they didn't work and they also looked like nasty hacks. If you have ideas/suggestions of how to achieve that easily (or maybe it's just possible in GTK4, if the menus changed there?...) I'd be happy to try them, but I think I've run out of patience for now. The existing changes should already be a noticeable improvement, and maybe someone else can come in to fix the remaining part (right-click menus) on top afterwards.

I have tested this to be working both in English and my locally translated version, so while this adds some new translatable strings, it should still be entirely translatable. I would probably provide an updated FR translation separately from this branch, afterwards.

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -282,17 +282,20 @@ class MainWindow(Gtk.ApplicationWindow):
         self.calendar.connect("date-changed", self.on_date_changed)
 
     def _set_defer_days(self, timer=None):
-        """Set days for the defer task menu."""
-
-        # Today is day 0, tomorrow is day 1. We don't need
-        # to calculate the weekday for those.
-
+        """
+        Set dynamic day labels for the toolbar's task deferring menubutton.
+        """
         today = datetime.datetime.today()
-
+        # Day 0 is "Today", day 1 is "Tomorrow",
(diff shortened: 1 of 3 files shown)


diegogangl:
I think the proper way to do this would be to create the menu using code rather the XML. Then we have references to the `GMenuItems` that we can change using `setlabel()`
https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/MenuItem.html#Gtk.MenuItem.set_label

diegogangl:
This looks ok tho, we could merge this until we re-do the menu in 0.7

nekohayo:
@diegogangl I could do that if you prefer it that way, but I thought the emdash approach—even if I cheated and added spaces around it—looked cleaner and clearer than with the parentheses… Don't the parentheses look a bit cluttered?

diegogangl:
It does look a bit more cluttered, but I found the dash more jarring. Anyways, we can re-harmonize properly once we're in Gtk4

PR #781, opened 2022-02-11
Traceback catcher "Error popup" UI/UX refinements

nekohayo:
Probably of interest to @Neui (as this builds upon his merged work from MR #582) and others!

This is how it looked like before:

2022-02-11 debug mode and traceback catcher GUI - before

And now:

2022-02-11 debug mode and traceback catcher GUI - after 2, v3

The about dialog also shows a human-readable app name for the `launch.sh` version:

2022-02-11 debug mode and traceback catcher GUI - about dialog after 1 2022-02-11 debug mode and traceback catcher GUI - about dialog after 2


GTG/gtk/errorhandler.py
```diff
@@ -24,25 +24,23 @@ def __init__(self, exception=None, main_msg=None, ignorable: bool = False, conte
 
         formatting = {
             'url': info.REPORT_BUG_URL,
-            'urlname': info.REPORT_BUG_URL,
-            'appname': info.NAME,
+            'urlname': _("our issue tracker"),
```
Neui, line 27: Add a new line before it with a comment for the translators to have some context: `# Translators: Used as the link name in the error handler dialog (that goes to the issue tracker)`

GTG/gtk/errorhandler.py
```diff
@@ -24,25 +24,23 @@ def __init__(self, exception=None, main_msg=None, ignorable: bool = False, conte
...
-        desc2 = _("""Please report the error with the following details and steps to trigger it to:
-<a href="{url}">{urlname}</a>
-Warning: Exiting may lose unsaved changes.""")
-        desc2 = desc2.format(**formatting)
-
         if ignorable:
-            title = _("{appname} encountered an error")
-            desc = _("""{appname} encountered an internal error, but it'll continue running.
-However, more unexpected things can happen, and thus be careful.""")
+            title = _("Internal error")
```
Neui, line 31: I would have included the appname at least in the title so that you can see which application created this window without reading the content.

GTG/gtk/errorhandler.py
nekohayo: I actually hesitated in just hardcoding this right into the sentence's string, instead of keeping it as a variable. Or do we suspect distros might patch this and the URL to point to their bug tracker?

GTG/gtk/errorhandler.py
nekohayo: Would appending " — GTG" at the end of those titles do the trick?

Neui: Yes

GTG/gtk/errorhandler.py
```diff
@@ -108,27 +108,31 @@ def _update_additional_info(self):
 
         text = ""
         if self._context_info is not None:
-            text = text + "Context: " + self._context_info + "\n\n"
-        text = text + body + "\n\n"
+            text = text + "**Context:** " + self._context_info + "\n\n"
+        text = text + "```\n" + body + "```\n\n"
```
Neui, line 112: Maybe use

    ```python-traceback

so we have some colours to look at. Completely optional.

nekohayo: I didn't know that was a thing! Very cool. Will do :)

GTG/gtk/errorhandler.py
Neui: If we use english there I am ok with embedding `urlname` in the description, but I'd keep the `url` separate so we (and distro maintainers/packagers) can change the issue URL without invalidating the translations.

Neui:
I'm OK with this, but now the window got seemingly taller when expanded, which I'd say isn't optimal. Maybe look if we can "fix" this before merging.

Neui:
This fixed the tall window issue for me:

```diff
diff --git a/GTG/gtk/errorhandler.py b/GTG/gtk/errorhandler.py
index 70b3ea11..5fbdf8d4 100644
--- a/GTG/gtk/errorhandler.py
+++ b/GTG/gtk/errorhandler.py
@@ -68,6 +68,8 @@ Please report the bug in <a href="{url}">{urlname}</a>, with steps to trigger th
         self._expander.set_label(_("Details to report"))
         self._expander.add(expander_content)
         self.get_content_area().add(self._expander)
+        self.props.width_request = 450  # So window won't get unconformable tall
+        expander_content.props.height_request = 200
         self._expander.bind_property("expanded", self, "resizable",
                                      GObject.BindingFlags.SYNC_CREATE)
         self._expander.show_all()
```

Neui (approved):
Looks OK to me.

Neui:
> * The difference in padding in the first paragraph looks kind of weird, can it be made to be the same as the others?

This seems because both text are centered (so they stay centered when resized), but are limited to 60 characters in width in a variable width font and word wrapping.
Setting `halign` to `START` and `xalign` to `0,0` for both labels makes it better, but then they stay on the left when resized, so you also need to make the parent `GtkBox` `halign` to `CENTER`.

> * Is there really a need for the title to end with "- GTG"? These dialogs are modal and stuck on top of the app, so you always know where they are coming from

Yes, I look at the title first to know which application is wanting attention from me (especially when something happened while in the background). Also I think it doesn't hurt, I don't really see any upside when removing that part.

nekohayo:
> The difference in padding in the first paragraph looks kind of weird, can it be made to be the same as the others?

Could it be that this is standard GNOME GTK styling thing with the primary vs secondary dialog text? I simply assumed it was a enforced by GTK, as I didn't see something that explicitly sets that kind of alignment/margins/padding in the existing code (and my newly added borders that serve as margins have nothing to do with that). In the code, this is where the text insertion happens:

```
        self.set_markup(desc)
        self.props.secondary_text = desc2
```

As you can see the GtkDialog seems to have this distinction between primary and secondary text...

> The "exit" button performs a destructive action (quit GTG and accept potential data loss), so it should have the destructive action CSS class.

Great suggestion, and thank you for pointing for the exact code example. I've been able to add this to my code and will push another commit on top for that.

> I'd also add some padding to the left and the right of the code expander

I think I tried to do that and was unable to, it either did nothing or added padding above/below it, or something like that.

diegogangl:
Code and recent changes look good.

> Yes, I look at the title first to know which application is wanting attention from me (especially when something happened while in the background). Also I think it doesn't hurt, I don't really see any upside when removing that part.

Ah sorry, I haven't used i3 in several years. I assumed Gtk3 apps overlaid their dialogs on top of the app like they do in shell. Yeah, it's only a cosmetic detail. Won't hurt to leave it there.

> Could it be that this is standard GNOME GTK styling thing with the primary vs secondary dialog text? I simply assumed it was a enforced by GTK, as I didn't see something that explicitly sets that kind of alignment/margins/padding in the existing code (and my newly added borders that serve as margins have nothing to do with that). In the code, this is where the text insertion happens:

Yeah it's most likely the secondary text thing. Maybe we can not set the secondary text, and just put everything in the same string?

Neui:
> Could it be that this is standard GNOME GTK styling thing with the primary vs secondary dialog text? I simply assumed it was a enforced by GTK, as I didn't see something that explicitly sets that kind of alignment/margins/padding in the existing code (and my newly added borders that serve as margins have nothing to do with that)

This is because they use 2 labels that are centered but have different "fixed width", see https://github.com/getting-things-gnome/gtg/pull/781#issuecomment-1037616693, which also mentions a possible fix.

diegogangl:
The eagle has landed :+1:

PR #786, opened 2022-02-13
Fix adaptive widget offset when maximized in gnome

### LINKED ISSUE #785: Adaptive headerbar defer ("Do it tomorrow") button label is misplaced when window is tiled or maximized
On GNOME Shell 3.38 (can't test anything else) on Xorg, when GTG is a non-maximized and non-tiled window, it renders fine:

freestanding window

However, if you half-tile it (`Super+Left`) and look closely, you can see the label has shifted its position:

tiled

...and if you maximize the GTG window, the label appears invisible (but it's actually lower below, "behind" the treeview widgets; I spotted it positioned there once, but was not able to reproduce it that way for screenshots):

maximized

Neui:
For some reason when using gnome-shell and maximizing GTG the widget
will be offset (just enough you can still see part of it outside the
button).

After reading more code it seems we need to resize the Gdk.Window when
doing size_allocation (which some of the GTK code doesn't do because it
was hidden behind some CSS helper functions it seems). We just call the
Gtk.Widget.do_size_allocation because they do it already for us.

Fixes #785

### DIFF
--- a/GTG/gtk/browser/adaptive_button.py
+++ b/GTG/gtk/browser/adaptive_button.py
@@ -209,6 +209,11 @@ class AdaptiveFittingWidget(Gtk.Container):
         self.set_realized(True)
 
         allocation = self.get_allocation()
+        if self._spammy_debug:
+            log.debug("allocation={x=%r, y=%r, w=%r, h=%r}",
+                      allocation.x, allocation.y,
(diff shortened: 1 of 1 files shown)


diegogangl:
Thanks!

PR #787, opened 2022-02-13
Gamify 0.1.1

### LINKED ISSUE #782: Gamify plugin tracebacks when trying to change the number of daily tasks
This is what I got the first time I tried to bump the number of daily tasks, in the git version:

```python-traceback
  File "GTG/plugins/gamify/gamify.py", line 411, in on_preferences_changed
    ui_mode = int(self.get_ui_mode_combo_value())
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

Strangely, in subsequent tries, it didn't seem to traceback. Maybe something to do with saving stuff?

### LINKED ISSUE #783: Gamify substracts 1 when marking a task as not to be done (dismissing), and when reopening a task that was dismissed
Marking a task as done (`Ctrl+D`) increases the daily tasks counter by 1, however:

* Dismissing a task (`Ctrl+i`) _substracts_ 1 from the counter; shouldn't it leave the counter untouched?
* "Reopening" a task substracts 1 from the counter, which is correct if the closed task was a closed "done" task, but if it was a closed "dismissed" task then you're creating a point deficit, whereas it should probably be neutral instead? (currently, combined with the other issue above, this means you can actually -2 the total if you dismiss a task and then undismiss it)

I'm just guessing that these are bugs rather than intended behavior :)

zeddo123:
Patching some bugs present in the first version of gamify.

closes #782
closes #783

### DIFF
--- a/GTG/core/requester.py
+++ b/GTG/core/requester.py
@@ -36,7 +36,7 @@ class Requester(GObject.GObject):
     Multiple L{Requester}s can exist on the same datastore, so they should
     never have state of their own.
     """
-    __gsignals__ = {'status-changed': (GObject.SignalFlags.RUN_FIRST, None, (str, str,))}
+    __gsignals__ = {'status-changed': (GObject.SignalFlags.RUN_FIRST, None, (str, str, str,))}
 
     def __init__(self, datastore, global_conf):
         """Construct a L{Requester}."""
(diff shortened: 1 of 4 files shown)


diegogangl (approved):
LGTM

nekohayo (approved):
Changes have been rebased and merged to master

PR #801, opened 2022-02-20
Fix menu button alignment from plugins

### LINKED ISSUE #800: "Send via e-mail" is indented in the task menu

It's a small bug, but still a visible glitch.

Neui:
It seems using Gtk.ModelButton.set_label() inserts a label widget, but
Gtk.ModelButton.props.text also adds a Gtk.Box around the label.

The first method makes the label centered, which looks out of place with
the other menu entries, but the latter does not. So just switch it out
to maintain consistency.

Interestingly, the dev console already uses the `text` property, so it
wasn't affected since it was already doing it correctly.

Fixes #800


diegogangl:
Thanks!

PR #804, opened 2022-02-20
Add new files to POTFILES.in for translation

### LINKED ISSUE #799: Missing translations for plugins

There are missing translation source for a Hamster button and a task menu item. 

I'm also not sure what is the reason, but none of the string in "Plugins" list from menu is localized (should be easy to check if it's also the same, I just have to many small tasks right now to make sure).

Neui:
I used the following command to find out some of the files:

    diff -u \
	<(sort po/POTFILES.in | egrep -v '\.ui|\.desktop') \
	<(grep -r '[^_]_(' GTG --files-with-matches --include='*.py' | sort) | xclip -in -selection clipboard

Fixes #799, at least they show up (but haven't checked whenever they also really are applied, just took a look at `po/gtg.pot` after `ninja -C .local_build/ gtg-pot gtg-update-po`)

### DIFF
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -21,9 +21,11 @@ GTG/gtk/data/preferences.ui
 GTG/gtk/data/task_editor.ui
 GTG/plugins/export/export.ui
 GTG/plugins/gamify/prefs.ui
+GTG/plugins/hamster/prefs.ui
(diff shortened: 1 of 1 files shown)


diegogangl:
Funny hamster wasn't in that list already though, thanks!

kocio-pl:
It should change the `gtg.pot` file, I think this step is lacking in the current code.

PR #805, opened 2022-02-20
Remove .gtg-plugin in ./launch.sh for translators

Neui:
More of a workaround since external build system setups (not
`./launch.sh`) still have the buggy behaviour.

From #729 as a suggested alternative workaround (and lastly mentioned in #799 because I forgot already).


diegogangl:
Will do until we figure out a better way, thanks!

PR #819, opened 2022-02-26
Fix missing translator in pl

kocio-pl:
Adding missing translator credits in Polish and cleaning language team e-mail (there's no such thing).


diegogangl:
Thanks!

PR #824, opened 2022-02-27
Skip ending dots in tags

### LINKED ISSUE #807: Dots are treated as part of a tag while commas are not

It's an interesting question how GTG should treat commas, dots and other punctuation marks at the end of a tag, because task description can be simply a sentence. Currently it looks strange to me that dots get included as part of the tag.

kocio-pl:
Fixes #807

Allows dots only to be used only inside tags, but not at the end, so tags can be clearly used in sentences like:

> will also appear when you select `@money`.

> I wish I have more `@time`...


diegogangl:
Great PR, thanks!

PR #833, opened 2022-02-28
Fixing export plugin PDF template

### LINKED ISSUE #831: Export plugin PDF template not working
Export plugin PDF template is not working because it lacks `pdf90` and `pdfnup` commands. They are are required by `GTG/plugins/export/export_templates/script_pocketmod`, but are not visible as dependencies in plugin description: 

https://github.com/getting-things-gnome/gtg/blob/4bd89aa55cb15b5cd077afd5fc08c2718ef51528/GTG/plugins/export.gtg-plugin#L9

nor in documentation:

https://github.com/getting-things-gnome/gtg/blob/master/README.md#solving-dependencies-for-plugins-optional

I have found only the latter:

https://pypi.org/project/pdfnup/

but I can't install it under Ubuntu 20.04 via pip because of an error:

```
    Complete output (8 lines):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-6jendb1c/pdfnup/setup.py", line 7, in <module>
        import ez_setup
      File "/tmp/pip-install-6jendb1c/pdfnup/ez_setup.py", line 94
        except pkg_resources.VersionConflict, e:
                                            ^
    SyntaxError: invalid syntax
```

kocio-pl:
Fixes #831

Converting old wrapper scripts `pdf90` and `pdfnup` directly to `pdfjam`.

It works for me with default first-run task list, but please test with much more tasks, at least to fill the whole 8 pages.

### DIFF
--- a/GTG/plugins/export/export_templates/script_pocketmod
+++ b/GTG/plugins/export/export_templates/script_pocketmod
@@ -27,9 +27,9 @@ pdflatex -interaction nonstopmode -output-directory $TMPFOLDER -jobname temp $SO
 PDFFILE=$TMPFOLDER/temp.pdf
 pdftk $PDFFILE cat 6 7 8 1 output $TMPFOLDER/seite6781.pdf   1>&2
 pdftk $PDFFILE cat 5 4 3 2 output $TMPFOLDER/seite5432.pdf   1>&2
-pdf90 $TMPFOLDER/seite5432.pdf --outfile $TMPFOLDER/seite5432-r.pdf      1>&2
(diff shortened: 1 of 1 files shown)


diegogangl:
You can test with a different set of tasks using the `-s` parameter. You could try with the screenshots set for instance, `./launch -s screenshots`. That should be enough to cover 8 pages I think

diegogangl:
🤔️ Let's merge this for now, since it seems to work and I doubt it can break the export plugin more. We can go back when we rewrite for the new core.
Thanks!

PR #849, opened 2022-03-05
Fixing PDF export plugin

kocio-pl:
This commit completes the fix for #831 - follow up from #833.

Also dropping German language specific TeX option.

gtg


diegogangl:
Thanks!

Issue #803, opened 2022-02-20
Problem with user guide tags localization


kocio-pl:
Forked from https://github.com/getting-things-gnome/gtg/issues/793#issuecomment-1046144111

When I try to make localized tags in GTG first run user guide instead of original ones, it has unexpected consequences:
- After the first run only original tags are visible in the sidebar
- At opening the task with localized tags GTG finds, that there are some missing tags in task description (defined in `/GTG/core/firstrun_tasks.py`) and simply adds them there
- At this moment it also adds localized tags (which it finds in the task description) to the sidebar

kocio-pl:
@Neui What do you think about that as a translator?

kocio-pl:
Proper solution for this issue should be to drop static tags initialization with first-time user guide and collect them from target (translated) tasks content. 

However that will immediately trigger #807 - which means adding to the sidebar unexpected `@money.` tag along with standard `@money` tag (or their translated equivalents).

PR #852, opened 2022-03-06
Dynamic tags in the first run

kocio-pl:
Fixes https://github.com/getting-things-gnome/gtg/issues/803

Replaces initial static tags with dynamic ones, collected from localized task content. This allows proper tags translation.

Since `&quot;@waitingfor&quot;` string is not matched by `extract_tags_from_text()`, `&quot;` part has been dropped. Because of that regenerating POT file (and translations) is needed.


diegogangl:
LGTM, thanks!

PR #853, opened 2022-03-06
Open tag panel by default

kocio-pl:
Left panel seems to be important tool for managing tags, so I think it should be open by default. 

It's especially important for the first time users to see it. Opening panel is described in the initial guiding tasks, but let's make it as easy as possible to discover this important feature, even without getting through the guide.

Current initial state looks like this:
Zrzut ekranu z 2022-03-06 11-49-44


diegogangl:
I personally agree, but let's see what @nekohayo think about this

diegogangl:
Thanks!

Issue #851, opened 2022-03-05
[Untouched tasks] plugin not initializing modified time from data file


kocio-pl:
It seems to me that this plugin does not work. I tested it with 0.6 flatpak version making just 1 day update period for `@untouched` tag, but the tag is not appearing in automatic mode, nor after clicking this item in menu.

Could anyone confirm it?

kocio-pl:
It looks like the problem is wrong modified time - it's evaluated as "now" (today) while starting GTG, instead of reading it from the data file. Modified time then gets updated every time you open it and stays the same if you don't (which is expected), but this means that you have to have GTG running for a month to see the plugin really working.

PR #856, opened 2022-03-07
Keep modified date when reading from file

kocio-pl:
Fixes https://github.com/getting-things-gnome/gtg/issues/851

When reading from file, there are multiple actions which are changing task modification date (like `set_date()` with other keys or `set_recurring()`). Therefore setting modification time needs to be the last action to not be overridden with current time. We don't want to treat mere initialization process as real task modification.

It affects untouched tasks plugin in the obvious way - it gets current time as task modification time, so it can't recognize forgotten ones. It may also affect other actions relying on modification time (maybe synchronization).


diegogangl:
Thanks! Sorry for the delay

PR #857, opened 2022-03-07
Fixing @untouched tag problems

kocio-pl:
Untouched tasks plugin adds `@untouched` tag with `@` included in the name, which confuses GTG - each task opening adds another copy of this tag. Skipping this character fixes the problem.


diegogangl:
A gift from the days of gtg sometimes adding an @, sometimes not. Merged, thanks!

Issue #825, opened 2022-02-27
Hide special tags


kocio-pl:
I'm not sure what is the use of showing special tags in tag list browser, I guess they should be hidden here:

![Zrzut ekranu z 2022-02-27 08-00-18](https://user-images.githubusercontent.com/5439713/155872035-2ca55015-597a-482f-b37b-d1cb91ef2b8b.png)

diegogangl:
Definitely a bug

kocio-pl:
It's good to remember to also hide saved searches, which look like `@__SAVED_SEARCH_$name` tag.

PR #858, opened 2022-03-08
Hide special tags in task editor popup

kocio-pl:
Fixes https://github.com/getting-things-gnome/gtg/issues/825

This patch hides all the special tags (including saved searches) from the tag popup in task editor.


diegogangl:
Works as advertised, thanks!

PR #860, opened 2022-03-09
Fixed plural forms for Slovak language

rzelnik:
I noticed that my GTG app shows mismatched dates in a Due column, so I fixed it according to the Czech translation, which has equal grammar rules for plural forms.


rzelnik:
I added the commit to the master branch by mistake. I created a new branch and will create a new pull request.

PR #861, opened 2022-03-09
Fixed plural forms for Slovak language

rzelnik:
I noticed that my GTG app shows mismatched dates in a Due column, so I fixed it according to the Czech translation, which has equal grammar rules for plural forms.


diegogangl:
I don't know much about slovak, but I'm guessing this is correct. Thanks!

Issue #684, opened 2021-06-15
When a tag is selected, prefill each new Task with the corresponding tag


RoneoOrg:
Hi,

when I'm working on a specific project and browsing the corresponding Tag, I would love to see this tag automatically added to each Task I create (with `Ctrl+N` in my situation)

It used to be the case in previous GTG releases and was really handy.

Now, when I create a new Task, I need to type the whole Tag name manually. If this tag is forgotten, the Task is lost in the middle of every other Tasks, and I need to remember the name of this Task, search for it, and finally add the Tag 😓

nekohayo:
I agree, it's a somewhat annoying minor regression in terms of UX, either because it was lost sometime in the bunch of code changes from previous years, or maybe as a way to work around some issue like #635 or maybe some other (somehow can't find it in [there](https://github.com/getting-things-gnome/gtg/issues?q=is%3Aissue+tag+new+task+)...). Will probably be dependent on the "big core rewrite" project (issue #553).

PR #862, opened 2022-03-10
Prefill new task with selected tags

kocio-pl:
Fixes https://github.com/getting-things-gnome/gtg/issues/684

Removes unneeded condition line that checks if tag name includes `@` character. Actual tag names are chains that follow this character, it only acts as an indicator that the tag name starts here.


diegogangl:
Great fix, thanks!

Issue #859, opened 2022-03-08
"Someday" should be sorted at the end (after tasks with no duedate) when sorting by end date


Gert-dev:
Hi

Thanks for maintaining this application. I've been using it with satisfaction for quite some time, and wanted to ask a question - possibly suggest an improvement - around the ordering of tasks when sorting by the end date column.

Currently, when sorting tasks in an ascending fashion by end date, you see:

1. Task A that ended five days ago.
2. Task B that needs to be done tomorrow
3. Task C that needs to be done at date Z in the future, but within the current year.
4. Task D with "someday" as end date.
5. Task E that needs to be done at date Y in the future, but not this year.
6. Task F with no end date set.

For example:

![afbeelding](https://user-images.githubusercontent.com/24809863/157307934-c802df8b-4978-40ce-baad-9ca7df7f1fdb.png)

Two things strike me as somewhat odd here: one is that apparently "someday" is seen as "somewhere within the current year", whilst I thought the purpose of it was "it needs to be done at some point, but it has no required end date, so don't show it in the Open view either", implying low priority - in the screenshot, the task due by 2023 is sorted after "someday" tasks, though.

The other thing is that I'm wondering if it would make more sense to always sort "someday" tasks at the far bottom, and sort the tasks with no end date before them. The reasoning is that tasks without end date might still need to be done in the near future ("soonish"), whilst for "someday" tasks priority doesn't matter - in a sense, tasks with "someday" have an end date that is later than tasks without an end date. Currently, if you have a lot of tasks "someday", they are all sorted before tasks that might be more "urgent", pushing them down.

There may be a reasoning for the current sorting, though, and maybe it just comes down to preference.

Thanks!

Neui:
About the current sorting, it seems for "someday" for comparing dates it is turned into a date where it is a year from now (aka today):

https://github.com/getting-things-gnome/gtg/blob/4dd1648572aa7d5eb3c4fdd370d08ad4a9610ab2/GTG/core/dates.py#L201-L203

nekohayo:
I wholeheartedly agree that this sorting doesn't make sense. This is just a wild guess, but this might historically have something to do with how sorting the GtkTreeview works, and if so, it might be something for @diegogangl and @jaesivsm to consider as part of #737 / New Core / GTK4 / new views ... so many moving pieces interacting with each other there!

[...]

PR #863, opened 2022-03-10
Sorting Actionable from most urgent due dates

kocio-pl:
Related to https://github.com/getting-things-gnome/gtg/issues/859

Current sorting in Actionable view starts with no date tasks and most urgent dates go to the end, which means `today` might not be even visible. This simple patch reverses the sorting order of due date, so it starts with `today` tasks.

### DIFF
--- a/GTG/gtk/browser/treeview_factory.py
+++ b/GTG/gtk/browser/treeview_factory.py
@@ -163,7 +163,7 @@ class TreeviewFactory():
             t1 = task1.get_due_date_constraint()
         if t2 == Date.no_date():
             t2 = task2.get_due_date_constraint()
-        return self.__date_comp_continue(task1, task2, order, t1, t2)
+        return self.__date_comp_continue(task1, task2, order, t2, t1)
 
     def sort_by_closeddate(self, task1, task2, order):
         t1 = task1.get_closed_date()


nekohayo:
Users can click twice on the column header to sort ascending vs descending, I'm not sure this code change here fixes the issue, and possibly introduces other issues?

kocio-pl:
1. I think it does fix the issue - this is the current state:

Zrzut ekranu z 2022-03-11 04-48-06

and this is how it looks with urgent tasks on top - it's hard to miss them :smile: ... :

Zrzut ekranu z 2022-03-11 04-51-50

2. It's actually just one click, but proper UX is (among other things) about sane defaults. Also the order is not preserved between changing views and restarts, so you have to click it every time.

3. I have double checked that `sort_by_duedate()` is called only by `active_tasks_treeview()`, so no side effects are expected here. Additional question I would ask just to make sure: what are the scenarios in which one would like to see less important tasks before urgent ones?

primal-buddhist:
I like this and would welcome it as a default. As it is it doesn't even remember that I have clicked to re-sort.

terceiro:
This is really useful; I'm a new gtg user and I always get confused when I click "Actionable" and don't see the most urgent tasks at the top.

kocio-pl:
@nekohayo Hi again, what do you think now? I still think this is very useful change.

nekohayo:
Hi @kocio-pl, the big refactoring and GTK4 port has finally landed (see the pinned 0.7 tracking issue for details).

I am under the impression that this implementation here is going to be obsolete even if you try to rebase it. Please check how `master` behaves for you now and see whether this pull request is still relevant or if it has been made obsolete.

kocio-pl:
You are right, this code is not relevant any more, so I will close it.

I still think that default sorting for Actionable should be ascending due date, but it needs to be redone.

nekohayo:
You might be interested in #1020 and #1021 that affect the current development version, for starters ;)

Issue #794, opened 2022-02-18
Day change for the Actionable view is hardcoded to 5 AM


kocio-pl:
It looks to me like day change has a problem in Actionable view. It's hard to observe such infrequently reproducible thing, but from my experience GTG with default settings (day change at 00:00) does not show "today" tasks there for a few hours after midnight. Actual change takes place probably at 5:00 AM (my local time is CET, so it's 4:00 AM UTC). 

It's also even more complicated. Before that hour I can not do anything to show "today" tasks, but even after that I need to do something, Actionable view does not refresh automatically. Changing views between Open, Actionable and Closed also does not help, nor restarting the app, but it seems that moving across tags does the trick.

This behavior is consistent on both 0.5 flathub version and a current code from a github.

Is this problem known already or somebody could confirm that or prove me wrong?

kocio-pl:
Well, it's a deliberate choice!

https://github.com/getting-things-gnome/gtg/blob/54b20ba6958b0fe67228cb83c84b6bb783c432a8/GTG/core/treefactory.py#L176-L177

I think this is wrong - if someone wants "today" to start later than on midnight, we have a configuration option exactly for that. It's highly confusing to me to wait additional few hours before showing current tasks. I don't see what was the reasoning for this tweak, but I would remove it.

nekohayo:
Congratulations, you have found some really, really ancient hardcoded stuff :D this is from over 12 years ago, before commit 18d28b4b9 (I couldn't figure out exactly what commit it was introduced in before that), and it is possible/likely that it predated the configurable setting for the virtual midnight, that it just forgotten there. A question though is how much of this (if any at all) would survive Diego's new core implementation for 0.7 anyway...

kocio-pl:
I'd like to prepare a patch for that, but I'm not a programmer. Could anybody tell me what exactly to do to access configuration options (namely `hour` and `minute`) from `GTG/core/treefactory.py`?

PR #867, opened 2022-03-15
Start using day change from config for Actionble

kocio-pl:
Fixes https://github.com/getting-things-gnome/gtg/issues/794

Replaces old hardcoded day change for Actionble tasks (5 AM) with a time shift defined by user in the configuration (`hour`:`min`).


diegogangl:
Thanks!

Issue #645, opened 2021-04-23
[RFC] Refurbishing Recurring/Repeating tasks


zeddo123:
We have received many issues and in hindsight I took short sided decisions about recurring tasks. That's why I opening this issue so that we can discuss how we might proceed.

### New features
* Ability to repeat every x weeks/month/days #627
* Ability to have multiple recurring tasks #614
* Ability to select whether the task repeats on due dates or start dates

For these features, we could use dateutil proposed by @jaesivsm and @diegogangl.

### History and duplicate tasks
This issue was raised here #636. The best behavior (if we want to preserve the "history") would be that then the task is re-opened, the new tasks gets deleted and replaced by the old one. Otherwise, we just update the due date.

### Changing the name
I don't even know why I first chose recurring tasks instead of repeating. That's why I'm proposing to change the naming. The only issue with this is that some users might have gotten used to it.

I had an unexpected busy and stressful week, so I couldn't discuss it on IRC or even label issues. Now hopefully I can.

jaesivsm:
I haven't had time to touch what I started recently (busy week also) but I can open a draft PR with my work in progress. 

I'm more of a backend guy than anything. So, unless there is some amazing designer hidden somewhere I want to underline that  the recursion rule RFC ([somewhere in there](https://tools.ietf.org/html/rfc5545#section-3.3.10)) is very (and I mean *very*) flexible. I don't think we will (at least not on the first draft) have a UI that'll enable every possible combination. I don't think it's needed also.

### What we have : 
 * every day (iCalendar : `DAILY`)
 * other day (iCalendar `DAILY, INVERTVAL=2`)
 * every week (iCalendar `WEEKLY,BYWEEKDAY=(<str>)`)
 * every month (iCalendar `MONTHLY,BYDAY=(<int>)`)
 * every year (iCalendar `YEARLY,BYMONTH=(<int>),BYDAY=(<int>)`)

### What's easy to add
 * Support for whatever interval : turn `other-day` into `every {2} {days}` by making interval a parameter (we then could have for example `every {3} {month}` => quarterly)
 * Multiple recursion : weekly, multiple day => rrule basics + checkbox instead of select on weekdays
 * Adding dtstart selection seems overkill and counter intuitive (see my proposal in the last section of this post)

### What'll be impossible
[...]

PR #868, opened 2022-03-16
New Repeating tasks

zeddo123:
*I'll fill this up later*

### DIFF
--- a/GTG/core/meson.build
+++ b/GTG/core/meson.build
@@ -47,6 +47,7 @@ gtg_core_sources = [
   'base_store.py',
   'tasks2.py',
   'tags2.py',
+  'repeating.py',
   'saved_searches.py',
   'datastore2.py',
 ]
--- a/GTG/core/repeating.py
+++ b/GTG/core/repeating.py
@@ -1,0 +1,51 @@
+from datetime import datetime
+from enum import Enum
+
+
+class RepeatingOn(Enum):
+    DUE, START, BOTH = list(range(3))
+
+
+class Repeating:
+    def __init__(self, enabled=False, ruleset=None, repeats_on=RepeatingOn.DUE, count=0, old_tid=None):
+        self.enabled = enabled
+
+        # dateutil ruleset
+        self.rset = ruleset
+        self.repeats_on = repeats_on
+
+        self.old_tid = old_tid
+        self.count = count
+
+        self.timestamp = datetime.now()
+
+
+    def set_ruleset(ruleset):
+        self.rset = ruleset
+        self.update_date()
+
+
+    @property
+    def repeats_on_start(self):
+        return True if self.repeats_on in (RepeatingOn.START, RepeatingOn.BOTH) else False
+
+
+    @property
+    def repeats_on_due(self):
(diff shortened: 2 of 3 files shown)


diegogangl:
oooohhhh I like where this is going!

zeddo123:
@diegogangl I'm trying to write unittests along side the new code, but I'm having issues with importing the datastore2. However this does not happen in the dev console.

```python
>>> from GTG.core.datastore2 import Datastore2
  File "<stdin>", line 1, in <module>
  File "/home/zeddo/Documents/github.com/zeddo123/gtg/GTG/core/datastore2.py", line 37, in <module>
    import GTG.core.info as info
ModuleNotFoundError: No module named 'GTG.core.info'
>>>
```

diegogangl:
:P I just realized I didn't write _any_ tests for the datastore. The info file has some stuff that is filled in by Meson, so it's not available for tests unless you build (check out `core/info.py.in`. It kind of sucks, but one way we can do it is by mocking it in `datastore2.py`.

```python
try:
    # Running the app
    import GTG.core.info as info
except ImportError:
    # Unit tests
    import mock
    info = mock.Mock()
    info.VERSION = 'v0.6'
```

zeddo123:
@diegogangl For what I have understood from the new core is that we no longer have a requester but a datastore that has a Taskstore. Will that tasks have a "pointer" to the datastore as it was the case with the previous implementation?
Before we were able to do, inside the task, `self.req.get_task()`. What is the alternative now?

From the point of view of repeating tasks, I need to get a certain number of tasks to get the history of occurrences.

diegogangl:
> From the point of view of repeating tasks, I need to get a certain number of tasks to get the history of occurrences.

hmm, I hadn't considered getting other arbitrary tasks from inside a task. I guess the only solution for now is to pass the taskstore to each task.

diegogangl:
There's also another option, which would be to have each task emit a signal when the status changes. Then we can connect the signal of each task to a handler in task store and do the repeating check in there. Don't know about the performance of this though

zeddo123:
> There's also another option, which would be to have each task emit a signal when the status changes. Then we can connect the signal of each task to a handler in task store and do the repeating check in there. Don't know about the performance of this though

If the performance hit isn't catastrophic, this is definitely a better solution.

pentux-GitHub:
Good news looking at this fresh: the recurrence work here actually made it into the new core. Current `tasks.py` has `set_recurring`, `toggle_recurring`, `inherit_recursion`, `get_next_occurrence` and `duplicate_for_recurrent`, and `dates.py` parses recurring terms, so the internal repeating feature is live on master.

What predates the port is the file layout (`datastore2.py`, `tasks2.py`, `repeating.py`), so this branch itself isn't rebasable as-is, but the feature it pioneered isn't lost. The remaining gap is CalDAV: the RRULE mapping in the backend is still commented out, so repeating tasks don't sync yet. That's tracked by the #645 RFC, which already sketches the term to RRULE table.

@zeddo123 thanks for kicking this off, it's live in the core now. I'd think this can be closed in favour of #645 for the CalDAV side.

Issue #821, opened 2022-02-27
E-mail plugin problems


kocio-pl:
When trying to send a parent task from initial run via e-mail, I get:

```
Subject: Task: Getting Started with GTG (read me first)

Status: Active
Tags:
Subtasks:
Learn How to Use Subtasks
- Learn How to Use Tags and Enable the Sidebar
- Learn How to Use the Actionable View Mode
- Learn About Plugins
- Reporting Bugs
- Learn How to Use the Quick Add Entry
- Learn About Synchronization Services
- Learn How to Search for Tasks
Task content:
```

Specific problems I see:

1. `Learn How to Use Subtasks` does not start with a `-` for some reason
2. Task content is missing
3.  Also when I try to run it in localized environment, task status (Active, Done, Dismiss) is not translated

kocio-pl:
After https://github.com/getting-things-gnome/gtg/pull/823 it's better, because main problem 2. is gone, but `get_text()` shows tids (task ids) in the content, which makes no sense probably. However it's translated at least:

```
Jeśli chcesz się dowiedzieć więcej o innych funkcjach GTG, to przejdź do tych tematów:
{! 262d1410-71aa-4e35-abec-90ef1bab44d3 !}
{! 2d427a77-3077-4277-904c-073fcfcb4842 !}
{! 0653765c-f5d4-4b7f-9903-0065bb258940 !}
{! 3efd01b7-343d-4be1-9bac-8ed944517e3f !}
{! 9b05a6c5-81e3-4fe5-a70d-3b569f65409c !}
{! 21ed54a8-e7ff-408a-b648-8d12cc75e162 !}
{! d1c73224-9eec-4889-b3c4-bb0281e5c1d3 !}
{! 586dd1a7-0772-4d0a-85db-34edfc8ee30c !}
```

So we need some wrapper to convert them to task names - maybe `get_excerpt()` with some options would do it? But then it's unclear to me if a separate list of subtasks is needed. If someone gets the role of this plugin better, please share, so we can think about proper solution.

PR #869, opened 2022-03-17
Correct subtasks formatting in e-mail plugin

kocio-pl:
Fixes part of https://github.com/getting-things-gnome/gtg/issues/821

Subtask list formatting:
- if there are any tasks, put a dash character before each of them - instead of skipping it for the first subtask
- if there are none, print nothing - instead of an empty line

Translation regeneration required.


diegogangl:
Thanks! Sorry for the delay

Issue #835, opened 2022-03-02
Improve the tag deletion warning dialog's message to make it clear it won't destroy tasks


nekohayo:
This is a super easy fix, but not doing it until we have released 0.6, as it would break translations: GTG/gtk/browser/delete_tag.py has a pair of strings that say:

> Deleting a tag cannot be undone, and will delete the following tag: 

Plural:

> Deleting a tag cannot be undone, and will delete the following tag: 

...it would make sense to rework those strings to explicitly reassure the user that while this perma-deletes the tag and removes it from all the tasks it was in, it will not actually delete the tasks that were associated with that tag.

Also, check if it's better to have a %s for the tag name to be positionable and quotable inside the translatable string... i.e., will there really be cases where the user can delete multiple tags at once (rather than one-by-one, as is currently the case) in practice?

martroxxx:
@nekohayo please review. I am pretty sure, that there is no real usage scenario for deleting multiple tags. Just assuming from my own workflow

PR #871, opened 2022-03-23
GTK/gtk/browser/delete_tag.py: Small language improvement

martroxxx:
Makes clear that only the tag is deleted, no tasks

Closes https://github.com/getting-things-gnome/gtg/issues/835


diegogangl:
LGTM, let's see what @nekohayo thinks

PR #873, opened 2022-03-27
Add Snap support

### LINKED ISSUE #872: Snap support
I would like to propose support for Snap, similar to Flatpak. I intend to create the Snap in a couple of days, and I will submit a patch to include the code in the gtg main repository.

mhalano:
Add Snap support

The Snap uses metadata from AppData, so it always gets the latest version, summary, description and so on, so no need to edit the file each new release.

The version of liblarch is hard-coded.

To generate the .snap file, you just have to clone the repository with my PR in it and execute `snapcraft` command in the project root directory.

The last remain question about the support is how to publish in the Snap Store in an official way, similarly to how Flatpak is published.

Please help me to test. I don't have a CalDAV server to test CalDAV backend, for instance. I allowed network communication so I think it will be enough.

Closes #872

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,58 @@
+name: getting-things-gnome
+base: core20
+compression: lzo
+grade: stable
+license: GPL-3.0
+adopt-info: gtg
+confinement: strict
+
+apps:
+  gtg:
+    command: usr/bin/gtg
+    extensions: [gnome-3-38]
+    environment:
+      PYTHONPATH: $SNAP/usr/lib/python3/dist-packages:$PYTHONPATH
+    common-id: org.gnome.GTG
+    plugs: [network]
+    slots: [dbus-service]
+ 
+slots:
+  dbus-service:
+    interface: dbus
+    name: org.gnome.GTG
+    bus: session
+
+parts:
+  gtg:
+    source: .
+    plugin: meson
+    build-environment: [PYTHONPATH: '']
+    build-packages: [gettext, itstool]
+    meson-parameters: [--prefix=/usr]
+    parse-info: [usr/share/metainfo/org.gnome.GTG.appdata.xml]
+    stage-packages:
+      - python3-caldav
+      - python3-dateutil
+      - python3-dbus
+      - python3-lxml
+      - python3-vobject
+    stage:
+        - usr/bin/gtg
+        - usr/lib/python3/
+        - usr/share/
+        - -usr/share/doc/
+        - -usr/share/lintian/
+        - -usr/share/man/
+
+  liblarch:
+    source: https://github.com/getting-things-gnome/liblarch
+    source-tag: v3.2.0
+    source-type: git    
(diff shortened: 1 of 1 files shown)


mhalano:
I requested the 'gtg' name in the Snap store, so I can upload all new releases. If you want to use an official account (like getting-things-gnome) or something like that, please let me know.

diegogangl:
Hi, thanks for the work. The only thing that worries me about having this in the repo is that we have a policy of only supporting flatpak officially. Will having the snap recipe here give snap users the wrong impression?

Also AFAIK none of us use ubuntu, so if something about this breaks it will have to wait until someone motivated fixes it. Will you be using this snap version?

> I requested the 'gtg' name in the Snap store, so I can upload all new releases. If you want to use an official account (like getting-things-gnome) or something like that, please let me know.

I think we settled for the name gtg (no acronym).

mhalano:
You can describe the support for Snap as unofficial or better, experimental. Could put into a contrib folder so people know it's a contribution. I can manage the Snap version. Yes I will be using, but like I said, I don't have some features like CalDAV backend. I would be uncovered on that. You can install snapd package on Fedora and use Snaps, if you want it. I created because Debian/Ubuntu doesn't have gtg package anymore (the last one it was 0.3 series).

diegogangl:
> You can describe the support for Snap as unofficial or better, experimental. 
Awesome

Hit me on the Matrix channel or send me an email at diego@sinestesia.co, and I'll send you some credentials for a caldav test server

mhalano:
One question I did to myself and elaborate an answer. Maybe would be better to save the data inside home directory directly, not under the subfolder `~/snap ` (used to store userdata for Snap applications). The way I didn't do that is because even there is a way to do it, needs to enable manually interface connection. I prefer to keep the userdata in `~/snap` instead.
There is a possibility to request an auto-connection for the interface, but I think it's an improvement for the future.

mhalano:
I published the Snap: https://snapcraft.io/getting-things-gnome. As I mention in the description, this is not an official Snap and I'm working with you pals to change that.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+apps:
+  gtg:
+    command: usr/bin/gtg
+    extensions: [gnome-3-38]
+    environment:
+      PYTHONPATH: $SNAP/usr/lib/python3/dist-packages:$PYTHONPATH
+    common-id: org.gnome.GTG
+    plugs: [network]
+
```
sergiusens, line 17:
```suggestion
slots:
    session-dbus-interface:
        interface: dbus
        name: org.gnome.GTG
        bus: session

```

mhalano: I got it... Because the integration with Hamster is made using D-BUS...

mhalano:
@sergiusens implemented your change and already published in the Snap Store.

diegogangl:
LGTM, thanks for the work!

PR #875, opened 2022-03-31
Fix collaborative calendar tags

FooSensei:
Since CalDav sync is still very wonky and in fact deleted all of my tasks (luckily I managed to restore all of them) I decided to only connect GTG to shared calendar and log with separate account (very easy to set up in Nextcloud).

The problem that I noticed was, that for shared calendars the identifying tag would be of format `@DAV_NAME_(user)` and with every sync the part `(user)` would be multiplied to the next line effecting in absolute madness of a spam at the top of every task.

This should fix the issue.


diegogangl:
Sorry to hear about the tasks. Thanks for the fix!

PR #886, opened 2022-04-24
Preparing for Gtk4 - Part 2

diegogangl:
Includes cherry picks from @ranchester2 's Gtk4 branch, updated to the latest changes in master. This PR leaves us "at the edge" of moving to Gtk4.

### DIFF
--- a/GTG/core/plugins/api.py
+++ b/GTG/core/plugins/api.py
@@ -19,6 +19,7 @@
 import os
 import pickle
 import logging
+from gi.repository import GLib
 from GTG.core.dirs import plugin_configuration_dir
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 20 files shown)


diegogangl:
@Neui or @ranchester2  you around? Would be neat if you could take a quick look

GTG/core/plugins/api.py
```diff
@@ -146,21 +147,42 @@ def remove_active_selection_changed_callback(self, plugin_class):
...
+        # when adding a new one. A reliable solution is to instead find the first one
+        # with the same label as the given one.
+        _, _, menu = self.__builder.get_object(
+            'editor_menu' if self.is_editor() else 'main_menu'
+        # all menu items are added to the first section
+        ).iterate_item_links(0).get_next()
+
+        length = menu.get_n_items()
+        i = 0
+        while i < length:
```
Neui, line 172:
```python
for i in range(menu.get_n_items()):
    ...
```
?

GTG/core/plugins/api.py
```diff
@@ -146,21 +147,42 @@ def remove_active_selection_changed_callback(self, plugin_class):
...
 
-        @param item: The Gtk.MenuItem that is going to be removed.
+        @param item: The Gio.MenuItem that is going to be removed.
         """
-
-        menu_box = self.__builder.get_object('menu_box')
-        menu_box.remove(item)
+        # you cannot remove items by identity since there values are simply copied
+        # when adding a new one. A reliable solution is to instead find the first one
+        # with the same label as the given one.
```
Neui, line 164: Maybe for the future, require plugins to insert some kind of UUID as an [attribute](https://lazka.github.io/pgi-docs/#Gio-2.0/classes/MenuItem.html#Gio.MenuItem.set_attribute) to work from? So then we don't have to do some label matching. (So maybe just put an `# TODO` here or so about this)

Edit: Or maybe also check action (if set)?

GTG/gtk/editor/editor.py
```diff
@@ -331,6 +331,35 @@ def open_tags_popover(self):
...
+        # causes a crash and a weird reference count assertion error. Retrieving it from
+        # the main menu instead avoids the cryptic crash
+        __, __, editor_menu_con_sec = self.editormenu.iterate_item_links(0).get_next()
+
+        # Plugins may add menu items to here, which is why we need to only remove
+        # the specific item that we need instead of everything
+        # Manual iteration without GI overrides is ugly
+        length = editor_menu_con_sec.get_n_items()
+        i = 0
+        while i < length:
```
Neui, line 350: Like above
```python
for i in range(editor_menu_con_sec.get_n_items()):
    ...
```

Neui (commented):
I only read the changes and mostly skipped the `.ui` files. I haven't tested it yet actually.

Issue #877, opened 2022-04-03
GTG Unable to Start After Recent Install


martinlastname:
I just installed GTG via Flatpak, did some poking around, updated my system, and now can no longer launch even after reinstalling and rebooting. What exactly I did (to the best of my memory from 10m ago):

1. Installed GTG
2. Ran `flatpak update` and `yay -Syu` in separate terminals
3. Opened GTG
4. Read some of the tutorial tasks
5. Opened the synchronization menu and started adding a CalDAV sync (stopped at the login screen)
6. flatpak and yay finished updating
7. I got a notification that core system packages had been updated and warranted a reboot
8. Rebooted
9. GTG crashes on startup with the following error

> **Context:** Startup
> 
> ```python-traceback
> Traceback (most recent call last):
>   File "/app/lib/python3.9/site-packages/GTG/gtk/application.py", line 110, in do_startup
>     for backend_dic in BackendFactory().get_saved_backends_list():
>   File "/app/lib/python3.9/site-packages/GTG/backends/__init__.py", line 143, in get_saved_backends_list
>     module = self.get_backend(settings.get('module'))
>   File "/app/lib/python3.9/site-packages/GTG/core/config.py", line 164, in get
>     raise ValueError(
> ValueError: No valid configuration value or default value was found for %s in %s
> ```
> 
> **Software versions:**
> * Getting Things GNOME! v0.6
> * CPython 3.9.9 (main, Nov 10 2011, 15:00:00)   [GCC 11.2.0]
> * GTK 3.24.33, GLib 2.70.5
> * PyGLib 3.42.0, PyGObject 3.42.0
> * Linux-5.17.1-arch1-1-x86_64-with-glibc2.33

I tried rebooting again, didn't fix it. I tried reinstalling via flatpak, same error.

If I try to launch GTG again without closing the original error dialogue, I get this error instead:

> **Context:** Global generic exception
> 
> ```python-traceback
> Traceback (most recent call last):
>   File "/app/lib/python3.9/site-packages/GTG/gtk/application.py", line 144, in do_activate
>     if self._check_exception():
>   File "/app/lib/python3.9/site-packages/GTG/gtk/application.py", line 195, in _check_exception
>     GLib.Source.remove(self._exception_dialog_timeout_id)
> TypeError: Argument 0 does not allow None as a value
> ```
> 
> **Software versions:**
> * Getting Things GNOME! v0.6
[...]

PR #889, opened 2022-05-05
Uniformizing error message and error message formatting

jaesivsm:
After launching GTG from flatpak I got an ill formatted error (showing `%s` instead of the proper values). It's due to a `raise ValueError` in `GTG/core/config.py`. After looking at that I started just fixing and uniformizing the error message formatting across GTG. Here's what's in the PR : 
* uniformizing error message formating :
  * using the same string formatting mechanism
* raise `NotImplementedError` not `NotImplemented` which isn't an exception
```
>>> raise NotImplemented
  File "<stdin>", line 1, in <module>
TypeError: exceptions must derive from BaseException
```
* do not explicitly reraise current error in an `except:` block, juste use bare `raise` statement
* raise error objects, not classes

Edit: encountered same error than #877

### DIFF
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -224,7 +224,7 @@ class Backend(PeriodicImportBackend):
             BackendSignals().interaction_requested(
                 self.get_id(), f"{message} {error!r}",
                 BackendSignals().INTERACTION_INFORM, "on_continue_clicked")
-            raise error
+            raise
         for calendar in principal.calendars():
             self._cache.set_calendar(calendar)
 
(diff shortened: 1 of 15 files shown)


Neui (approved):
Haven't explicitly tested, but seems fine to me.

pentux-GitHub:
Most of this has been overtaken by the port: the files it touches (`datastore2.py`, `tasks2.py`, `task.py`) no longer exist, and the main fix, `raise NotImplemented` → `NotImplementedError`, is already gone from master.

Thanks @jaesivsm for the cleanup. Given the target files are gone, a rebase isn't really possible; if any error-formatting inconsistencies remain, they'd be better as a fresh pass on the current code. Otherwise I'd think this is safe to close.

PR #897, opened 2022-05-15
Improve Synchronization UI

### LINKED ISSUE #837: Synchronization backends dialog doesn't look nice when there are no truly available sync backends to add
The default state of the "Synchronization" (backends) dialog looks like this:

Screenshot from 2022-03-02 12-35-19

First, the "Local File" backend's label and button are totally wrong there. It's not a synchronization service, it's the default file storage backend, so the label should reflect that, and the "Disable syncing" button should be hidden rather than disabled in that case.

If you want the UI to be more helpful, the label could instead explain where the currently used XML data file is located on the filesystem, and a button (or hyperlink?) to `xdg-open` (or whatever is the proper way in the flatpak portal-enabled world these days) the file browser to its containing location.

Then, if you don't have "python-caldav" installed/available (as is the case on Fedora, for example), if you click the "+" button you get this UI:

Screenshot from 2022-03-02 12-35-22

* That combobox is totally useless when it has zero items in it, and so are the labels and buttons shown there. Instead, it would make much more sense to at least show a human-readable explanation that details which dependencies are missing to allow which backends to be available (but then there's also the question of what to do when only some of the backends are unavailable instead of all of them, if we have more than one sync backend in the future).

runkaiz:
Partially implements #837.

This is my first pull-request (well, first time attempting to contribute to a public open source project in fact). Code changes are minimum so please give me some advice if I am doing something wrong here, any help would be appreciated, just trying to gain some experience.

### DIFF
--- a/GTG/backends/__init__.py
+++ b/GTG/backends/__init__.py
@@ -54,6 +54,7 @@ class BackendFactory(Borg):
             # This object has already been constructed
             return
         self.backend_modules = {}
+        self.inactive_modules = {}
         backend_files = self._find_backend_files()
         # Create module names
         module_names = [f.replace(".py", "") for f in backend_files]
@@ -67,10 +68,12 @@ class BackendFactory(Borg):
                 # Something is wrong with this backend, skipping
                 log.warning("Backend %s could not be loaded: %r",
                             module_name, exception)
+                self.inactive_modules[module_name] = str(exception)
                 continue
-            except Exception:
+            except Exception as exception:
                 # Other exception log as errors
                 log.exception("Malformated backend %s:", module_name)
+                self.inactive_modules[module_name] = str(exception)
                 continue
 
         def browse_subclasses(cls):
@@ -110,6 +113,13 @@ class BackendFactory(Borg):
         Returns a dictionary containing all the backends types
         """
         return self.backend_modules
+
+    def get_inactive_modules(self):
+        """
+        Returns a dictionary mapping the name of each backend module that
+        failed to load to the error message explaining why it failed
+        """
+        return self.inactive_modules
 
     def get_new_backend_dict(self, backend_name, additional_parameters={}):
         """
--- a/GTG/gtk/backends/addpanel.py
+++ b/GTG/gtk/backends/addpanel.py
@@ -18,7 +18,7 @@
 
 from functools import reduce
 
-from gi.repository import Gtk
+from gi.repository import Gtk, GLib
 
 from GTG.gtk.backends.backendscombo import BackendsCombo
 from GTG.backends import BackendFactory
@@ -71,15 +71,15 @@ class AddPanel(Gtk.Box):
 
         @param box: the Gtk.Box to fill
         """
-        label = Gtk.Label(label=_("Select synchronization service:"))
-        label.set_xalign(0)
-        label.set_yalign(0.5)
+        self.select_label = Gtk.Label(label=_("Select synchronization service:"))
+        self.select_label.set_xalign(0)
+        self.select_label.set_yalign(0.5)
         self.combo_types = BackendsCombo(self.dialog)
         # FIXME
         # self.combo_types.get_child().connect(
         #     'changed', self.on_combo_changed)
         self.combo_types.connect('changed', self.on_combo_changed)
-        box.append(label)
+        box.append(self.select_label)
         box.append(self.combo_types)
 
     def _fill_middle_box(self, box):
@@ -135,6 +135,9 @@ class AddPanel(Gtk.Box):
     def refresh_backends(self):
         """Populates the combo box containing the available backends"""
         self.combo_types.refresh()
+        # An empty combo box emits no 'changed' signal, so refresh the
+        # description explicitly to handle the no-backends case
+        self.on_combo_changed()
 
     def on_confirm(self, widget=None):
         """
@@ -166,7 +169,12 @@ class AddPanel(Gtk.Box):
         """
         backend_name = self.combo_types.get_selected()
         if backend_name is None:
+            self.on_no_backend_available()
             return
+        self.select_label.set_visible(True)
+        self.combo_types.set_visible(True)
+        self.ok_button.set_sensitive(True)
+        self.label_author.set_visible(True)
         backend = BackendFactory().get_backend(backend_name)
         self.label_description.set_markup(backend.Backend.get_description())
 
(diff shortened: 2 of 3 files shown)


Neui:
Not entirely sure if you want some feedback now, but:
* You hide the button, but never make it `.show()` it again (at least from what I can see from the changes, I haven't tested the change). There is [`.set_visible()`](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Widget.html#Gtk.Widget.set_visible) you can use with an boolean parameter. if you want to make it a one-liner.
* `print()` is OK when debugging, just make sure to remove them before committing/submitting. (I'd say in a draft PR it would be OK, since code is likely to be changed anyway). You could also use the logging system to print stuff out like `log.debug("xyz")` if you think it could be useful later when debugging (you need to `import logging` and `log = logging.getLogger(__name__)` near the top of the file, like for example https://github.com/getting-things-gnome/gtg/blob/5f72c8a9319e06d42967fb8461ea832e8cd095d9/GTG/gtk/browser/adaptive_button.py#L1-L8

For the link stuff (mentioned in the linked issue), https://docs.gtk.org/gtk3/class.Label.html#links would be a start.

runkaiz:
I believe most of not all of the requested changes have been applied. I might have misinterpreted the part about hiding the enable / disable button on the default back-end so I would appreciate it if someone can help me review this.

GTG/gtk/backends/addpanel.py
```diff
@@ -170,7 +172,16 @@ def on_combo_changed(self, widget=None):
         """
         backend_name = self.combo_types.get_selected()
         if backend_name is None:
+            if 'backend_caldav' in inactive_modules:
+                markup = '<big>Error: Python package \'caldev\' not installed.</big>'
```
Neui, line 176:
```suggestion
                markup = '<big>{}</big>'.format(_('Error: Python package \'caldav\' not installed.'))
```
Typo in `caldav`, and make it translatable.

GTG/gtk/backends/addpanel.py
```diff
@@ -170,7 +172,16 @@ def on_combo_changed(self, widget=None):
...
+                self.label_name.set_markup(markup)
+                return
+
+
+            markup = '<big>Error: An unknown backend could not be loaded.</big>'
```
Neui, line 181:
```suggestion
            markup = '<big>{}</big>'.format(_('Error: An unknown backend could not be loaded.'))
```

Make it translatable.

GTG/backends/__init__.py
```diff
@@ -67,10 +67,12 @@ def __init__(self):
                 # Something is wrong with this backend, skipping
                 log.warning("Backend %s could not be loaded: %r",
                             module_name, exception)
+                inactive_modules.append(module_name)
```
Neui, line 70:
```python-traceback
  File "/app/lib/python3.9/site-packages/GTG/backends/__init__.py", line 65, in __init__
    __import__(extended_module_name)
  File "/app/lib/python3.9/site-packages/GTG/backends/backend_caldav.py", line 30, in <module>
    import caldav
ModuleNotFoundError: No module named 'caldav'

During handling of the above exception, another exception occurred:

  File "/app/lib/python3.9/site-packages/GTG/gtk/application.py", line 124, in do_startup
    for backend_dic in BackendFactory().get_saved_backends_list():
  File "/app/lib/python3.9/site-packages/GTG/backends/__init__.py", line 70, in __init__
    inactive_modules.append(module_name)
AttributeError: 'set' object has no attribute 'append'
```

You need to use [`.add()`](https://docs.python.org/3.8/library/stdtypes.html#frozenset.add) since it is a set, not an list.

GTG/backends/__init__.py
```diff
@@ -67,10 +67,12 @@ def __init__(self):
...
                 continue
             except Exception:
                 # Other exception log as errors
                 log.exception("Malformated backend %s:", module_name)
+                inactive_modules.append(module_name)
```
Neui, line 75: Ditto. You need to use [`.add()`](https://docs.python.org/3.8/library/stdtypes.html#frozenset.add) since it is a set, not an list.

GTG/backends/__init__.py
```diff
@@ -32,7 +32,7 @@
 from GTG.backends.generic_backend import GenericBackend
 
 log = logging.getLogger(__name__)
-
+inactive_modules = set()
```
Neui, line 35: Might be better to move this to the `BackendFactory` class below. The [`Borg`](https://github.com/getting-things-gnome/gtg/blob/5f72c8a9319e06d42967fb8461ea832e8cd095d9/GTG/core/borg.py) class it inherits is basically making it a singleton (state shared on every instance).

Neui (changes requested):
I am not sure if we should make a link to the XML file. At least for me, clicking on it does nothing, and if something did, it may encourage people to modify the file while GTG is still open.

Images would have been nice to look at. Just launch `gnome-screenshot -i`, select region, copy to clipboard and paste inside GitHub textbox thing.

Default

After fixing the `.append()` issue:
caldav not found

I don't quite like how it looks like. Maybe allow Caldav to be selectable, but name it like "caldav (error)" or something, so if you have multiple backends you can see which ones have an error.

nekohayo:
The big refactoring has finally landed (see the pinned 0.7 tracking issue for details)!

Please rebase this to the latest `master`. Apologies for the inconvenience; the new codebase is expected to be much better to work with.

nekohayo:
Once this gets rebased, I would like @jaesivsm to help reviewing this.

pentux-GitHub:
The need is still real, #837 is still open. The blocker is that this predates the GTK4 port and touches backend UI files reworked since, so it needs a rebase, as @nekohayo asked back in 2024.

@runkaiz if you'd like to rebase it on current master, it'd be welcome; otherwise it could be closed and #837 left open for a fresh attempt.

runkaiz:
Alright I am on it.

runkaiz:
Rebased onto current `master` (the three old commits are replaced by a fresh implementation on top of the new codebase).

Since the GTK4 refactor removed `backend_localfile`, parts of the original approach no longer applied, so the scope is now:

- **`BackendFactory` records failed modules**: `inactive_modules` maps each backend module that failed to import to its error message, exposed via `get_inactive_modules()`. This also fixes the `set`/`.append()` mix-up @Neui caught in the original review.
- **Add panel explains the empty state**: with no loadable service (e.g. python-caldav not installed), the panel used to show an empty combo box, blank labels, and a no-op OK button. It now hides the combo, disables OK, and lists which modules failed and why — generically, not hardcoded to caldav. One subtlety: an empty combo never emits `changed`, so `refresh_backends()` now calls `on_combo_changed()` explicitly, otherwise the explanation would never render.
- **Configure panel hides (rather than greys out) the Enable/Disable syncing button** for the default backend, per #837.

Dropped from the original PR based on review feedback: the hyperlink to the XML data file (it also relied on a hardcoded Flatpak path) and the caldav-specific error string.

Verified locally on a machine without python-caldav (the exact #837 scenario): the factory reports `{'backend_caldav': "No module named 'caldav'"}` and the add panel renders the explanation with the OK button disabled.

Written by AI.

runkaiz:
Definitely need a live run from someone, I do not have a live linux environment at the moment. Compilation and tests passes.

pentux-GitHub:
Live run done on Linux with the bryce sample dataset, on top of current master (merges cleanly, no conflict with the recent CalDAV commits).

Nominal case: the add dialog lists CalDAV as before, no visible regression. Simulated missing dependency (caldav import blocked): the dialog now shows "No synchronization services available" with the failing module and the reason, OK disabled. That is exactly what #837 asked for and it works as described.

One honest side note: a few GTK warnings show up in the console when the empty state is displayed (a GtkStack child name lookup and two negative size reports). I have not established whether they come from this branch or were already there, and they don't affect the behavior.

From my side this looks ready for review. The CI still needs a maintainer to approve the workflow run on this fork.

pentux-GitHub:
The rebase requested above is done and live tested (previous comments). Two things remain: the CI workflow on this fork still needs a maintainer approval to run, and nekohayo wrote earlier that once rebased he would like @jaesivsm to review. The rebase condition is now met, so pinging you here, when you have a moment.

PR #899, opened 2022-05-29
Error handler: Add Flatpak infos

Neui:
Just in case it matters, and wanting to trying out to get the dev version of GTG running inside flatpak.

Error handler demonstration with the new lines


diegogangl:
This will definitely be useful, thanks!

PR #894, opened 2022-05-14
Gtk4 + New Core + Listview Megaport

diegogangl:
red-pill

--- 

This PR rebases and solves the conflicts in @ranchester2 's Gtk4 port (sometimes succesfully). It will also include my work on listviews and will finally plug the new core completely into the new UI.

:fire:  Note that everything in this branch is broken and on fire. It will take _quite a while_ before this is even testable. :fire:

## Current Status

_Will fill this up more as I go along._

- The Gtk4 port is complete
- The new widgets are mostly done, except for the horde of hidden bugs and regressions
- The port to the new core is missing several actions, plugins and backends (also tons of bugs probably)
- I plan on going back to the new core once everything is plugged in to streamline and simplify things (look at the cleanup section below)

## Other Changes

- The Tags section can now be collapsed, like saved searches. They also look more like the sidebar items in Gnome Builder
- Tags can now be renamed without destroying their parent relationship or doing "weird stuff"
- Saved searches have a name that is independent of the query
- The saved search query can be edited in a new dialog
- The saved search context menu is different from the tags too
- Tasks can be dragged over tags in the sidebar to tag them
- Multiple tags can be selected in the sidebar, the selected tasks are the ones that have all the selected tags. (eg. if you select "@errands" and "@prio1", only tasks that have both of those tags show up).
- There is a new sort menu above the tasks list
- Tasks can be sorted by added/modified/start/due dates, alphabetically and by tags

## Clean-up and other todos

- [ ] Add new unit tests for: `replace_tag` and `remove_tag` (in tasks)
- [x] Use TagPill in the sidebar
- [x] Stop using ScrolledWindow for taskpane and sidebar, they are already inside ScrolledWindow
- [ ] Put `unwrap()` in a utils somewhere
- [x] Add proper icons for due and start date
- [ ] Change the way color is stored in the file, document and add versioning. 
- [x] Remove all the stupid `[:2]` hacks for colors
- [x] Move the code to add row color to a method
- [x] Look into the padding of the rows in the task list
- [ ] Consider extending `Gio.ListModel` in the store classes directly
- [x] Make tags in tasks a set instead of list (duh)
- [x] Use `bind()` in task pane
- [ ] Figure out how to make `bg_func` in plugins now that we use css
- [x] Figure out how to connect signals for tasks being modified (probably through the model?)

## Fixes, etc for Github

fixes #307
fixes #874 
closes #308
fixes #904
fixes #817
fixes #302 
fixes #240 
fixes #420
fixes #123
fixes #859

Issue #737, opened 2021-12-02
Tracking Issue for 0.7


diegogangl:
We're getting to the end of the 0.6 cycle, so it's a good time to start planning for 0.7 since we have lots of large, sweeping changes to land. This issue is meant to centralize, keep track of all the work and plan ahead so we don't end up with another dev hell scenario.

## Landing current PRs

- [ ] Land the new core PR #668 
- [ ] Land the Gtk4 port #723 
- [ ] Pin an issue mentioning the drag and drop regression, so we don't get extra tickets about that

## Polishing and completing the port

The new core was designed with Gtk3 in mind, with the idea to port over to listboxes and then onto Gtk4. @ranchester2 port changed the schedule a bit, so we have to move directly to Gtk4's listviews and models. We also need to finish some missing features.

- [ ] Add code for recurring tasks (@zeddo123 's territory)
- [ ] Check backends code with @jaesivsm 
- [ ] Port some properties in data classes to GObject properties (things that will need to be bound in widgets)
- [ ] Add a function to generate child models in store classes (using the children list)
- [ ] Add a model property to store classes
- [ ] Move the filtering  code to several `Gtk.Filter` subclasses in a their own modules (probably in a models folder)
- [ ] Add a `Gtk.Filter` subclass for search
- [ ] Move the sorting  code to several `Gtk.Sorter` subclasses in a their own modules too

## Plugging in the new core

Slowly slide in the new core so we can validate how it works

- [ ] Initialize `Datastore2` to `Application` and read the `gtg_data.xml` file
- [ ] Check the tasks are read correctly (using the dev console)
- [ ] Add a call to save with the new core in `requester.save_datastore`, set the path to `gtg_data2.xml`
- [ ] Check the xml files are the same
- [ ] Add calls to all the other parts of the requester , so both cores do the same (remove task, new task, etc.)
- [ ] Use the new core to setup the task editor, tag editor and batch tag editor

## Porting to listviews

There's a lot of new stuff regarding lists in Gtk4. I've written a quick 'n dirty (mostly dirty) script to test them out:
https://gist.github.com/diegogangl/eba75d467854068ea75cc9a03ce61495

This shouldn't include a lot of UI changes unless they are really simple, or necessary (like the new sorting)

- [ ] Design a UI for the new sorting, since we won't have treeview columns (kind of like the nautilus menu)
- [ ] Subclass Gtk.Box for each data type, so it can hold a pointer to the original object (_not sure this is the best way though_)
[...]

PR #931, opened 2022-09-19
index backends by id, not by name

### LINKED ISSUE #930: GTG allows creating multiple backends of type CalDAV, but they do not persist across restarts
The UI allows creating multiple CalDAV backends (and optionally giving them different, custom names), and they seem to work OK, but exiting and reopening GTG loses all but one backend of each kind.

The cause seems to be that the config file `backends.conf` has a single section per backend kind (i.e. one for `backend_localfile` and one for `backend_caldav`).

Maybe it would be enough to use backend IDs instead of backend names in `backends.conf`.

cpitclaudel:
This allows GTG to have multiple backends of one kind at the same time.

* GTG/backends/__init__.py (BackendFactory.get_saved_backends_list): Rename
`backend` to `backend_id` for clarity.  Call `config.rename_backend_section` on
each backend for compatibility with previous versions of GTG.
* GTG/core/config.py (CoreConfig_.rename_backend_section): New function to
rename a config section (from backend name to backend ID).
(CoreConfig_.get_all_backends): Rename `backend` to `backend_id` for clarity.
* GTG/core/datastore.py (DataStore.save): Retrieve config section by backend id,
not by backend name.

Fixes #930.

### DIFF
--- a/GTG/core/config.py
+++ b/GTG/core/config.py
@@ -203,10 +203,14 @@ class SectionConfig(GObject.Object):
         self._save_function()
 
 
-class CoreConfig:
+class CoreConfig_:
     """ Class holding configuration to all systems and tasks """
 
+    _instance = None
+
     def __init__(self):
+        assert self._instance is None
+
         self._conf_path = os.path.join(CONFIG_DIR, 'gtg.conf')
         self._conf = open_config_file(self._conf_path)
 
@@ -215,6 +219,11 @@ class CoreConfig:
 
         self._backends_conf_path = os.path.join(CONFIG_DIR, 'backends.conf')
         self._backends_conf = open_config_file(self._backends_conf_path)
+
+    @classmethod
+    def get_instance(cls):
+        cls._instance = cls._instance or cls()
+        return cls._instance
 
     def save_gtg_config(self):
         self._conf.write(open(self._conf_path, 'w'))
@@ -254,3 +263,6 @@ class CoreConfig:
             self._backends_conf[backend],
             DEFAULTS['backend'],
             self.save_backends_config)
+
+def CoreConfig():
+    return CoreConfig_.get_instance()
--- a/GTG/core/config.py
+++ b/GTG/core/config.py
@@ -251,16 +251,25 @@ class CoreConfig_:
             DEFAULTS['task'],
             self.save_task_config)
 
+    def rename_backend_section(self, backend_name, backend_id):
+        """Rename section `backend_name` to `backend_id` if it exists."""
+        if backend_name in self._backends_conf:
+            assert backend_id not in self._backends_conf
+            self._backends_conf.add_section(backend_id)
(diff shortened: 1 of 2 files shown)


diegogangl:
@cpitclaudel hey, thanks! We're in the middle of a big port ATM and I'm focusing all the time I can find in that. Once #894 is ready to land, I'll review this and merge. Luckly it looks like this branch won't generate any conflicts, though we'll have to port the change in datastore over to the new datastore.

cpitclaudel:
Looking forward to that :) Good luck with the port!

GTG/core/config.py
```diff
@@ -180,10 +180,14 @@ def save(self):
         self._save_function()
 
 
-class CoreConfig():
+class CoreConfig_():
```
diegogangl, line 183: Why the underscore here?

diegogangl (changes requested):

GTG/core/config.py
cpitclaudel: For backwards compatibility: `CoreConfig` is redefined below:

```
def CoreConfig():
    return CoreConfig_.get_instance()
```

nekohayo:
It is time now! Monsieur @cpitclaudel, rebasez donc cette branche, et lâchez les chiens!

See also the updated general status and convenient links in #737.

nekohayo:
This will also be relevant to @jaesivsm's interests.

cpitclaudel:
Congrats on the big merge! :)

I rebased this, but I'll need help to test:

- I haven't used GTG since I opened this PR
- I'm not familiar with the code base anymore
- At first glance, I'm not 100% sure how the new datastore should be updated

After a quick glance through the code, I'm also not entirely certain of what a backend module refers to, so the change may not be right in the new architecture?

diegogangl:
Hi @cpitclaudel, sorry for the delay. At first glance, this sounds like a good idea. The problem is that the only backend module we have now is caldav, since the "local" backend is now integrated into the datastore. And the Caldav backend has a fair share of bugs too from the refactor. On top of that, there are some potential problems with having more than caldav backend like race conditions when synchronizing that need careful work and testing.

Since we are trying to finish 0.7 at this moment, I'll remove the milestone for this and leave it open for later

cpitclaudel:
Sure thing, good luck with the 0.7 release!  I look forward to using GTG when CalDAV is back :)

cpitclaudel:
Closing, given this:

> Same goal, converging implementations — yours just sat in the queue too long. :)
>  _@pentux-GitHub in https://github.com/getting-things-gnome/gtg/issues/1285#issuecomment-5027472421_

pentux-GitHub:
Thanks @cpitclaudel for the graceful close, and your PR framed the problem well before mine did; indexing backends by id instead of name was the right call. Glad we converged on it. 🙂

PR #935, opened 2022-10-20
Fix exception during tag removal

paulgear:
Task2.remove_tag() expects a string, not a tag object.

Also fix grammar in error message shown when this bug is triggered.  I think "should've" is not very easy for non-native English speakers, so I replaced all instances of it with the non-contracted form.


pentux-GitHub:
The underlying bug is already gone on current master: `remove_tag(self, tag_name: str)` now takes a string, and the caller in `modify_tags.py` passes `_tag.name` correctly. So the fix here has effectively been absorbed by the GTK4 port.

Thanks @paulgear for catching it. Safe to close unless something still reproduces.

PR #937, opened 2022-11-04
Quote SOURCEFILE in script_pocketmod to handle spaces in path

mzf-guest:
The export script `GTG/plugins/export/export_templates/script_pocketmod` does not quote the input parameter `SOURCEFILE` which make it fails with spaces in this path.

This pull request adds quotes when reading the input argument and when using the variable as command argument.

Patch from the Debian package team:
https://salsa.debian.org/python-team/packages/gtg/-/blob/debian/master/debian/patches/quote_source_file_in_script_pocketmod.patch


Neui (approved):
Currently I don't have the packages installed/set up to test, but the changes itself look good.

diegogangl:
Thanks, merged!

PR #938, opened 2022-11-04
Unknown macro .bu in man page

mzf-guest:
This pull request changes unknown macro .bu to valid groff syntax for bullet lists.

From the downstream debian package:
https://salsa.debian.org/python-team/packages/gtg/-/blob/debian/master/debian/patches/fix_man_page_bullet_list.patch


Neui (approved):
Changes works for me and looks better than inline text.

Neui:
It appears `.BU` was in a [troff macro pack called *ms*](https://www.troff.org/TheGroffFriendsHowto.pdf), but it appears [groff's *ms*](https://www.gnu.org/software/groff/manual/html_node/Lists-in-ms.html#Lists-in-ms) doesn't support that.

diegogangl:
Merged, thanks!

PR #946, opened 2022-12-30
Fixed missing build dependancy for Fedora

NULL0404:
Building GTG from source also requires gtksourceview4-devel, so I added it here

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ You can get most of those from your distribution packages:
 
 ```sh
 # On Fedora:
-sudo dnf install meson python3-cairo python3-gobject gtk3 itstool gettext python3-lxml libsecret gtksourceview4-devel
+sudo dnf install meson python3-cairo python3-gobject gtk3 itstool gettext python3-lxml libsecret gtksourceview4
 # On Debian 10 (buster), you need to install the backported version, activate it with:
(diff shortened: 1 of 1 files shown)


Neui:
Not entirely sure we need the full `-devel` package. It seems that package [ships a `GtkSource-4.gir` file](https://packages.fedoraproject.org/pkgs/gtksourceview4/gtksourceview4-devel/fedora-36.html), but according to [PyGObject documentation](https://pygobject.readthedocs.io/en/latest/index.html#how-does-it-work) it uses a "typelib" (confirmed by [girepository documentation](https://gi.readthedocs.io/en/latest/)  only using a typelib at runtime in the diagram), which probably is `GtkSource-4.typelib` [shipped by `gtksourceview4`](https://packages.fedoraproject.org/pkgs/gtksourceview4/gtksourceview4/fedora-36.html) on Fedora, which the -devel package depends on.

Debian also has that [`GtkSource-4.gir` in `libgtksourceview-4-dev`](https://packages.debian.org/bullseye/amd64/libgtksourceview-4-dev/filelist) but the [`GtkSource-4.typelib` in `gir1.2-gtksource-4`](https://packages.debian.org/bullseye/amd64/gir1.2-gtksource-4/filelist) (which is currently missing in Debian/Ubuntus `apt` command).

diegogangl:
Sorry for the delay, merged. Thanks!

PR #953, opened 2023-01-28
readme: Remove outdated and obsolete build status badge

picsel2:
Update build status badge to use the github workflow and move it more towards the top.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,9 +203,3 @@ yelp docs/user_manual/C/index.page
 * Our wiki serves as our website: https://wiki.gnome.org/Apps/GTG
 * Check out the [docs folder in the main repository](./docs/) for more
   information and documentation for contributors
-
-## Test suite status
-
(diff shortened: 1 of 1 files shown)


Neui:
I think it can either be after the social media badges or on a new line below the other badges, since that beginning section is primary information including normal users, so I don't think it is "important" for normal users to know the tests state.

Also I am not sure if even need such badge at all, since GitHub already has such feature next to the commit (and below PRs), and is visible without scrolling (I almost never look at the README if I am not a user).

picsel2:
I've updated the PR to remove the obsolete badge completely.

wieslawmlynarski:
What about just replacing it with something like this

`
[Unit Tests](https://github.com/getting-things-gnome/gtg/actions/workflows/unit_tests.yml)
`

[Unit Tests](https://github.com/getting-things-gnome/gtg/actions/workflows/unit_tests.yml)

diegogangl:
I think deleting it is the best way to go, thoughts on this @nekohayo ?

diegogangl:
Merged, thanks!

PR #966, opened 2023-04-08
flatpak: Generate python dependencies with flatpak-pip-generator

picsel2:
This should ease updating dependencies.


diegogangl:
Nice one, thanks!

PR #967, opened 2023-04-08
Register pytest test with meson

picsel2:
This lets us run tests using the build system.
It also makes GNOME Builder display the test in its tests list so we can execute them from there.

### DIFF
--- a/GTG/gtg.in
+++ b/GTG/gtg.in
@@ -70,10 +70,8 @@ if __name__ == "__main__":
     except ImportError:
         pass
 
-import gi
-gi.require_version('Gdk', '4.0')
-gi.require_version('Gtk', '4.0')
-gi.require_version('GtkSource', '5')
+from GTG import gi_version_requires
+gi_version_requires()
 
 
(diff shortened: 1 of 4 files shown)


nekohayo:
@sqatx this is also relevant to your #999 I think?

SqAtx:
Cool! Does this mean we can get rid of the `run-tests` script?

If yes, could you add a section to the README after "Running the beast" with instructions explaining how to run the tests via Meson? I tried the obvious thing and it didn't work:

```
~/src/gtg$ meson test

ERROR: No such build data file as '/home/kevin/src/gtg/meson-private/build.dat'.
```

picsel2:
The meson tutorial shows [how to build and test with meson](https://mesonbuild.com/Quick-guide.html#compiling-a-meson-project).
I found some _really_ outdated instructions inside the [`docs/` directory](https://github.com/getting-things-gnome/gtg/blob/master/docs/contributors/coding%20best%20practices.md).
I think I will update them and point there.

> Cool! Does this mean we can get rid of the run-tests script?

I hope so :smile:

diegogangl:
LGTM, merging. Thanks!

PR #981, opened 2023-07-15
Properly maintain the task count in Tag.set_parent (fix #404)

### LINKED ISSUE #404: Parent tag's tasks count is not updated when adding/removing children tags to them
In the sidebar, let's say you have a tag called "foo". If you drag other tags onto "foo", or drag children tags of foo out of foo (so that they become top-level), the task counter for "foo" will change only after you restart the app.

This is probably a trivial fix, so marking as a low-hanging-fruit for anyone looking at this to tackle for fun.

gycsaba96:
Hi!

I am a newcomer. This PR fixes issue #404 .

Parent tag's task count was not updated when adding/removing children tags to them.
The problem is solved by overriding the `set_parent` method in the `Tag` class and introducing a `_get_ancestors`  helper method.

### DIFF
--- a/GTG/core/datastore.py
+++ b/GTG/core/datastore.py
@@ -222,7 +222,7 @@ class Datastore:
                 if not task.tags:
                     count['untagged'] += 1
 
-                for tag in task.tags:
+                for tag in { t for owned_tag in task.tags for t in [owned_tag] + owned_tag.get_ancestors() }:
                     val = count.get(tag.name, 0)
                     count[tag.name] = val + 1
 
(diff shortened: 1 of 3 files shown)


nekohayo:
Hi @gycsaba96, the big refactoring has finally landed (see the pinned 0.7 tracking issue for details)!

Could you check whether this is still needed with the current git `master` version? It might have been fixed as part of the mega branch. Otherwise, you will need to `git rebase` this PR.

gycsaba96:
Hi Jeff,

The problem is still present in the new version. However, I might need some time to find a proper solution using the new codebase. (My old solution relied on the base class of Tag, which had changed.)

gycsaba96:
Hi Jeff,

I have fixed the bug:
- Now, the ancestors of a tag are also considered when updating the task counts.
- After a drag-and-drop, these counts are recalculated.

diegogangl:
Sorry for all the rebasing, this LGTM
Merging, thanks!

PR #986, opened 2023-09-18
Fix an inappropriate test expression to remove a logical short circuit

munahaf:
In file: main_window.py, the comparison of Collection length creates a logical short circuit. I suggested that the Collection length comparison should be done without creating a logical short circuit. 

Sponsorship and Support:

This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -1418,7 +1418,7 @@ class MainWindow(Gtk.ApplicationWindow):
         if self.tagtreeview:
             taglist = self.tagtreeview.get_selected_nodes()
         # If no selection, we display all
-        if not nospecial and (not taglist or len(taglist) < 0):
+        if not nospecial and (not taglist):
             taglist = ['gtg-tags-all']
         if nospecial:
             special = ['gtg-tags-all', 'gtg-tags-none',
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -1418,7 +1418,7 @@ class MainWindow(Gtk.ApplicationWindow):
         if self.tagtreeview:
             taglist = self.tagtreeview.get_selected_nodes()
         # If no selection, we display all
-        if not nospecial and (not taglist):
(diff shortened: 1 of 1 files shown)


Neui:
The change itself looks good to me (I doubt that `len()` here will return a negative value, and untested), but the commit message is a bit too generic. Could it be a bit more specific? Like "Remove tautology regarding selected tags filter" or something.

nekohayo:
@munahaf Could you use `git rebase -i` onto the latest version of our master branch, to reword your commit message as per the feedback above?

nekohayo:
Related:
* #992
* #999

diegogangl:
Agree with @Neui , `len()` never returns a negative unless you implement the dunder method yourself and call it manually like `tags.__len__()`

munahaf:
Changed the commit message per suggestion.

GTG/gtk/browser/main_window.py
```diff
@@ -1418,7 +1418,7 @@ def get_selected_tags(self, nospecial=False):
         if self.tagtreeview:
             taglist = self.tagtreeview.get_selected_nodes()
         # If no selection, we display all
-        if not nospecial and (not taglist or len(taglist) < 0):
+        if not nospecial and (not taglist):
```
SqAtx, line 1421:
```suggestion
        if not nospecial and not taglist:
```

SqAtx (approved):
That was an odd piece of code. The check is indeed not necessary, and we can also remove the extra brackets now.

diegogangl:
@munahaf, sorry for pushing here and un-drafting. I'm trying to get things in faster :)
Thanks for the work!

Issue #390, opened 2020-07-05
Clicking / Switching between saved searches is broken (does not re-set the search field, stacks the searches, or breaks filtering in general)


nekohayo:
If you have some existing saved searches, and you create a new one, and then you click one of the other saved searches, you get 0 results because the newly created saved search is still active and applied on top of the now-clicked older saved search. Some might consider this a feature (?) but I consider it confusing unexpected behavior... I was rather expecting the search field's contents to be re-set to whatever saved search I just clicked.

There's also the question of if the saved search should also be cleared entirely when focusing something other than a saved search in the sidebar, such as one of the tags or "All tasks" or "Tasks with no tags"...

nekohayo:
Additionally: when switching between tags selection in the sidebar, the existing filtering search query in the searchbar is not applied to the newly chosen tag view.

diegogangl:
> Additionally: when switching between tags selection in the sidebar, the existing filtering search query in the searchbar is not applied to the newly chosen tag view.

This last bit is kind of challenging with our current architecture. When clicking on a saved search we set the search query, so when clicking on a tag we have to unset it. But we can't distinguish when the query was set by the user using ctrl+f or when it was set programatically.

pentux-GitHub:
The issue this targets, #390, has since been closed, and as @nekohayo noted the GTK4 port changed the symptoms here. So this PR is likely superseded.

@AidanV if the saved-search behaviour still looks wrong to you on current master, a fresh issue would be the way to go; otherwise this can probably be closed.

PR #989, opened 2023-10-11
Clicking on a saved search updates the search bar to match the saved search.

AidanV:
Fixes issue #390

The approach is to have all saved searches act as restoring the search bar to when it was saved.


nekohayo:
Hi, this needs to be rebased and/or redone because of the big changes that landed with the GTK4 port now.

The symptoms have changed.

Would you like to reattempt fixing it? Please have a look at the new comment and demonstration video I have added in #390.

Marking this as a draft until then.

PR #992, opened 2023-12-03
migrate from external mock package to stdlib unittest.mock

eli-schwartz:
It was moved to the stdlib in python 3.3, and the PyPI version is a backport for python versions less than 3.3.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -1,7 +1,7 @@
 # Run unit tests script
 
 name: Unit Tests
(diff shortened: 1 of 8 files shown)


nekohayo:
Relevant to @sqatx's #999 I think?

eli-schwartz:
Yes, I would say so.

This change gets you closer to a testsuite that is self contained and requires less mysterious state before you can start, which can only be a good thing.

diegogangl:
Merged, thanks!

Issue #399, opened 2020-07-08
Better testing, with automated UI integration tests [RFC]


diegogangl:
This is an issue to discuss how we could improve our automated testing (or lack of). Some things that have been suggested:

- Move unit tests to meson, so they don't depend on a separate makefile 
- Use [Dogtail](https://gitlab.com/dogtail/dogtail) for automated UI testing

nekohayo:
Having greater test coverage, especially UI "integration testing" (i.e. what a user would actually encounter in practice) with Dogtail (I don't know if there's anything better these days on the Linux desktop, but I'm guessing this is still the go-to thing to get it done easily) would allow refactoring and merging improvements with the proven assurance that things are improving (or at least not regressing).

I think **covering the Task Editor should be the first target** of the test suite, because that's where a lot of stuff can go wrong in different ways, and it's difficult to manually test everything all the time. Dogtail would be particularly useful if it can handle rich text formatting somehow, for the gazillion potential little bugs and gotchas in GTG's "Task Editor".

The main window ("Task Browser") is less urgent, in part because it is supposed to already be covered by liblarch's test suite (?) and in part because Diego is considering providing a new GtkListBox-based widget backend for liblarch instead of something based on GtkTreeView.

I'm considering this ticket a low-hanging-fruit because in theory Dogtail is so easy and fun to use (you tell it that the GUI should react this way or that way, or have this widget or this text or this property after typing some text or clicking some button, and if it doesn't match your expectations, you raise a test failure... here are some examples of [how it was done in Pitivi many years ago](https://gitlab.gnome.org/GNOME/pitivi/-/tree/27b3bdd10fa6b752c6507c17fb0ca1dc979ab51e/tests/dogtail_scripts), with an old version of Dogtail, and here are some more recent [examples from the Dogtail project](https://gitlab.com/dogtail/dogtail/-/tree/master/examples))... that any new contributor could work on this (and provide a very valuable contribution by doing so).

nekohayo:
Fast-forward nearly 4 years later:
* Unless someone comes up with a better alternative, I'd still love to be able to use Dogtail for UI integration testing, but Dogtail's development seems to have been quite dead for the past few years.
[...]

PR #999, opened 2024-02-24
Fix tests

SqAtx:
Or start to..

Tests are very broken, with some test files not even importing https://github.com/getting-things-gnome/gtg/actions/runs/7966225770/job/21747114127

I'd like to first get to a point where the test suite is green again so we can add new tests when fixing bugs etc. Even if it means temporarily skipping a bunch of them - it's still better than having a test suite that's perpetually red :)

It isn't much right now, but I thought I'd share in case someone with more free time than me wants to give it a go as well.

### DIFF
--- a/tests/core/test_tag.py
+++ b/tests/core/test_tag.py
@@ -18,7 +18,7 @@
 
 from unittest import TestCase
 
-from GTG.core.tag import Tag
+from GTG.core.tags import Tag
 
 
 class TestTag(TestCase):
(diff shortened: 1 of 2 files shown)


nekohayo:
In case you're interested in the broader picture, too: #399

SqAtx:
We can close this in favour of #1036. I'll focus on making changes on top of that to make the test suite green unless @picsel2 beats me to it :)

Issue #622, opened 2021-04-11
Compatibility with GNOME Builder: `bwrap: execvp gtg: No such file or directory` when running GTG from Builder


severin-lemaignan:
### Dev environment:

- Ubuntu 20.04
- Builder 3.38.2

### Steps to reproduce:

- launch Builder
- clone the github GTG repo (cloned at 68305160b6)
- Builder asks to install the Gnome 40 runtimes; they successfully download + install
- Pressing the 'Run' button triggers the installation of all the Python3 dependencies (as execpted)
- However execution ultimately fails with `bwrap: execvp gtg: No such file or directory`

Note that `gtg` runs well if started from the command-line with `launch.sh`. As such, this issue is purely a Gnome Builder integration issue. Please let me know if you think it should be reported to the Gnome Builder issue tracker instead.

diegogangl:
I can build fine with master, running Fedora 33. Maybe it's an issue with the flatpak version?

nekohayo:
I'm marking this as a low hanging fruit to encourage packaging-savvy or potential new contributors to contribute a fix for this, otherwise I think it's unlikely to be addressed in the near future.

PR #1025, opened 2024-02-27
Fix Flatpak for GNOME Builder compatibility 

TheEvilSkeleton:
Fixes https://github.com/getting-things-gnome/gtg/issues/622

Includes: https://github.com/getting-things-gnome/gtg/pull/1035, https://github.com/getting-things-gnome/gtg/pull/1034


nekohayo:
One question I have is, you are renaming the app ID from `org.gnome.GTGDevel` to `org.gnome.GTG.Devel`. While I have no particular opinion on that, IIRC it might break some people's workflows as they will have to uninstall and reinstall the nightly flatpak, I think (though I guess that's way less problematic than if it was the main stable version's flatpak). I once [tried](https://gitlab.gnome.org/GNOME/nautilus/-/issues/3141) to propose this kind of change for Nautilus and got scolded for that ;)

Do you see a technical advantage to this?

flatpak/org.gnome.GTG-stable.json
nekohayo, line 1: This seems to remove the stable version's flatpak, with no replacement?

flatpak/requirements.txt
nekohayo, line 1: This (and other changes) seems to remove the liblarch dependency. I just want to make sure, is this correct @diegogangl ? I saw some liblarch mentions with `git grep liblarch GTG/` (and elsewhere), particularly in `gtk/browser/main_window.py`, not just `plugins/hamster/hamster.py`, though I suppose maybe those are just outdated comments that ought to be renamed/updated in a separate commit (possibly prior to this)?

flatpak/org.gnome.GTG-stable.json
TheEvilSkeleton: Yeah, because the stable version is in Flathub's organization. The one we have here is strictly for development purposes

flatpak/requirements.txt
TheEvilSkeleton: It's right here https://github.com/getting-things-gnome/gtg/blob/28bd2170a7c368400a2c558d8fc898e9ecaadd21/build-aux/pypi-dependencies.json#L15

TheEvilSkeleton:
> Do you see a technical advantage to this?

No, it's just consistent with the [GNOME Developer Documentation](https://developer.gnome.org/documentation/index.html)

nekohayo:
Along with @diegogangl and possibly @leio, I believe @Neui might be interested in reviewing this, as I don't really have the expertise to make a judgement on the changes this PR introduces.

If there are no problems with this, I would be looking forward to this landing soon, to facilitate the development workflow of the wider community/newcomers who may be relying on GNOME Builder instead of our native launch script.

picsel2 (commented):
I think this PR does too many things at once:

1. It changes the app id of the development build.
2. It updates the runtime dependencies.
3. It changes the python dependencies module to json format and renames it from python3-requirements-dev.yaml to pypi-dependencies.json.

I currently use the org.gnome.GTGDevel variant but actually do not care if it gets renamed.

Any reason not to use the yaml output flatpak-pip-generator? I think the yaml format is far more readable here and has real comment support.

It looks to me that only updating the runtime is relevant for GNOME Builder. Maybe breaking this up in smaller PRs could be a good thing?

diegogangl:
Thanks for working on this @TheEvilSkeleton . I think it's ok for a single PR, but I'd prefer multiple commits for the changes. Also, as @picsel2 asked why the change to json?

As for the appid, devel versions are for breaking stuff. If it follows Gnome's standards better then it's a good idea to change it.

TheEvilSkeleton:
> Thanks for working on this @TheEvilSkeleton . I think it's ok for a single PR, but I'd prefer multiple commits for the changes. Also, as @picsel2 asked why the change to json?

Gonna be honest, I overlooked it

TheEvilSkeleton:
As for splitting into different commits, I'll do that

TheEvilSkeleton:
Uhhh, I think I broke GitHub...

image

TheEvilSkeleton:
Okay, it fixed itself

Neui:
Personally, I am not a fan of removing the "stable" flatpak JSON, since that makes it a bit harder to run a non-development themed GTG (before GTK4 there was this stripe in the window bar, is one there for GTK4 version? I don't remember...), for example for users who patches GTG before turning into a flatpak or for (release) screenshots (although you could use the GTK inspector to remove the CSS class it adds). It's not a deal breaker for me.

Changes otherwise looks OK to me, but I haven't tried it in Builder or manually nor am I an flatpak packaging expert.

TheEvilSkeleton:
> Personally, I am not a fan of removing the "stable" flatpak JSON, since that makes it a bit harder to run a non-development themed GTG (before GTK4 there was this stripe in the window bar, is one there for GTK4 version? I don't remember...), for example for users who patches GTG before turning into a flatpak or for (release) screenshots (although you could use the GTK inspector to remove the CSS class it adds). It's not a deal breaker for me.

No patches needed; the default option is `default` and not `development`:

https://github.com/getting-things-gnome/gtg/blob/2ecc997d3fb3f6d102615dafa83bda765cebd420/meson_options.txt#L8

picsel2 (approved):
Looks good to me.
The only thing that makes me a bit nervous is the the use of the gnome nightly remote. It looks like without GNOME Builder this requires to install the nightly Sdk (Builder does its own weird stuff with flatpak repos?).
But maybe this could provide more confidence towards platform updates with stable.

picsel2:
Oh, it fails now when I run flatpak-builder from cli:
```
$ flatpak-builder --install --user --force-clean fpkbuild build-aux/org.gnome.GTG.Devel.json
[...]
Rewriting contents of org.gnome.GTG.Devel.desktop
Error: Can't load desktop file /home/sebastian/git/gtg/.flatpak-builder/rofiles/rofiles-Q8jeuP/files/share/applications/org.gnome.GTG.Devel.desktop: Error opening file /home/sebastian/git/gtg/.flatpak-builder/rofiles/rofiles-Q8jeuP/files/share/applications/org.gnome.GTG.Devel.desktop: No such file or directory
```

picsel2:
Nevermind. I just fell again over the issue I already try to solve with #1035.

TheEvilSkeleton:
Oh damn, I completely forgot to change that 😅 

GNOME Builder automatically changes to use the code

TheEvilSkeleton:
@picsel2 I added you as a co-author, since you've helped a lot :)

build-aux/org.gnome.GTG.Devel.json
```diff
@@ -30,17 +30,23 @@
...
   ],
+  "add-extensions": {
+    "org.freedesktop.Sdk.Extension.texlive": {
+      "directory": "texlive",
+      "version": "23.08"
+    }
+  },
   "modules": [
-    "python3-requirements-dev.yaml",
+    "python3-requirements.yaml",
```
picsel2, line 40: I have the texlive extension installed.
When running the flatpak I get the error message:
```
~$ flatpak run org.gnome.GTG.Devel
bwrap: Can't mkdir /app/texlive: Read-only file system
Fehler: ldconfig ist fehlgeschlagen, Exit-Status 256
```

```suggestion
    {
      "name": "extension-points",
      "buildsystem": "simple",
      "build-commands": [
        "mkdir ${FLATPAK_DEST}/texlive"
      ]
    },
    "python3-requirements.yaml",
```

picsel2 (approved):

nekohayo:
I have no opinions / expertise on the diffs there, but from a purely "functional" smoke-testing standpoint: I have tested this now, and it works. I can confirm that the instance it launches has its data separated from the user's regular home data, and from the data that would be created/used by the `launch.sh` script, so in practice it remains compatible and doesn't seem to break anything for me. Approving from that standpoint; I guess now it's up to @diegogangl to have a last look at this to see if everything is OK and merge ;)

It might be worth documenting somewhere in the dev docs that https://github.com/flathub/org.freedesktop.Sdk.Extension.texlive can be installed (and how) for those who may want to develop latex-dependent features in GTG using GNOME Builder, at the cost of an extra 5 GB of disk space for those dependencies.

nekohayo (approved):

diegogangl (approved):

diegogangl:
Works as advertised :)
Thanks for the hard work!

Issue #1027, opened 2024-02-28
Preview the parent task's title summary with a tooltip on the "Open Parent" button in the Task Editor


nekohayo:
Sometimes I just want to quickly have an idea what parent a particular task is blocking.  
A tooltip to reveal this could be nice.

PR #1030, opened 2024-02-28
Show parent's title & children count in "Open Parent" button tooltip

nekohayo:
This allows checking at a glance if the current task blocks a project:

Screenshot from 2024-02-28 02-34-09

Screenshot from 2024-02-28 02-02-19

Screenshot from 2024-02-28 02-02-34

Fixes https://github.com/getting-things-gnome/gtg/issues/1027

---

Potentially, I could maybe also add another commit on top that recursively counts how many parent levels there are, and adds an extra line to say something like:

> In total, %d parent tasks are blocked by the current task.

…what do you think, would that extra info be helpful too, or too much?

### DIFF
--- a/GTG/gtk/editor/editor.py
+++ b/GTG/gtk/editor/editor.py
@@ -471,8 +471,17 @@ class TaskEditor(Gtk.Window):
         if self.task.parent:
             # Translators: Button label to open the parent task
             self.parent_button.set_label(_('Open Parent'))
-            # TODO: preview the parent task's title here:
-            self.parent_button.set_tooltip_markup(_('View the Parent Task'))
+            __tip_contents = _('View the Parent Task:')
+            __tip_contents += f'\n<small><i>{self.task.parent.title}</i></small>'
+            __nb_siblings = len(self.task.parent.children) - 1
+            if __nb_siblings > 0:
+                __tip_contents += "\n\n"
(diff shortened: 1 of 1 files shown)


GTG/gtk/editor/editor.py
```diff
@@ -473,8 +473,17 @@ def refresh_editor(self, title=None, refreshtext=False):
         if self.task.parent:
             # Translators: Button label to open the parent task
             self.parent_button.set_label(_('Open Parent'))
-            # TODO: preview the parent task's title here:
-            self.parent_button.set_tooltip_markup(_('View the Parent Task'))
+            __tip_contents = _('View the Parent Task:')
+            __tip_contents += '\n<small><i>%s</i></small>' % self.task.parent.title
```
diegogangl, line 477: I would prefer an [f-string ](https://docs.python.org/3/tutorial/inputoutput.html)here, since that's what we use in other places.

diegogangl (changes requested):
Looks good, save one nitpick.

GTG/gtk/editor/editor.py
nekohayo: Thank you for that suggestion! I have now done that in this rebased version.

diegogangl:
Looks good!

PR #1034, opened 2024-02-29
Update flatpak runtime

picsel2:
This updates the runtime to Gnome Platform 45.
It required an update of a dependency (I forgot which one) so I updated them all.

Summary of version changes:

* Platform:       40 ->    45
* lxml:        4.5.2 -> 4.9.4
* caldav:      0.8.0 -> 1.3.9
* dbus-python: 1.2.8 -> 1.2.18

### DIFF
--- a/flatpak/requirements-dev.txt
+++ b/flatpak/requirements-dev.txt
@@ -1,5 +1,5 @@
-lxml==4.5.2
-caldav==0.8.0
-dbus-python==1.2.8
+lxml==4.9.4
+caldav==1.3.9
+dbus-python==1.2.18
 git+https://github.com/getting-things-gnome/liblarch.git#egg=liblarch
 pytest
--- a/.github/workflows/weekly_flatpak.yml
+++ b/.github/workflows/weekly_flatpak.yml
@@ -10,7 +10,7 @@ jobs:
     name: "Flatpak"
     runs-on: ubuntu-latest
     container:
-      image: bilelmoussaoui/flatpak-github-actions:gnome-41
+      image: bilelmoussaoui/flatpak-github-actions:gnome-45
       options: --privileged
     steps:
     - uses: actions/checkout@v2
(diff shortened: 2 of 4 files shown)


flatpak/requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
-lxml==4.5.2
-caldav==0.8.0
-dbus-python==1.2.8
+lxml==4.9.4
+caldav==1.3.9
+dbus-python==1.2.18
```
TheEvilSkeleton, line 3: I don't think this one is needed anymore. I removed `dbus-python` in https://github.com/getting-things-gnome/gtg/pull/1025, and it seems to run "just fine" (I haven't tested it extensively, but `import dbus` doesn't fail)

Neui: Note that the Hamster plugin still uses this. https://github.com/getting-things-gnome/gtg/pull/676 is supposed to port from dbus-python to gio dbus. On a quick test in a flatpak SDK shell, `org.gnome.Sdk/x86_64/45` has `dbus-python` but current `org.gnome.Sdk/x86_64/master` doesn't.

TheEvilSkeleton: Ah, thanks

picsel2: Interesting! It looks like it got added with version 45 to the runtime.

TheEvilSkeleton:
I hope I don't come off as rude, but is there a reason we can't merge this MR to https://github.com/getting-things-gnome/gtg/pull/1025 ? I updated the deps as well

picsel2:
> I hope I don't come off as rude, but is there a reason we can't merge this MR to #1025 ? I updated the deps as well

It is all fine. ;)
@nekohayo already asked the same in the matrix room but the conversation got lost on the way. I opened this MR due to a race condition and left it open. I'll better close it.

PR #1035, opened 2024-02-29
flatpak: Use currently checked out source for devel manifest

picsel2:
Currently the devel flatpak manifest "hardcodes" the git master branch as source.
This leads to the surprising effect that freshly hand-crafted flatpaks exclude the hottest local changes of the sources.

This change makes it easier to build and run all the hacks with the cli. 😎


picsel2:
Closing in favour of #1025

PR #1036, opened 2024-03-01
Fix the test suite

picsel2:
This is a follow up on #999 

The goal was to get the CI tests running again and later the scope was extended to fix the whole test suite.


.github/workflows/unit_tests.yml
```diff
@@ -11,18 +11,25 @@ jobs:
   build:
     runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python 3.10
-      uses: actions/setup-python@v2
+    - uses: actions/checkout@v4
+    - name: Set up Python 3.8
```
SqAtx, line 15: What breaks in 3.10?
Reverting to 3.8 makes us change some type annotations that we'll want to change again later, which isn't great.
But if supporting 3.10 is painful, we can go one step at a time - this is already an improvement.
Eventually, I'd like to support every supported Python version? Does that sound reasonable?

picsel2: Thank you for reviewing!

> What breaks in 3.10?

Actually nothing. The readme names 3.8 as minimal supported version, so I figured to set the version used in the ci accordingly.
I would also say that targeting at least the lowest version of Python that still receives security fixes would be a nice for users. 

> Reverting to 3.8 makes us change some type annotations that we'll want to change again later, which isn't great.

I think that this can be done with ease using ruff (or similar): [FA100](https://docs.astral.sh/ruff/rules/future-rewritable-type-annotation/ ), [FA102](https://docs.astral.sh/ruff/rules/future-required-type-annotation/)
This is also the first time that I came across the __future__ module usage for type hints. Maybe we should use these here?

I think that testing against multiple versions (in a matrix?) is something yet to be discussed in #399

SqAtx: Yep okay; it makes sense to target Py38 then.

For the type annotations, and generally for the Py39+ features, we can use future but I think it's also reasonable to just write Python 3.8 code. Anyway, not a conversation specific to this CR.

GTG/core/versioning.py
```diff
@@ -25,7 +25,7 @@
 from lxml import etree as et
 from GTG.core.dates import Date
 from GTG.core.dirs import DATA_DIR
-from GTG.core import xml
+#from GTG.core import xml
```
SqAtx, line 28: What is this? Can be just delete the line?

SqAtx (approved):
I can't run the tests locally right now, but if this diff takes us to a point where individual tests are failing instead of import errors etc, this is a step in the right direction :)

GTG/core/versioning.py
picsel2: I've deleted it now. Thanks!

SqAtx:
It doesn't block the current changes, but I'm still getting failures when running the tests locally:

```
$ ./run-tests 
==================================================================== test session starts =====================================================================
platform linux -- Python 3.11.6, pytest-8.0.1, pluggy-1.4.0
rootdir: /home/kevin/src/gtg
configfile: pyproject.toml
collected 151 items / 3 errors                                                                                                                               

=========================================================================== ERRORS ===========================================================================
___________________________________________________ ERROR collecting tests/backend/backend_caldav_test.py ____________________________________________________
ImportError while importing test module '/home/kevin/src/gtg/tests/backend/backend_caldav_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/backend/backend_caldav_test.py:11: in <module>
    from GTG.core.datastore import Datastore
GTG/core/datastore.py:37: in <module>
    import GTG.core.info as info
E   ModuleNotFoundError: No module named 'GTG.core.info'
```
(and 2 more)
It seems to not like the fact that the extension of https://github.com/getting-things-gnome/gtg/blob/master/GTG/core/info.py.in is `.py.in` maybe?

It might be a Python 3.11 thing - I haven't taken the time to install 3.8 locally.

Anyway, I can fix that separately.

picsel2:
## tldr

Install the project in a prefix of your choice using meson and add the directory with the installed package to PYTHONPATH. Run the tests with `PYTHONSAFEPATH=1 PYTHONPATH="${prefix}/lib/python3.11/site-packages ./run-tests"

## The lenghty explanation

Yeah... running locally is somewhat more complicated since GTG uses [Meson](https://mesonbuild.com/) to build its sources.

### Configuring `info.py`

The file `info.py.in` is a template that gets filled (also called configured) by the meson setup step.
The resulting file is placed into the build directory.
This is the file we actually want Python to find and use.

### Installing the project

Next we have to tell Python of the extra location to search for.
Unfortunately it is not enough to just add the build directory to `PYTHONPATH`.

The sources currently are a non-namespace package ([PEP420](https://peps.python.org/pep-0420/)).
This means here that we need all sources to be merged in one location since Python stops searching at the first `GTG` directory  found, even if it does not contain a `info.py` file.

So instead we do a `meson install` and use the installed location in `PYTHONPATH`.

### Disabling implicit search paths

You now might notice that Python still picks up the git sources, although `PYTHONPATH` points to the installed directory.
The reason is that Python silently adds the current working directory in front `sys.path`, rendering our previous efforts useless.
But you are in luck because Python 3.11 [added a new flag](https://docs.python.org/3/using/cmdline.html#cmdoption-P) that you can enable by setting `PYTHONSAFEPATH=1`.
It disables the implicit addition of the current directory to the search path. Exactly what we need!

GTG/core/datastore.py
```diff
@@ -281,7 +281,7 @@ def do_first_run_versioning(self, filepath: str) -> None:
             self.first_run(self.data_path)
 
 
-    def find_old_path(self, datadir: str) -> None | str:
+    def find_old_path(self, datadir: str) -> Union[None, str]:
```
SqAtx, line 284: `Optional[str]` would work here as well - no need to import Union, actually.

Longer-term, I'd love to avoid `Optional[str]` and just use an empty string as the default value.

GTG/core/saved_searches.py
```diff
@@ -44,7 +44,12 @@ def __init__(self, id: UUID, name: str, query: str) -> None:
         self._query = query
         self._icon = None
 
-        super(SavedSearch, self).__init__()
+        super().__init__()
+
+    # TODO probably not needed
```
SqAtx, line 49: I think we should just remove that and use the constructor in tests. The only thing the method adds is that you don't have to generate the UUID yourself, which I don't think is worth it.

It looks like the code only creates a SavedSearch object when reading from XML. I haven't looked at what happens on the GTK side when we create one, but this concerns me a bit.

SqAtx (approved):
2 minor comments but other than that, LGTM :)

Happy to have a green test run again! Thanks for your work there. I got it running locally as well. We'll add something to the README about that, because it's not obvious that we need to run Meson before we can run the tests.

GTG/core/saved_searches.py
```diff
@@ -160,23 +165,10 @@ def to_xml(self) -> Element:
         return root
 
 
-    def new(self, name: str, query: str, parent: UUID = None) -> SavedSearch:
```
picsel2, line 163: This method is still used by the frontend: https://github.com/getting-things-gnome/gtg/blob/master/GTG/gtk/browser/main_window.py#L496

I think we should keep this for now and start refactoring once the tests are actually green.

GTG/core/datastore.py
picsel2: Uh, nice! Good catch!

GTG/core/saved_searches.py
picsel2: I have removed it and adapted the test.

> It looks like the code only creates a SavedSearch object when reading from XML. I haven't looked at what happens on the GTK side when we create one, but this concerns me a bit.

I'm not sure what you mean here. Do you mean usages like I mentioned [in the other comment](https://github.com/getting-things-gnome/gtg/pull/1036#discussion_r1517246168)?

GTG/core/saved_searches.py
SqAtx: Because it does 2 things that are already done elsewhere: it creates a SavedSearch (which is the responsibility of the SavedSearch class) and it adds it to the store (which is what the `add()` method already does).

So I'd rather not have the duplication.

> I think we should keep this for now and start refactoring once the tests are actually green.

That's fair enough. This is not strictly related to what you're doing in this PR, so we can do it separately.

GTG/core/saved_searches.py
SqAtx: Oh yeah, I missed that. Please ignore my comment ^^

GTG/core/saved_searches.py
picsel2: I have reverted it partially and kept it for later reference: https://github.com/getting-things-gnome/gtg/pull/1036/commits/8d8ffcdf3b82c207947163c87a14c55ad4467836

diegogangl:
Thank you for working on this! Great to see the checks green again :)

PR #1037, opened 2024-03-01
Append short commit hash in development profile 

TheEvilSkeleton:
This explicitly checks the command before assigning to the `version` variable.

This also addresses the following warning during building:

```
WARNING: You should add the boolean check kwarg to the run_command call.
It currently defaults to false,
but it will default to true in future releases of meson.
See also: https://github.com/mesonbuild/meson/issues/9300
```

### DIFF
--- a/GTG/core/meson.build
+++ b/GTG/core/meson.build
@@ -1,13 +1,16 @@
-# Find the version number and put it in info.py.
-# Use git --describe if available.
+# Append short commit hash and put it in
+# info.py when using development profile.
+# Use git rev-parse --short HEAD if available.
 
 git = find_program('git', required : false, disabler : true)
-git_description = run_command(git, 'describe', '--dirty')
 
-if not git.found() or git_description.returncode() != 0
-  version = meson.project_version()
+if git.found() and get_option('profile') == 'development'
+    version = meson.project_version() + '-' + run_command(
+            ['git', 'rev-parse', '--short', 'HEAD'],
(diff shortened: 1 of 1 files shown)


diegogangl:
Wouldn't it be better to pass true? I think there's a compelling argument about not ignoring errors (https://github.com/mesonbuild/meson/issues/9300)

picsel2:
> Wouldn't it be better to pass true?

I think this will lead to failing builds with shallow git clones where tags are missing. I expect `git describe` to fail then.

picsel2:
> Should we experiment with it?

Sure! 😁 [This script](https://gist.github.com/picsel2/e49364835585ec7e74e2a1ca2830f952) sets the value to `true` and fails with the message

    GTG/core/meson.build:5:18: ERROR: Command `/usr/bin/git describe --dirty` failed with status 128.

TheEvilSkeleton:
> > Should we experiment with it?
> 
> Sure! 😁 [This script](https://gist.github.com/picsel2/e49364835585ec7e74e2a1ca2830f952) makes a shallow clone, sets the value to `true` and fails with the message
> 
> ```
> GTG/core/meson.build:5:18: ERROR: Command `/usr/bin/git describe --dirty` failed with status 128.
> ```

I made it run another `git` command instead. When using a development profile, it'll append the commit hash, like so:

image

diegogangl:
Great work, thanks!

Issue #1044, opened 2024-03-03
Saved searches problems


kocio-pl:
1. It's possible to add multiple identical saved searches. GTG should check if such name is already saved.
2. Saving searches shows them immediately on the list, but deleting does not work until application restart (trying to delete them more than once raise an error of course).
3. It's possible to save searches with invisible names, for example only whitespace(s) or even "null" search (saving without typing anything). I think it should be not possible to save them or at least give them some visible name.

nekohayo:
Regarding point no.3, I believe the current code validated for non-empty _names_, but not non-empty _queries_ ; in think it should also check for that (and ideally do some more advanced checks for valid syntax or something, but I don't know exactly what criteria to apply there)

nekohayo:
For your problem no2, I bet there is some relationship between this and #1005 (for tags)...

kocio-pl:
Regarding 2, if I add explicit definition for `ds.saved_searches.remove`, the problem is reverse - search is deleted from the list immediately, but after restart it returns (so it's not saved probably):

    def remove(self, search_id: uuid4) -> None:
        """Remove saved search"""
      
        search = self.lookup[search_id]
        self.data.remove(search)
        pos = self.model.find(search)
        self.model.remove(pos[1])

I'm not a coder, so this might be broken in many ways, but this is what I thought might work at least. It's just some input which could help debug the problem.

nekohayo:
Related: #1048

PR #1046, opened 2024-03-03
Check if search is already saved to avoid duplicates 

kocio-pl:
Fixes 1. problem from #1044

Saves search only if this name (=query) is not already saved to avoid duplicate searches.


diegogangl:
LGTM, thanks!

PR #1049, opened 2024-03-04
Switch Dismiss/Undismiss buttons

### LINKED ISSUE #1043: Newly created tasks have "Undismiss" instead of "Dismiss" in their menu
Task menu for a new task contains "Undismiss" item, which makes no sense for me, since it was never dismissed in the first place. The same is true for the first run tutorial tasks.

Selecting this item does nothing.

kocio-pl:
Fixes #1043.

`Dismiss`/`Undismiss` buttons display is currently switched: active tasks have `Undismiss` button, while closed tasks have `Dismiss` button visible.

This is a problem with menu preparing logic - we are REMOVING items here instead of adding them, so we should reverse checking process:
- if state is dismissable and we find `Dismiss` item, all is good, so we should ignore it, not remove it
- the same if state is NOT dismissable and we find `Undismiss` item, all is good, so we should ignore it, not remove it

Simply swaping both item checks allows to remove proper item.

### DIFF
--- a/GTG/gtk/editor/editor.py
+++ b/GTG/gtk/editor/editor.py
@@ -342,7 +342,7 @@ class TaskEditor(Gtk.Window):
             )
             # this might cause problems with localization if it isn't translated
             # here the same
-            if name == (_("Undismiss") if dismissable else _("Dismiss")):
+            if name == (_("Dismiss") if dismissable else _("Undismiss")):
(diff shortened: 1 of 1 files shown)


kocio-pl:
I'm not sure if I really understand why this works, but it works...

kocio-pl:
This seems to be not proper solution.

PR #1050, opened 2024-03-04
Switch Dismissed/Undismissed buttons

### LINKED ISSUE #1043: Newly created tasks have "Undismiss" instead of "Dismiss" in their menu

kocio-pl:
Fixes #1043.

`Dismissed`/`Undismissed` buttons are switched because of reversed assigning value of `self.set_dismissable_in_menu`.

### DIFF
--- a/GTG/gtk/editor/editor.py
+++ b/GTG/gtk/editor/editor.py
@@ -459,15 +459,15 @@ class TaskEditor(Gtk.Window):
         if status == Status.ACTIVE:
             self.donebutton.show()
             self.undonebutton.hide()
-            self.set_dismissable_in_menu(False)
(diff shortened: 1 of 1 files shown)


SqAtx (approved):
Eventually, we can probably get smarter and write something like 

```
dismissable = status == Status.ACTIVE
self.set_dismissable_in_menu(dismissable)
```
which I find both explicit, clear, and short.

But the current change LGTM.

Side note: my spell checker is telling me that the correct word is "dismissible".

diegogangl:
LGTM, thanks!

PR #1054, opened 2024-03-06
Tag regex refinement

### LINKED ISSUE #987: Ignore @ for tags if part of an URL (such as Mastodon user/posts links)
URLs that contain an `@` (mainly from social media platforms, such as one of these for Mastodon: https://wikis.world/@Poslovitch/111121879906550235) and that are pasted in a task's title or description causes the part of the URL after the `@` to be interpreted as a tag (e.g. `@Poslovitch/111121879906550235`).

This is, at least for me, unwanted behavior.

kocio-pl:
Fixes #987

After replacing string `\B` with `(?<!\/|\w)` addresses like https://wikis.world/@announcements (@ inside URL) are not treated as tags any more.


diegogangl:
Looks good, thanks!

PR #1059, opened 2024-03-10
Update the pane's filter when updating the start date of a task

### LINKED ISSUE #1039: Changing Start Date doesn't remove task from Actionable view
This is related to other problems that I've noticed/filed on the megabranch, but I couldn't find a duplicate here. Similar to https://github.com/diegogangl/gtg/issues/21 in that the DB has the correct data, but the UI hasn't been updated.

This is on commit [2ecc997](https://github.com/getting-things-gnome/gtg/commit/2ecc997d3fb3f6d102615dafa83bda765cebd420)

### Steps

* Go to the Actionable tab
* Select a task (single click)
* Click "Start Tomorrow"

or:

* Go to the Actionable tab
* Open a task (double click)
* Change the start date to a date in the future

### Expected

The task disappears from the Actionable view.

This is the behaviour in GTG 0.6.

### Actual

The task is still visible in the Actionable view. I can update the view and make the disappear by:
* switching to (for example) Open and back to Actionable
* marking an Actionable task as done

SqAtx:
Fixes #1039

Updating the start date of a task may influence whether or not it needs to be visible. For example, if you're on the Actionable pane and you set the start date of a task to some time in the future, then the task is no longer actionable and needs to disappear from that view.

Testing by having some tasks in the Actionable pane, then verifying that the following make the task disappear from the pane:
* click on a task and then on Start Tomorrow
* right-click on a task > Set Start Date > "anything in the future"
* double-click on a task > modify "Starts on" to a date in the future > close task editor
* create a new task > without closing the editor, add a child to it > click on child > click on Open Parent

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ flatpak/bundles
 
 # src/ usually has development version of liblarch
 src/
+
+# Python virtual environments
+venv*/
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,3 @@ flatpak/bundles
 
 # src/ usually has development version of liblarch
 src/
-
-# Python virtual environments
-venv*/
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -1036,7 +1036,7 @@ class MainWindow(Gtk.ApplicationWindow):
         # Changing the start date of a task may take it out of the current pane
         # (e.g. setting the start date of an Actionable task to tomorrow)
         # See #1039
-        self.get_pane().task_filter.changed(Gtk.FilterChange.DIFFERENT)
+        self.get_pane().refresh()
 
     def update_start_to_next_day(self, day_number):
         """Update start date to N days from today."""
@@ -1049,7 +1049,7 @@ class MainWindow(Gtk.ApplicationWindow):
         # Changing the start date of a task may take it out of the current pane
         # (e.g. setting the start date of an Actionable task to tomorrow)
         # See #1039
-        self.get_pane().task_filter.changed(Gtk.FilterChange.DIFFERENT)
+        self.get_pane().refresh()
 
 
     def on_mark_as_started(self, action, param):
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -1033,6 +1033,11 @@ class MainWindow(Gtk.ApplicationWindow):
         for task in self.get_pane().get_selection():
             task.date_start = new_start_date
 
+        # Changing the start date of a task may take it out of the current pane
+        # (e.g. setting the start date of an Actionable task to tomorrow)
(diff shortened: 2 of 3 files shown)


GTG/gtk/browser/main_window.py
```diff
@@ -1041,6 +1046,11 @@ def update_start_to_next_day(self, day_number):
         for task in self.get_pane().get_selection():
             task.date_start = next_day
 
+        # Changing the start date of a task may take it out of the current pane
```
SqAtx, line 1049: Note that we're again running into the fact that the same functionality (updating the start date of a task) is implemented in two different methods (`update_start_date()` and `update_start_to_next_day()`).

I'd like to have one call the other instead, but that's not really related to the current change so I didn't include it here.

.gitignore
```diff
@@ -35,3 +35,6 @@ flatpak/bundles
 
 # src/ usually has development version of liblarch
 src/
+
+# Python virtual environments
+venv*/
```
diegogangl, line 40: Minor nitpick, could we put this in a separate commit?

GTG/gtk/browser/main_window.py
```diff
@@ -1033,6 +1033,11 @@ def update_start_date(self, widget, new_start_date):
         for task in self.get_pane().get_selection():
             task.date_start = new_start_date
 
+        # Changing the start date of a task may take it out of the current pane
+        # (e.g. setting the start date of an Actionable task to tomorrow)
+        # See #1039
+        self.get_pane().task_filter.changed(Gtk.FilterChange.DIFFERENT)
```
diegogangl, line 1039: This can also be achieved by calling the `refresh()` method of the task pane:
Like so: `self.get_pane().refresh()`

diegogangl:
Looks good, except for a couple nitpicks and it works. Thanks! There's one part missing though, if the start date is changed in the task editor using the calendar widget. Maybe we should call refresh() when the window is destoryed?

SqAtx:
> There's one part missing though, if the start date is changed in the task editor using the calendar widget. Maybe we should call refresh() when the window is destoryed?

Good point! I actually noticed that earlier this week too, but didn't have time to look into it further. Will do.

.gitignore
SqAtx: Fair enough - moved to #1064

GTG/gtk/browser/main_window.py
SqAtx: Oh nice. Much more readable!

SqAtx:
I also added a `.refresh()` to `TaskEditor.destruction()`. I was originally wondering if I shouldn't add it to `Application.close_task()`, whose docstring is "Close a task editor window."

But `destruction()` is linked to the "close-request" callback of the editor, so I think it makes most sense there.

diegogangl:
Looks good now :+1: 
Merging

Issue #605, opened 2021-04-01
Partial tags and tags duplication into gtg_data.xml when typing tags at various speeds in the TaskEditor


nekohayo:
This is the remainder/successor to issue #557. Essentially, when you type tags "slowly enough for parsing to happen mid-way" in the task editor, especially if you type "back and forth" (ex.: type a partial word, wait, type some more, wait, type too much, wait, backspace a couple of characters, wait and then close the task editor), then you will encounter "intermediate" tags that are not used in the final saved tasks, but are saved as orphan tags in the gtg_data.xml file format:

![2021-03-31-tags-duplication-on-type](https://user-images.githubusercontent.com/479401/113230475-0d752580-9267-11eb-9bf0-499b68ff39fc.png)

Or another example from the last comment of https://github.com/getting-things-gnome/gtg/issues/557#issuecomment-788102087 :

![2021-03-01-duplication-still-happens-in-the-data-file](https://user-images.githubusercontent.com/479401/109530247-7f99f500-7a84-11eb-8ec4-5e01dc8cacb0.png)

Fixing this is not super trivial as, according to @diegogangl:

> It gets more convoluted because tags are sometimes tags, and sometimes strings; there's the special tags, but saved searches are also tags and "special" as their parent is "search", but they are not really considered special in the code though

What are "special" tags if they are not searches?
"All tasks", "Tasks with no tags", and... "separator"!

Solving this might depend on some refactoring (and ideally, tests!), so at this point this is probably 0.6 material rather than something to be fixed while on the verge of releasing 0.5.

nekohayo:
@gycsaba96's #1062 tried to fix this, but the fix kind mostly "changed the behavior". There are some issues (arguably regressions) from that change:

* Whenever there is no space after a tag, it will no longer recognize it as a tag, and it will immediately remove it
* Empty tags can still get created and saved; while they immediately get removed from the sidebar, they get saved to the XML data file and will reappear on startup, i.e. the reverse of #1007...

Demonstration video with current git `master`, where you can see several unexpected behaviors if you look closely:

[2024-03-16 GTG task editor tags saving smoke-testing.webm](https://github.com/getting-things-gnome/gtg/assets/479401/eaf8a1b1-58cb-4b99-a89c-2e3bcf3dff7e)

gycsaba96:
>  Whenever there is no space after a tag, it will no longer recognize it as a tag, and it will immediately remove it

[...]

Issue #1058, opened 2024-03-09
Extra tags after tag renaming


gycsaba96:
Renaming a tag that is a prefix of another tag adds unwanted tags to other tasks.

Steps to reproduce:
1. Create a task with an `@apple` tag.
2. Create a task with an `@applepie` tag.
3. Rename the `@apple` tag using the sidebar's _"Edit..."_ feature to `@orange`.
4. Open the second task and observe the unwanted `@orangepie` tag.

The main cause is possibly that the [corresponding method](https://github.com/getting-things-gnome/gtg/blob/2ecc997d3fb3f6d102615dafa83bda765cebd420/GTG/core/tasks.py#L361) uses a simple search-and-replace for tag renaming. (The code should ensure that it replaces only whole tag names.)

nekohayo:
Hmm, I wonder if this might be an explanation for the long-standing super annoying tags duplication issue, #605 ?

It would be awesome if you could try fixing both your issue and that one (if they are not exactly the same) simultaneously :)

gycsaba96:
I think the unwanted tags appear for different reasons in the two cases. Here, we have a faulty renaming. In #605, the tag parsing kicks in too early. (There, we could consider preventing the processing of the current word [when parsing the tags](https://github.com/getting-things-gnome/gtg/blob/083353fd9160c7afb05b8ea2925ce61589b08b80/GTG/gtk/editor/taskview.py#L222).)

Nevertheless, I can take a look at both of them. :)

nekohayo:
Does this also get fixed by your PR #1062, or that one was only meant for 605?

gycsaba96:
No, this is still a work in progress. PR 1062 was only for 605.

PR #1062, opened 2024-03-14
Prevent parsing the current word as a tag when editing a task

gycsaba96:
This fixes issue #605 (as promised in #1058).

- It solves the problem by masking out the currently edited word to prevent parsing half-typed tags. 
- It also ensures to parse everything when closing the window. 
- (Note that this does not eliminate already existing partial tags in the `gtg_data.xml`.)


diegogangl:
This feels great when editing, kudos! 
Merging, thanks

PR #1063, opened 2024-03-14
Add parent to `self.add`

TheEvilSkeleton:
Since the default value for `UUID` is `None`, we can safely avoid unnecessarily checking if `UUID` exists.


diegogangl:
Nice catch, thanks

PR #1064, opened 2024-03-17
Add standard Python virtualenv names to gitignore

SqAtx:
I do my dev work using virtual environments, like venv/ or venv38/ to test different versions of Python.

This removes them and the many files they contain, from git.


diegogangl:
Thanks, merging :+1:

PR #1065, opened 2024-03-17
Add simple pre-commit hook

SqAtx:
Relates to #237

This adds a simple configuration for [pre-commit](https://pre-commit.com) and runs it on GHA. It doesn't do very much; mostly removing trailing whitespaces.

My next step would be to add pylint or pyflakes to this, and tell it to ignore every warning that GTG currently breaks. This would:
* prevent us from breaking any more
* allow us to fix the warnings one at a time, when we can

I tested the GHA on a fork.

This commit doesn't contain the changes that fix the pre-commit check, so GHA will fail. To fix it, we'll just need to run `pre-commit run --all-files` and commit the result.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,29 @@ No data should be lost since it is just re-generateable build files.
 [pudb]: https://pypi.org/project/pudb/
 [gtk-interactive]: https://developer.gnome.org/gtk3/stable/gtk-running.html#interactive-debugging
 
+## Setting up a development environment
+
+Now that you have installed all the dependencies and know how to run the tests,
+it's possible that you would like to make code changes and submit them for
+review.
+
+We use [pre-commit](https://pre-commit.com) to run some checks that ensure that
+our codebase is clean and consistent.
+
+Install it with `pip install pre-commit`, and install the pre-commit hooks with
+`pre-commit install`.
+
+The checks will now run every time you make a commit, and fix files if
+necessary. It's generally a good idea to always run them, but you can also
+bypass them by running `git commit --no-verify`, if necessary.
+
+The CI system runs the checks on each commit to master, and on each pull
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -160,6 +160,21 @@ No data should be lost since it is just re-generateable build files.
 [pythondevmode]: https://docs.python.org/3/library/devmode.html
 [pudb]: https://pypi.org/project/pudb/
 
+## Setting up a development environment
```
SqAtx, line 163: This is still WIP.
I'd like having a section like this, but it isn't trivial to integrate it into the current documentation.

SqAtx:
@diegogangl @nekohayo what do you guys think of this one?

diegogangl:
Hey, thanks for working on this! LGTM, though I'd split it at least the whitespace changes from the yaml changes in two different commits. Also, looks like I caused some conflict here (hopefully just whitespace stuff)

SqAtx:
Yep that' s a good point. I'll rebase this with just my changes. After you merge it, you can just run pre-commit and make a new commit that will fix the build.

I also just noticed that https://github.com/getting-things-gnome/gtg/blob/master/Makefile is a thing, and we probably want to eventually move pep8/pyflakes out of it and into pre-commit. I think between meson and pre-commit, that makefile will be pretty much redundant?

azmeuk:
Hi @SqAtx 
For your future step, a python linter that gathers attention lately is [ruff](https://docs.astral.sh/ruff/). It can serve as a replacement for pylint and pyflakes, and boast to be faster than both.

SqAtx:
I've never used it but I don't mind trying it! It seems pretty young but has quite a few developers behind it. And it integrates with pre-commit, so why not.

PR #1066, opened 2024-03-17
Rework About dialog

TheEvilSkeleton:
This makes a few changes and additions:
- This adds a new dependency on libportal. It's used to detect if it's sandboxed.
- Various `GtkButton`s are added to provide external links.
- The Credits section has been reworked to prevent the about dialog from increasing.
- New System section was added to help users copy and paste system information for troubleshooting purposes.
- The about dialog is now a modal window.

Screenshots:

| About | System |
|--------|--------|
| image | image |

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -20,7 +20,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         sudo apt-get update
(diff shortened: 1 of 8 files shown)


TheEvilSkeleton:
Sure

diegogangl:
LGTM in general. I would remove the "donate" button though, at least until we make a GTG account in Liberapay and/or find a way to split donations among the top contributors

PR #1071, opened 2024-03-23
Remove inline references when unparenting a task

### LINKED ISSUE #1029: After reparenting children via drag and drop, doing "Open Parent" twice causes a `KeyError` traceback, can lead to data file corruption from base_store
Possibly a closely related issue to #1018, but the symptom and reproduction steps seems different.  
However the steps are very consistent and easy to reproduce, so this might be quite useful for diagnosis by @diegogangl or anyone else.

---

With git `master`:

1. `rm -R tmp/default/ && ./launch.sh`; this will launch with the clean default dataset (i.e. demo/tutorial tasks)
2. Unfold the main task
3. Drag one of the child tasks (ex: "Learn About Plugins") to become a child of another (ex: "Reporting Bugs")
4. Double-click "Learn About Plugins" (now a child of "Reporting Bugs")
5. Click "Open Parent" once, everything is fine
6. Click "Open Parent" a second time to go up one more level

Result: Global generic exception below.

What's worse is that this apparently(?) corrupts the data file; on subsequent startup, if you open "Reporting Bugs" and click to "Open Parent" only once, you will get the same traceback.

```python-traceback
  File "GTG/gtk/application.py", line 444, in open_parent_task
    editor.open_parent()
  File "GTG/gtk/editor/editor.py", line 805, in open_parent
    self.app.open_task(parent)
  File "GTG/gtk/application.py", line 558, in open_task
    editor = TaskEditor(app=self, task=task)
  File "GTG/gtk/editor/editor.py", line 192, in __init__
    self.textview.insert(text)
  File "GTG/gtk/editor/taskview.py", line 788, in insert
    self.process()
  File "GTG/gtk/editor/taskview.py", line 228, in process
    self.delete_subtask_cb(tid)
  File "GTG/gtk/editor/editor.py", line 762, in remove_subtask
    self.app.ds.tasks.unparent(tid, self.task.id)
  File "GTG/core/tasks.py", line 936, in unparent
    super().unparent(item_id, parent_id)
  File "GTG/core/base_store.py", line 165, in unparent
    raise KeyError
KeyError
```

gycsaba96:
Fixes #1029

When unparenting a task, inline references of the old subtask must be removed from the old parent's content.
(Otherwise, the reference is parsed again and removed unsuccessfully, causing the crash described in the issue.)

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -939,6 +939,9 @@ class TaskStore(BaseStore):
         item = self.lookup[item_id]
         parent = self.lookup[parent_id]
 
+        # remove inline references to the former subtask
+        parent.content = re.sub(r'\{\!\s*'+item_id+'\s*\!\}','',parent.content)
+
         self.model.append(item)
         parent.notify('has_children')
 


diegogangl:
Looks good, thanks!

PR #1076, opened 2024-03-28
Avoid deleting an unedited new task twice

### LINKED ISSUE #1026: KeyError Traceback on closing an unmodified new task editor
Post GTK4 port, in git `master`:

1. Click the `+` button to create a new Task Editor window
2. Without making any changes, press the `Escape` key

```python-traceback
  File "GTG/gtk/editor/editor.py", line 889, in destruction
    self.app.ds.tasks.remove(tid)
  File "GTG/core/tasks.py", line 914, in remove
    item = self.lookup[item_id]
           ~~~~~~~~~~~^^^^^^^^^
KeyError: UUID('d1313822-295b-478b-932c-19ea1c7b4178')
```

What is strange is that the "traceback catcher" error handling GUI doesn't even catch this traceback...

gycsaba96:
Fixes #1026

The crash happened because the program tried to delete the same unmodified new task twice. After hitting `Escape`, the `TaskEditor`'s `close` method was called. This override called `self.destruction()`, thus deleting the new task. After that, `super().close()` was also called emitting a `close-request` signal. This `close-request` signal was also [connected](https://github.com/getting-things-gnome/gtg/blob/b2dec14ca98db5453bd74ca524ded6d0a0a8bc3e/GTG/gtk/editor/editor.py#L241) to the `destruction` method, thus trying to delete the same task the second time.

The `TaskEditor`'s `close()` override is now removed. Its only function was to call the `destruction()` method. However, the `destruction()` method was already connected to the `close-request` signal.

### DIFF
--- a/GTG/gtk/editor/editor.py
+++ b/GTG/gtk/editor/editor.py
@@ -864,15 +864,6 @@ class TaskEditor(Gtk.Window):
             elif focus == self.closed_entry:
                 self.show_popover_closed()
 
-    # We define dummy variable for when close is called from a callback
-    def close(self, action=None, param=None):
-
-        # We should also destroy the whole taskeditor object.
-        if self:
-            self.destruction()
-            super().close()
-            self = None
-
 
     def is_new(self) -> bool:
         return (self.task.title == DEFAULT_TITLE 


diegogangl:
Great catch, thanks!

PR #1082, opened 2024-04-05
Highlight title when creating a new task

### LINKED ISSUE #1055: Highlight "New Task" text when creating a new task
So that if you start typing immediately, it writes the task title.

### Steps

* Click on the + on the top left corner
* It opens the task editor with a task named "New Task" and no task body

### Expected

The text "New Task" is highlighted. When you start typing, "New Task" gets replaced by what you're typing.

This is the behaviour in GTG 0.6.

### Actual

The cursor is on the next line. When you start typing, you're writing the body of the task.

This is on current master, https://github.com/getting-things-gnome/gtg/commit/2ecc997d3fb3f6d102615dafa83bda765cebd420

SqAtx:
Fixes  #1055

When you create a new task, the default title is selected so that when you start typing, you write the title of the task.

### DIFF
--- a/GTG/gtk/editor/editor.py
+++ b/GTG/gtk/editor/editor.py
@@ -209,10 +209,8 @@ class TaskEditor(Gtk.Window):
             if sub.id not in self.textview.subtasks['tags']:
                 self.textview.insert_existing_subtask(sub)
 
-        # if thisisnew:
-        #     self.textview.select_title()
-        # else:
-            # self.task.set_to_keep()
+        if self.is_new():
+            self.textview.select_title()
 
         self.textview.buffer.end_irreversible_action()
 


GTG/gtk/editor/editor.py
```diff
@@ -209,10 +209,8 @@ def __init__(self, app, task):
             if sub.id not in self.textview.subtasks['tags']:
                 self.textview.insert_existing_subtask(sub)
 
-        # if thisisnew:
-        #     self.textview.select_title()
-        # else:
-            # self.task.set_to_keep()
```
SqAtx, line 215: I'm not sure what this was doing, and since the method no longer exists, I figured we probably don't need to keep this line around.

diegogangl: Definitely, it was related to liblarch's way of managing things. I think it would delete the task if it didn't have the else block? I also can't remember now

diegogangl:
Merged, thanks!

Issue #1077, opened 2024-03-28
Reopening a task with children adds a blank line before the list of children


SqAtx:
Steps:
* Create task with the following text:
```
Parent
- child
```
* close the task
* double-click on the task
* the text is now:
```
Parent

- child
```
You can repeat that and add more lines.

---

```
Parent
text
- child
```
becomes
```
Parent
text

- child
```
and
```
Parent
- child1
text
- child2
```
becomes
```
Parent
text

- child1
- child2
```

PR #1083, opened 2024-04-06
Write unit test for TaskView.get_title()

SqAtx:
I started to look at #1077 and I thought it would be nice to be able to unit test this type of thing.

A simple unit test for TaskView would be

* create a task
* create a view for it
* check that the title of the view is the title of the task

In the current state of affairs, this is difficult because the code that puts the Task data into the view lives in TaskEditor. But we pass a Task to the constructor of TaskView, so TaskView actually has all the data it needs. So I moved the logic there. I also moved the `is_new()` logic from the TaskEditor down to the Task.

I feel like it should be possible to eventually do this in the TaskView constructor but for now, TaskEditor calls `self.textview.set_text_from_task()` from the same place it used to run that code.

The test triggers a segmentation fault when running on GitHub Actions, so it's skipped there while I figure out why.

### DIFF
--- a/GTG/gtk/editor/editor.py
+++ b/GTG/gtk/editor/editor.py
@@ -39,7 +39,7 @@ from GTG.gtk.editor.calendar import GTGCalendar
 from GTG.gtk.editor.recurring_menu import RecurringMenu
 from GTG.gtk.editor.taskview import TaskView
 from GTG.gtk.colors import rgb_to_hex
-from GTG.core.tasks import Task, Status, DEFAULT_TITLE
+from GTG.core.tasks import Task, Status
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


GTG/gtk/editor/editor.py
```diff
@@ -179,40 +168,7 @@ def __init__(self, app, task):
         textview_focus_controller.connect("leave", self.on_textview_focus_out)
         self.textview.add_controller(textview_focus_controller)
 
-        tags = task.tags
-        text = self.task.content
-        title = self.task.title
-
-        # Insert text and tags as a non_undoable action, otherwise
```
SqAtx, line 186: This block of code has moved to the TaskView class. The logic hasn't changed besides the way the objects are named (`self.insert()` instead of `self.textview.insert()` etc) and the fact that the TextView calls `is_new()` on the task object.

tests/core/test_taskview.py
```diff
@@ -41,3 +44,15 @@ def test_no_detect_tags(self):
         matches = re.findall(TAG_REGEX, content)
 
         self.assertEqual([], matches)
+
+    def test_get_title(self):
+        task_title = 'Very important task'
+        task = Task(id = uuid4(), title=task_title)
+        view = TaskView(Datastore(), task, None, False)
```
SqAtx, line 51: These 3 lines can probably be encapsulated into a `create_task_view_from_task()` test helper, but I can do that when I add another test.

SqAtx:
Ugh okay, let's not merge that yet... GHA fails with "Fatal Python error: Segmentation fault", which is going to be a joy to debug.

SqAtx:
OK so I have confirmed that this is caused by `test_get_title()` - skipping it makes the tests green again. Trying Python 3.9 doesn't, and I've only seen it o GHA so far.

SqAtx:
I think this is worth merging even though I haven't solved the segfault issue (it anyone wants to look: https://github.com/getting-things-gnome/gtg/actions/runs/8864137269/job/24339006474?pr=1083), and even though  it looks like the bug I set out to fix will be fixed by #1102

I'll work on unskipping the test separately.

gycsaba96:
I managed to reproduce the segfault outside of the GitHub Action using a small server VM. The main problem seems to be that GA runs without any graphical display features, and the new GUI test wants to use these features. If you provide a fake display server, everything works fine:

```
xvfb-run python3 -m pytest
```

SqAtx:
Thanks a lot for looking into it! That would have taken me a while to figure out.

It feels a bit like a hack somehow? But I can't really articulate why, nor do I have a better way to fix the problem =D

gycsaba96:
> It feels a bit like a hack somehow? But I can't really articulate why, nor do I have a better way to fix the problem =D

I think we can defend the move:
- The test ensures that a GUI component is correctly displayed. Using a display server during the process sounds reasonable.
- Moreover, this would only change the GHA workflow without limiting anything else.

diegogangl:
Hi @SqAtx, this LGTM. Is this ready to merge or is there something else to do in this branch?

SqAtx:
@diegogangl It's ready to merge! It's a good first step.

diegogangl:
Thanks!

PR #1085, opened 2024-04-13
Use Date.no_date() instead of None for Task.date_due

### LINKED ISSUE #1068: TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'
(this report is a bit rushed because it happened while I was in the middle of something)

Not 100% sure what I was doing.

I clicked on a tag, created an issue (tag got added automatically), set it due to "soon", then toggled between Open and Actionable because the sidebar was showing 3 issues for that tag, but the Actionable pane showed 2 and the Open pane showed 0 (not even the one I just created).

Then I got the "GTG has crashed dialog" but clicking on the "Details" link did not work (might be a separate issue there). I found this in the logs:
```
Gtk-Message: 10:32:20.575: GtkDialog mapped without a transient parent. This is discouraged.
  File "<< path to GTG >>/.local_build/install/lib/python3/dist-packages/GTG/core/sorters.py", line 73, in do_compare
    if first > second:
       ^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'
```

gycsaba96:
Fixes #1068

In some cases, `None` was used instead of `Date.no_date()`, which caused a crash when sorting the task list. These changes follow the already existing type hints and comments.

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -414,7 +414,7 @@ class Task(GObject.Object):
             Return a tuple of (bool, Date)
             """
             if recurring_term is None:
-                return False, None
+                return False, Date.no_date()
 
             try:
                 # If a start date is already set,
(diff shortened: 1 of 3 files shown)


diegogangl:
Nice catch, thanks!

PR #1086, opened 2024-04-14
Fix check for empty excerpts in popups

### LINKED ISSUE #1010: When a task in the Task Browser has no description to show, it shows "…" in the tooltip
Since the GTK4 port.

Shouldn't it show no tooltip at all in this case?

diegogangl:
If the content had some whitespace the if-guard would fail, creating a tooltip with just "...".
It would also fail because it would include tags and subtasks. So now we first process the content properly and then check.

Also fixes the regex to remove tags in a comma list properly from the excerpt

fixes #1010

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -307,16 +307,19 @@ class Task(GObject.Object):
 
     @GObject.Property(type=str)
     def excerpt(self) -> str:
-        if not self.content:
-            return ''
-
         # Strip tags
         txt = TAG_REGEX.sub('', self.content)
 
         # Strip subtasks
         txt = SUB_REGEX.sub('', txt)
 
-        return f'{txt.strip()[:80]}…'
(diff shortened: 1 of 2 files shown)


SqAtx (commented):
The change itself looks good to me.

> If the content had some whitespace the if-guard would fail, creating a tooltip with just "...".
> It would also fail because it would include tags and subtasks. So now we first process the content properly and then check.
> 
> Also fixes the regex to remove tags in a comma list properly from the excerpt

Could we add tests for these cases? Ideally break the existing `test_excerpt()` into one test for each case (empty task, task with only whitespace, test that tags are ignored, etc)

This type of functionality is just perfect for unit tests: it's just text in / text out.

diegogangl:
Good idea, thanks!
Added the tests

SqAtx (approved):
Nice!

PR #1087, opened 2024-04-14
Save sort setting in config

### LINKED ISSUE #353: Task browser's treeview columns sorting order is not remembered in the "Actionable" and "Closed" view modes
If you change the sorting by due date to have the more urgent tasks at the top instead of the bottom (which should probably be the default anyway), when you restart GTG you can see that it was remembered in the "Open" tasks view, but not the "Actionable" or "Closed" tasks.

I was thinking that all three should be controlled by a single sorting setting but then maybe some users won't like that, or particularly, the views offer different sorting columns (the Actionable view doesn't show the start dates for example)...

### LINKED ISSUE #1020: Sorting order not remembered & restored on startup
Since the GTK4 port, this is not automatically remembered and restored between app restarts.

GTG should store that info in realtime (so that it is saved even if the app crashes).

Related, the bug that affected previous versions (so fixing this here should fix it too): #353

diegogangl:
Save sort configurations and reload on startup

Also remove a duplicated method

closes #1020
closes #353

### DIFF
--- a/GTG/core/config.py
+++ b/GTG/core/config.py
@@ -50,6 +50,7 @@ DEFAULTS = {
         'autoclean_days': 30,
         'dark_mode': False,
         'maximized': False,
+        'sort_mode': 'title',
     },
     'tag_editor': {
         "custom_colors": [],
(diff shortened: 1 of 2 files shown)


SqAtx:
I think this fixes #1020 but not #353
It remembers the sort setting, but if I go to the Actionable pane, the order is different.

PR #1089, opened 2024-04-16
Fix alignment in hierarchical views

### LINKED ISSUE #1042: Expanders cause alignment problems in hierarchical tags list in Sidebar and tasks list in Task Browser
Tag names and icons for free, parent and child tags should be aligned, but currently they have problems:
- parent tags should be probably aligned with free tags
- leaf child tags should have some indentation compared to their parent tags
- nested parent tags should probably also have a better indentation compared to their parent tags

Zrzut ekranu z 2024-03-03 06-12-53

gycsaba96:
Fixes #1042

The new code leverages `GtkTreeExpander`'s ability to handle indentation both for tags in the sidebar and tasks in the task list. I also created a `children_count` property for the Tag class so that the expander can bind to it.

### DIFF
--- a/GTG/core/tags.py
+++ b/GTG/core/tags.py
@@ -80,6 +80,13 @@ class Tag(GObject.Object):
         return self.id == other.id
 
 
+    @GObject.Property(type=int)
+    def children_count(self) -> str:
+        """Read only property."""
+
+        return len(self.children)
+
+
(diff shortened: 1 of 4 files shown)


diegogangl:
Great work, this feels much better and the code leans on bindings so we won't have all those issues with interface updates.
Thanks!

PR #1090, opened 2024-04-16
Remove tests from checks

### LINKED ISSUE #838: Make check fails on a permissions error to nonexistent config file
Since I've been running the flatpak version for stable, my config files for GTG are not in the main XDG config folder, and apparently this causes the test suite to fail when launching `make check` in the git version:

```
_________ ERROR collecting tests/core/test_versioning.py _________
tests/core/test_versioning.py:21: in <module>
GTG/gtk/browser/adaptive_button.py:44:30 undefined name 'ChildItem'
GTG/gtk/browser/adaptive_button.py:47:39 undefined name 'ChildItem'
    from GTG.core import versioning
GTG/core/versioning.py:29: in <module>
    from GTG.core import datastore
GTG/core/datastore.py:43: in <module>
    class DataStore():
GTG/core/datastore.py:52: in DataStore
    def __init__(self, global_conf=CoreConfig()):
GTG/core/config.py:193: in __init__
    self._task_conf = open_config_file(self._task_conf_path)
GTG/core/config.py:81: in open_config_file
    open(config_file, "w").close()
E   PermissionError: [Errno 13] Permission denied: '/home/user/.config/gtg/tasks.conf'
```

diegogangl:
Running tests locally has become a bit more involved recently. We can keep them out of "make check" for now, so that we can at least get some basic style checks running again.

Fixes #838

(actually the error in that issue was fixed with the new core, but there was a new error related to `run-tests`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@
 PEP8=pycodestyle
 PYFLAKES=pyflakes
 
-check: tests pep8 pyflakes
+check: pep8 pyflakes
 
 # Run all of the tests.
 tests:


SqAtx (approved):
LGTM
As I mentioned in #1065, it looks like the functionality of the Makefile is going to move to pre-commit and meson, so we can probably remove it eventually? We'll have to update the documentation that references it.

diegogangl:
Hmm yeah maybe. I don't know enough about Meson to tell if all the targets in the make file are doable in there TBH

Issue #1018, opened 2024-02-27
Reparenting children tasks before and after switching between `Open` and `Actionable` views will lead to a freeze hang


nekohayo:
As discovered while reporting #1017:

The video below shows an issue with the completed GTK4 port, using the default dataset, where you can eventually make GTG freeze/hang.

The only relevant line in the terminal is this:

```
Gtk-CRITICAL **: gtk_widget_set_parent: assertion '_gtk_widget_get_parent (widget) == NULL' failed
```

[GTG-GTK4-reparenting-deselects-sidebar-tag-and-eventually-hangs.webm](https://github.com/getting-things-gnome/gtg/assets/479401/7c3c1b72-bba1-4885-84d7-6307c585b132)

diegogangl:
None of these are happening anymore in Master (`291d49cabb4f4b104e9c9a3eb1adcf7b8e021072`)

PR #1094, opened 2024-04-24
Update models for subtasks upon changes

gycsaba96:
This fixes one of the problems listed in GitHub issue #1018:

> If you drag a child from one parent to another task (re-parent), it duplicates it in the view

- The `tid_to_subtask_model` field keeps track of the `ListStore`s created for subtasks. 
- The `_remove_from_parent_model` and `_append_to_model` helper functions update these models when a change happens. 
- Thanks to this new functionality, some GUI updates are simplified.


diegogangl:
Great job! Now we can get rid of that awful collapse/expand hack to update the ui :)
Thanks, merging

PR #1095, opened 2024-04-30
Create a dummy button in the sidebar

### LINKED ISSUE #1017: Sidebar tag gets deselected, view resets to "All Tasks" when dragging a child (subtask) onto another task in "Open" tasks view mode
Since the GTK4 port, 

If you currently have:
* a (sub)tag selected in the sidebar,
* have the "Open" tasks view selected

...and drag a subtask onto another (to make it a child)...

...then the view will be reset to "All tasks" (it will deselect the tag).

This video shows the issue with the default dataset, in addition to a nastier issue where you can eventually make it freeze/hang (issue #1018):

[GTG-GTK4-reparenting-deselects-sidebar-tag-and-eventually-hangs.webm](https://github.com/getting-things-gnome/gtg/assets/479401/7c3c1b72-bba1-4885-84d7-6307c585b132)

gycsaba96:
Fixes #1017 and the following points of #1018:
> - If you drag a child out of a parent, the selected tag in the sidebar resets to "All Tasks" view
> - If you drag a standalone task into a parent (turning the task into a child), the selected tag in the sidebar resets to "All Tasks" view

Based on [a forum discussion](https://discourse.gnome.org/t/list-box-first-row-is-always-selected/14840/2), if the currently focused widget is removed, Gtk.Window selects the first suitable children. This is the "All Tasks" row in our case, thus causing a view reset. Other cases can also reproduce the anomaly: reparenting a tag, deleting a task, etc.

Now, an invisible button catches the focus instead of the "All Tasks" row. Thanks to this button, unintended view resets can be avoided when the focused widget is removed.

Theoretically, we could define a focus grab for each of these special cases. However, we would risk that we miss one and the bug comes back. Moreover, this change doesn't affect the complexity of other code parts.

### DIFF
--- a/GTG/gtk/browser/sidebar.py
+++ b/GTG/gtk/browser/sidebar.py
@@ -74,6 +74,12 @@ class Sidebar(Gtk.ScrolledWindow):
         wrap_box.set_vexpand(True)
         wrap_box.set_hexpand(True)
 
+        # Create an invisible dummy button to catch focus in some edge cases
+        self.dummy = Gtk.Button()
+        self.dummy.set_size_request(0, 0)
+        self.dummy.add_css_class('dummy')
+        wrap_box.append(self.dummy)
+
         # -------------------------------------------------------------------------------
         # General Filters
         # -------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


diegogangl:
LGTM, thanks for looking into this!
I'll mark the items on #1018 that are done

PR #1100, opened 2024-05-11
Always show tags with zero tasks

### LINKED ISSUE #1007: Tags with zero issues are not shown on startup, but get shown after switching "Actionable" and "Open" view modes
Since the GTK4 port:

Zero-tasks rogue/duplicate tags can all show up at the bottom of the sidebar now (this might be considered a feature rather than a bug, I suppose!), however they don't show up in "Open" tasks mode initially… they only show up when you switch to "Actionable" mode and back to "Open" mode.

gycsaba96:
Fixes #1007

The anomaly was caused by a lucky bug. `TagEmptyFilter` was checking for the wrong `self.pane` values except when `Sidebar.__init__` created the object, thus filtering the tags with zero tasks only after startup.  

Now, `TagEmptyFilter` checks for the correct `self.pane` values and considers the newly introduced `show_zero` field. (This field could be used in the future to show and hide tags with zero tasks.)

### DIFF
--- a/GTG/core/filters.py
+++ b/GTG/core/filters.py
@@ -42,19 +42,20 @@ class TagEmptyFilter(Gtk.Filter):
         super(TagEmptyFilter, self).__init__()
         self.ds = ds
         self.pane = pane
+        self.show_zero = True
 
 
     def do_match(self, item) -> bool:
         tag = unwrap(item, Tag)
 
-        if self.pane == 'open':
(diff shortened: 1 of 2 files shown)


diegogangl:
Ah yes, I recently hit in that difference in pane names again in another branch. We should fix that in the future
LGTM and merging, thanks!

PR #1101, opened 2024-05-15
Inherit the parent's tags when creating a new subtask

### LINKED ISSUE #1078: Children's tasks no longer inherit their parent's tags
Create a task with the following content:
```
parent
@mytag

- child
```
`child` become clickable. Click on it.
The child task contains
```
child

```
In 0.6, it contains
```
child
@mytag
```

gycsaba96:
Fixes #1078

If a parent is provided during task creation, all tags of the parent are added to the newly created task.

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -763,6 +763,11 @@ class TaskStore(BaseStore):
 
         self.add(task, parent)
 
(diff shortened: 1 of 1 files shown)


diegogangl:
Works as advertised, thanks!

PR #1102, opened 2024-05-18
Fix the mixed-use of UUID and str for task IDs

### LINKED ISSUE #1077: Reopening a task with children adds a blank line before the list of children

gycsaba96:
Fixes #1077

The cause of the (seemingly) extra empty line was that existing subtasks were [not correctly detected](https://github.com/getting-things-gnome/gtg/blob/b32ec933ba61ba2c6e3de563c5fa79dc3abd1f61/GTG/gtk/editor/taskview.py#L792) and thus moved to the end of the text. This happened because of the mixed use of strings and UUIDs for task IDs.

This PR converts strings to UUIDs when necessary and fixes the related type hints.

### DIFF
--- a/GTG/core/base_store.py
+++ b/GTG/core/base_store.py
@@ -49,7 +49,7 @@ class BaseStore(GObject.Object):
         raise NotImplemented
 
 
-    def get(self, key: str) -> Any:
+    def get(self, key: UUID) -> Any:
         """Get an item by id."""
 
         return self.lookup[key]
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -753,7 +753,7 @@ class TaskStore(BaseStore):
         return new_task
 
 
-    def new(self, title: str = None, parent: UUID = None) -> Task:
+    def new(self, title: str = None, parent: Optional[UUID] = None) -> Task:
         """Create a new task and add it to the store."""
 
         tid = uuid4()
@@ -777,7 +777,7 @@ class TaskStore(BaseStore):
         elements = list(xml.iter(self.XML_TAG))
 
         for element in elements:
-            tid = element.get('id')
+            tid = UUID(element.get('id'))
             title = element.find('title').text
             status = element.get('status')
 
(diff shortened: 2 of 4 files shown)


SqAtx:
Thank you for looking at it! I kind of got bogged down in #1083 on my way to add tests for this part of the code ^^'

I see that this fixes the bug, but I don't understand why. How were the subtasks not correctly detected? Weren't they detected by `line.lstrip().startswith('{!'):`, which would be the same whether we use UUID or str?

gycsaba96:
> Thank you for looking at it! I kind of got bogged down in #1083 on my way to add tests for this part of the code ^^'

Don't give up, QA is very important. :) 

> I see that this fixes the bug, but I don't understand why. How were the subtasks not correctly detected? Weren't they detected by `line.lstrip().startswith('{!'):`, which would be the same whether we use UUID or str?

Let me provide more details.
1. When the subtask is created, it is stored in the task store's `lookup` dictionary with a `UUID` key.
2. Line [772](https://github.com/getting-things-gnome/gtg/blob/b32ec933ba61ba2c6e3de563c5fa79dc3abd1f61/GTG/gtk/editor/taskview.py#L772) and [777](https://github.com/getting-things-gnome/gtg/blob/b32ec933ba61ba2c6e3de563c5fa79dc3abd1f61/GTG/gtk/editor/taskview.py#L777) extract and store the subtask's id as a string.
3. Line [791](https://github.com/getting-things-gnome/gtg/blob/b32ec933ba61ba2c6e3de563c5fa79dc3abd1f61/GTG/gtk/editor/taskview.py#L791) raises a `KeyError` because we want to do the lookup using a string key, although the subtask does exist in the store with a `UUID` key. (Sorry for referencing the wrong line.)
4. As a consequence, it remains in the `subtasks` list and gets erased from the text by the `for` loop at line [801](https://github.com/getting-things-gnome/gtg/blob/b32ec933ba61ba2c6e3de563c5fa79dc3abd1f61/GTG/gtk/editor/taskview.py#L801). This is how it is removed from the original position.
5. Finally, it gets added back to the end of the text by [editor.py#210](https://github.com/getting-things-gnome/gtg/blob/b32ec933ba61ba2c6e3de563c5fa79dc3abd1f61/GTG/gtk/editor/editor.py#L210) when you reopen the editor.

diegogangl:
Merged, thanks!

SqAtx:
@gycsaba96 ahhh I see, thank you! I haven't quite wrapped my head around the entire task/editor logic yet

Issue #1093, opened 2024-04-22
Infinite loop of errors under GTK<4.10 after recent commit


SqAtx:
After https://github.com/getting-things-gnome/gtg/commit/23a986847a61f4744a1e9a323a4f5b353358d2e5, opening GTG results in some sort of infinite loop of errors that spawns the "something unexpected happened" dialog continuously and makes it difficult to close GTG.

This is the end of my logs:
```
Gtk-Message: 08:52:07.152: GtkDialog mapped without a transient parent. This is discouraged.
Traceback (most recent call last):
  File "<< GTG source>>/.local_build/install/lib/python3/dist-packages/GTG/gtk/browser/task_pane.py", line 406, in task_setup_cb
    box = TaskBox(self.app.config, self.pane == 'workview')
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<< GTG source>>/.local_build/install/lib/python3/dist-packages/GTG/gtk/browser/task_pane.py", line 44, in __init__
    self.expander.set_indent_for_depth(True)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'TreeExpander' object has no attribute 'set_indent_for_depth'. Did you mean: 'set_indent_for_icon'?
Gtk-Message: 08:52:07.156: GtkDialog mapped without a transient parent. This is discouraged.
Traceback (most recent call last):
  File "<< GTG source>>/.local_build/install/lib/python3/dist-packages/GTG/gtk/browser/task_pane.py", line 488, in task_bind_cb
    expander = box.get_first_child()
               ^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get_first_child'
Gtk-Message: 08:52:07.160: GtkDialog mapped without a transient parent. This is discouraged.
Traceback (most recent call last):
  File "<< GTG source>>/.local_build/install/lib/python3/dist-packages/GTG/gtk/browser/task_pane.py", line 406, in task_setup_cb
    box = TaskBox(self.app.config, self.pane == 'workview')
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<< GTG source>>/.local_build/install/lib/python3/dist-packages/GTG/gtk/browser/task_pane.py", line 44, in __init__
    self.expander.set_indent_for_depth(True)
[...]

PR #1103, opened 2024-05-19
Require GTK 4.10

SqAtx:
Relates to #1093

GTK 4.10 has been required since
https://github.com/getting-things-gnome/gtg/commit/23a986847a61f4744a1e9a323a4f5b353358d2e5

Tested by setting the requirement to 4.13, and getting
```
Run-time dependency gtk4 found: NO (tried cmake)

../meson.build:55:11: ERROR: Dependency lookup for gtk4 with method 'pkgconfig' failed: Invalid version, need 'gtk4' ['>= 4.13'] found '4.12.3'.
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -52,10 +52,11 @@ configure_file(
   configuration: wrapper_config,
 )
 
+dep_gtk4 = dependency('gtk4', version: '>= 4.10')
+
 # Check for some runtime dependencies so they are displayed with an "NO"
 # when building which could give the user an idea what could be missing.
 dep_glib = dependency('glib-2.0', required: false)
-dep_gtk = dependency('gtk+-3.0', required: false)
 dep_libsecret = dependency('libsecret-1', required: false)
 dep_gtksourceview = dependency('gtksourceview-4', required: false)
 
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -9,13 +9,16 @@ on:
     branches: [ master ]
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v4
-    - name: Set up Python 3.8
+
+    # The minimum supported Python version is actually 3.8, but it's not
+    # available on GitHub's ubuntu-24.04 images.
+    - name: Set up Python 3.9
       uses: actions/setup-python@v5
       with:
-        python-version: "3.8"
+        python-version: "3.9"
 
     # Note: we could try and cache the pre-commit environment
     # See the pre-commit docs at https://pre-commit.com/#github-actions-example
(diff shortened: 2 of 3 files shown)


meson.build
```diff
@@ -52,10 +52,11 @@ configure_file(
   configuration: wrapper_config,
 )
 
+dep_gtk4 = dependency('gtk4', version: '>= 4.10')
+
 # Check for some runtime dependencies so they are displayed with an "NO"
 # when building which could give the user an idea what could be missing.
 dep_glib = dependency('glib-2.0', required: false)
-dep_gtk = dependency('gtk+-3.0', required: false)
```
SqAtx, line 58: I'm pretty sure GTK3 is no longer a dependency! Also, I think we want GtkSourceView 5, but one thing at a time.

SqAtx:
Hm, GHA doesn't like this because it uses Ubuntu 22.04 and GTK 4.6... Maybe we should just check at runtime then.

diegogangl:
Why not update the action to 24.04? 
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

There are packages for Gtk 4.14 for that version

SqAtx:
Because I didn't know that GHA already had 24.04 runners =D

But I was also wondering whether this should be a compile-time check. On one hand, it makes sense to fail early, and not compile something that we know won't run. On the other hand, the Ubuntu runners will eventually fall behind too much - if we want to require GTK >4.14 before 2026, we won't be able to use Ubuntu directly on GHA and do something as simple as
```
jobs:
  build:
    runs-on: ubuntu-xxx
```
But now that I've typed this out, I think the compile-time check makes the most sense. If/when we require a version that's not available on Ubuntu LTS, we'll adapt the GHA workflow.

diegogangl:
Ubuntu has backports: https://help.ubuntu.com/community/UbuntuBackports
Not sure if those apply to LTS, or if they will increase the GHA time too much though

SqAtx:
Hm Python 3.8 isn't available to be installed on Ubuntu 2404.04 - https://github.com/getting-things-gnome/gtg/actions/runs/9237207451/job/25413900625

I'm not sure where the limitation comes from exactly. Maybe from the Ubuntu repos. Or maybe form GitHub when they create the images from their VM. Either way, there doesn't seem to be a simple way to use a VM that has both Python 3.8 and GTK 4.10.

I'd be tempted to just raise the minimum requirement to 3.9. 3.8's EOL is at the end of October, and I'm guessing we'd probably be releasing GTG 0.7 before then, so this isn't the most technically precise. But that'd be the pragmatic option.

Or we could keep the minimum requirement at 3.8 nominally, and just check 3.9 on GHA. We can keep the 3.8 requirement in mind when coding, and just not check it automatically. If somebody feels strongly about being able to run GTG 0.7 on Python 3.8, we/they can do whatever backporting work is necessary then.

SqAtx:
I think this is a decent solution!

SqAtx:
Ah, but I just noticed that https://github.com/getting-things-gnome/gtg/commit/05dc6b182da8e9712df88e9c88fb1a1a21d6aa15 introduced a `match` statement, which means GTG needs Python 3.10 to run now (FYI @diegogangl).

diegogangl:
Merged, thanks!
I've been abusing `match` so much, that I forgot we couldn't use it here :)
We can replace that bit with some `if...elif` later

PR #1105, opened 2024-05-25
Apply simple Ruff fixes

SqAtx:
Relates to #237

With one exception, this is the output of `ruff lint --fix`. The changes look minor - it mostly removes unused imports.

Ruff reports a lot of other errors, including a fair amount of "Undefined name" that indicate broken code paths. We can tackle those separately, and decide whether we want to add Ruff to the pre-commit hook.

This commit in itself is still a good quality of life improvement.

I also did some smoke testing to make sure nothing is obviously broken.

### DIFF
--- a/GTG/core/config.py
+++ b/GTG/core/config.py
@@ -165,7 +165,7 @@ class SectionConfig():
         if value is None and default_value is None:
             raise ValueError(
                 'No valid configuration value or default value was '
-                'found for %s in %s'.format(option, self._section_name))
+                f'found for {option} in {self._section_name}')
         elif value is None:
             return default_value
         else:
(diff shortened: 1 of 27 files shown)


GTG/core/config.py
```diff
@@ -165,7 +165,7 @@ def get(self, option):
         if value is None and default_value is None:
             raise ValueError(
                 'No valid configuration value or default value was '
-                'found for %s in %s'.format(option, self._section_name))
+                f'found for {option} in {self._section_name}')
```
SqAtx, line 168: Ruff originally called this a violation of [string-dot-format-extra-positional-arguments (F523)](https://docs.astral.sh/ruff/rules/string-dot-format-extra-positional-arguments/) and proposed to make it
```
'found for %s in %s'.format())
```
which was obviously not what the code is intended to be. The correct fix would have been
```
'found for {} in {}'.format(option, self._section_name))
```
but I figured it makes more sense to use an fstring.

diegogangl:
Looks good, thanks!

Issue #952, opened 2023-01-28
`Ctrl+F` keyboard shortcut should allow refocusing the searchentry widget and select its contents, rather than toggling it off and clearing the search


nekohayo:
Same logic as https://gitlab.gnome.org/GNOME/nautilus/-/issues/2819 :

---

When you search for something, and interact with the search results, this defocuses the searchentry. The problem is that if you want te refocus it afterwards, you have no choice but to (remember to) use the mouse because the `Ctrl+F` keyboard shortcut will clear / toggle out of the search mode entirely, instead of simply focusing the searchentry (and hopefully doing a "select all text" on it).

I would like to be able to easily re-focus the searchentry.

If I want to clear the search entry from the keyboard, I would hit `Ctrl+F` then `Esc` or `Ctrl+F` then `Backspace`, especially given the huge performance impact of displaying an unfiltered view in GTG 0.6.

### Available features and workarounds

Remembering to clicking the searchbar's searchentry field with the mouse cursor, instead of reusing the keyboard shortcut... but I never remember. It is a reflex for me to hit Ctrl+F all the time I want to interact with the searchbar.

### Difficulties and insufficiencies

Most of the time I want to correct, refine, or _replace_ a search rather than clear it entirely. Clearing the search is unexpected behavior for me most of the time, and clearing the search has a huge performance cost/impact, because most of the time it means resetting/refiltering the view to hundreds of items, which in my case might mean 10-20 seconds waiting for the app to unfreeze due to the sheer amount of tasks I have.

Exiting/reentering filtering search mode takes much more time than refining a search.

### Suggested enhancements

When the searchbar is active but unfocused (because you interacted with the grid or listview), hitting Ctrl+F should bring focus back to the searchentry widget (instead of closing/dismissing the search mode) and select all of its text automatically;
* If the user then wants to clear it, they can either press `Esc` or `Backspace` (which would remove all the selected searchentry text anyway)
* If the user wants to replace the entire query with a new one, they can start typing straight away (and benefit from faster search)
* If the user wants to append the contents of the entry, they can use the keyboard's left or right arrows (or mouse cursor) to deselect-and-navigate the searchentry

...so this would be generally safer (in terms of not "losing" the user data that the search query text represents) and better for performance (in terms of not unexpectedly needing to show the unfiltered view if the user didn't intend to).

This "safe" approach is how many other apps (such as Firefox, LibreOffice, Thunderbird, etc.) work.

PR #1106, opened 2024-05-27
Add action to dismiss search bar

### LINKED ISSUE #1003: `Escape` key does not allow dismissing / closing the searchbar
Since the GTK4 port, the `Escape` (Esc) key does not dismiss the searchbar (it does in 0.6 and earlier).

Related: #952

diegogangl:
Pressing ESC now closes the search bar again. Good times

fixes #1003


nekohayo:
A quickie glance at the diff tells me that the `Ctrl+F` key would still clear the search instead of refocusing; would you want to address #952 at the same time in this MR in the same commit, or better to keep it separate?

diegogangl:
Best to keep the PRs short and sweet for now

PR #1107, opened 2024-05-27
Restore selected tag on startup

### LINKED ISSUE #551: Remember and activate the sidebar's last viewed tag filter on startup
Considering how @jaesivsm's optimization to process tasks only for the currently used tab (Open vs Actionable vs Closed) has pretty much cut startup time roughly in half (question: does the code explicitly restore only the last used tab view on startup, without processing the whole tree, or does it still somehow do the work "at least twice instead of thrice"?)... some crazy thought just occurred to me, not only as a usability enhancement, but also as a crazy startup performance hack: on startup, when initially determining which tasks to show/process, GTG should not only restore the state of "which tab is selected", but also the state of "which tag was selected in the sidebar" (if any), and filter to that.

Taking my own data as an example, even my most populous tag only encompasses a third of my tasks, and most tags encompass 1/10th or even 1/20th of the total tasks... so you would get the compounding effect of "2x faster startup from showing only actionable tasks, multiplied by 3-20x faster startup from showing only those for the selected tag".

If that actually works, that'd be a crazy cheap and effective optimization hack for most cases. Of course this is a _dirty, dirty hack_ sweeping the problem under the rug and this doesn't eliminate the need to truly make the software faster at processing task trees (issue #109), but it I see no reason not to do it as it would be an extra performance boost that compounds on top of those future optimizations, so when we eventually solve 109 we might get near-instantaneous startup ;)

diegogangl:
Saves and restores selected tag on startup

closes #551

### DIFF
--- a/GTG/core/config.py
+++ b/GTG/core/config.py
@@ -53,6 +53,7 @@ DEFAULTS = {
         'sort_mode_open': 'title',
         'sort_mode_active': 'title',
         'sort_mode_closed': 'title',
+        'selected_tag': '',
     },
     'tag_editor': {
         "custom_colors": [],
--- a/GTG/gtk/application.py
+++ b/GTG/gtk/application.py
@@ -237,6 +237,7 @@ class Application(Gtk.Application):
         # Browser (still hidden)
         if not self.browser:
             self.browser = MainWindow(self)
+            self.browser.restore_tag_selection()
 
             if self.props.application_id == 'org.gnome.GTG.Devel':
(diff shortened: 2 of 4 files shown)


SqAtx (approved):
LGTM
Note that this could be extended, because GTG support selecting several tags at the same time. In that case, it shows tasks that have all the tags. With the current implementation, only the first one is selected at startup. But we can improve that in a later commit.

diegogangl:
Definitely. Thanks for reviewing :+1:

PR #1108, opened 2024-05-30
Fix the type hints in base_store.py

gycsaba96:
Type hints could provide better maintainability. My final goal is to fix the typing of the main core components at least. I also hope to catch some bugs along the way. This PR fixes `base_store.py`.

- Use the `_Element` class instead of the `Element` factory for typing.
- Make `BaseStore` generic so that subclasses can have the precise return and parameter types instead of `Any`.
- The stored objects are required to have certain fields.

One can test the result using mypy:

```
mypy GTG/core/base_store.py 
```

(You might need to install the `lxml-stubs` python package.)

### DIFF
--- a/GTG/core/base_store.py
+++ b/GTG/core/base_store.py
@@ -25,7 +25,7 @@ from uuid import UUID
 import logging
 
 from lxml.etree import _Element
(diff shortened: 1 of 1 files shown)


diegogangl:
LGTM. Merged, thanks!

PR #1114, opened 2024-06-12
Fix sensitivity of start button

### LINKED ISSUE #1004: "Start Tomorrow" button and menubutton sensitivity state is broken
Since the GTK4 port, these two aspects are broken:

* [ ] The "Start Tomorrow" button remains sensitive even if no task is selected in the view
* [ ] The "Start Tomorrow" button's arrow menu is permanently insensitive/disabled, even if items are selected in the view

diegogangl:
Fixes #1004

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -401,7 +401,7 @@ class MainWindow(Gtk.ApplicationWindow):
         self.connect('notify::default-height', self.on_window_resize)
 
         for p in PANE_STACK_NAMES_MAP.keys():
-            pane = self.stack_switcher.get_stack().get_child_by_name(p).get_first_child()
+            pane = self.get_pane_by_name(p)
             pane.task_selection.connect('selection-changed', self.on_selection_changed)
 
         self.sidebar.connect('selection_changed', self.on_sidebar_select_changed)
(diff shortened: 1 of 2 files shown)


GTG/gtk/browser/sidebar.py
```diff
@@ -217,6 +217,13 @@ def change_pane(self, pane: str) -> None:
         self.tags_filter.changed(Gtk.FilterChange.DIFFERENT)
 
 
+    @GObject.Signal
+    def selection_changed(self) -> None:
+        """Selection of tags or searches has changed."""
+
```
SqAtx, line 223:
```suggestion
```

GTG/gtk/browser/main_window.py
```diff
@@ -400,6 +400,12 @@ def _init_signal_connections(self):
         self.connect('notify::default-width', self.on_window_resize)
         self.connect('notify::default-height', self.on_window_resize)
 
+        for p in PANE_STACK_NAMES_MAP.keys():
+            pane = self.stack_switcher.get_stack().get_child_by_name(p).get_first_child()
```
SqAtx, line 404: nit: Maybe at some point we can have a method that does this like `get_pane_by_name()`

SqAtx (approved):
LGTM
I don't really understand the GHA error. It's failing on `python -m pip install --upgrade pip` and not on our actual tests. So, re-running it would probably be fine?

GTG/gtk/browser/main_window.py
diegogangl: Good idea!

diegogangl:
Thanks for the reviews! The tests seem to be doing ok now

PR #1117, opened 2024-06-23
Fix (almost) every type hint in tags.py

gycsaba96:
- Use the _Element class instead of the Element factory for typing.
- Convert tag IDs to UUID during XML parsing. (The same thing caused many problems for tasks.)
- During XML parsing, the `parent` property contains a UUID in the new format. This allowed for some simplifications. A related test case is also fixed.

### DIFF
--- a/GTG/core/tags.py
+++ b/GTG/core/tags.py
@@ -19,15 +19,15 @@
 """Everything related to tags."""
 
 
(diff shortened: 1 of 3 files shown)


diegogangl:
LGTM, thanks!

PR #1119, opened 2024-06-30
Fix re-opening task editors

gycsaba96:
Task editors did not re-open because `MainWindow.restore_editor_windows` used strings instead of UUIDs to look up tasks in the data store.


diegogangl:
Thanks!

PR #1120, opened 2024-07-02
Simple typing fixes for tasks.py

gycsaba96:
- use the `_Element` class instead of the `Element` factory for typing
- expect `None` values often used by lxml
- rename the `title` setter to `set_title`
- fix default value for `duplicate_cb`

This fixes every typing error mypy indicates except for the signature errors of some overrides. (I want to investigate the saved searches before looking into those.)


diegogangl:
Merged, thanks!

Issue #708, opened 2021-10-07
Feature Request: Add flatpak nightly builds


BeatLink:
It would be great if GTG could have nightly flatpak builds from the master branch. This would make the bleeding edge features available to user, which would mean more users willing to beta test

BeatLink:
What about github actions?

SqAtx:
I believe that the flatpak outputted by https://github.com/getting-things-gnome/gtg/actions/workflows/weekly_flatpak.yml doesn't work. I downloaded the one from https://github.com/getting-things-gnome/gtg/actions/runs/9311435996#artifacts and got:
```
$ flatpak install org.gnome.GTG.Devel_20240530.flatpak
error: The application org.gnome.GTG.Devel/x86_64/master requires the runtime org.gnome.Platform/x86_64/master which was not found
```
Looking at
```
$ flatpak install org.gnome.Platform
Looking for matches…
Similar refs found for ‘org.gnome.Platform’ in remote ‘flathub’ (system):

   1) runtime/org.gnome.Platform/x86_64/3.24
   2) runtime/org.gnome.Platform/x86_64/3.26
   3) runtime/org.gnome.Platform/x86_64/3.28
   4) runtime/org.gnome.Platform/x86_64/3.30
   5) runtime/org.gnome.Platform/x86_64/3.32
   6) runtime/org.gnome.Platform/x86_64/3.34
   7) runtime/org.gnome.Platform/x86_64/3.36
   8) runtime/org.gnome.Platform/x86_64/3.38
   9) runtime/org.gnome.Platform/x86_64/40
  10) runtime/org.gnome.Platform/x86_64/41
  11) runtime/org.gnome.Platform/x86_64/42
  12) runtime/org.gnome.Platform/x86_64/43
  13) runtime/org.gnome.Platform/x86_64/45
  14) runtime/org.gnome.Platform/x86_64/44
  15) runtime/org.gnome.Platform/x86_64/46

Which do you want to use (0 to abort)? [0-15]: 
```
it seems that https://github.com/getting-things-gnome/gtg/blob/c501ff800e4517f66b3254f373d41bf9afd861fa/build-aux/org.gnome.GTG.Devel.json#L4 doesn't work and we should use something like `"runtime-version": "46"` instead, and keep it up to date.

When I have a minute, I'll be working on:
- Fixing this
[...]

PR #1122, opened 2024-07-04
Use GNOME 46 in weekly flatpak

SqAtx:
Relates to #708

This fixes the following error:
```
$ flatpak install org.gnome.GTG.Devel_20240530.flatpak
error: The application org.gnome.GTG.Devel/x86_64/master requires the runtime org.gnome.Platform/x86_64/master which was not found
```
by using the latest GNOME runtime, as opposed to the "master" one, which doesn't seem to exist.

Tested by running the workflow on my fork, downloading the artifact from GHA, installing and running the flatpak.

### DIFF
--- a/.github/workflows/weekly_flatpak.yml
+++ b/.github/workflows/weekly_flatpak.yml
@@ -12,11 +12,11 @@ jobs:
     # again
     runs-on: ubuntu-24.04
     container:
-      image: bilelmoussaoui/flatpak-github-actions:gnome-nightly
+      image: bilelmoussaoui/flatpak-github-actions:gnome-46
       options: --privileged
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/weekly_flatpak.yml
```diff
@@ -12,11 +12,11 @@ jobs:
     # again
     runs-on: ubuntu-24.04
     container:
-      image: bilelmoussaoui/flatpak-github-actions:gnome-nightly
+      image: bilelmoussaoui/flatpak-github-actions:gnome-46
```
SqAtx, line 15: This isn't strictly necessary. We could use nightly. I just figured this would be more stable, and it's okay if we lag behind a bit after new GNOME releases.

diegogangl: True, we can update every now and then

PR #1124, opened 2024-07-07
More Ruff fixes

SqAtx:
Relates to #237
This is a follow-up to https://github.com/getting-things-gnome/gtg/commit/c2f650f21e50f5d1dad1d9d6226909417656616d

I ran `ruff check` and looked at the errors that Ruff finds but doesn't want to fix automatically.

There are still 13 errors, which all seem to indicate a bug. They are mostly "Undefined name". I will fix them individually.

### DIFF
--- a/GTG/core/clipboard.py
+++ b/GTG/core/clipboard.py
@@ -56,7 +56,6 @@ class TaskClipboard():
                 if hasattr(ta, 'is_subtask'):
                     is_subtask = True
                     tid = ta.child
-                    tas = self.ds.tasks.lookup[tid]
                     self.content.append(['subtask', tid])
             if not is_subtask:
                 if end_line.get_line() < stop.get_line():
--- a/GTG/gtk/colors.py
+++ b/GTG/gtk/colors.py
@@ -155,15 +155,12 @@ def get_colored_tags_markup(ds, tag_names):
 
 def generate_tag_color():
 
-    maxvalue = 1.0
-    flag = 0
-    while(flag == 0):
+    while True:
         rgba = random_color()
         my_color = rgb_to_hex(rgba)
         if my_color not in used_color:
-            flag = 1
-    used_color.append(my_color)
-    return my_color
+            used_color.append(my_color)
+            return my_color
 
 
 def grgba_to_hex(rgba, ignore_alpha=True):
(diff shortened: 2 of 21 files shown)


GTG/core/clipboard.py
```diff
@@ -56,7 +56,6 @@ def copy(self, start, stop, bullet=None):
                 if hasattr(ta, 'is_subtask'):
                     is_subtask = True
                     tid = ta.child
-                    tas = self.ds.tasks.lookup[tid]
```
SqAtx, line 59: I wasn't sure what the clipboard is supposed to do, so I didn't check whether there is a bug here.

Copy/pasting the contents of a task with subtasks doesn't show the subtasks as subtasks, just text. But that's also the behaviour in 0.6.

GTG/gtk/colors.py
```diff
@@ -155,15 +155,12 @@ def get_colored_tags_markup(ds, tag_names):
 
 def generate_tag_color():
 
-    maxvalue = 1.0
```
SqAtx, line 158: The original error was that `maxvalue` is unused. The loop was written basically in C, so I wrote something more Pythonic. I checked that tag color generation still works.

tests/core/test_quickadd.py
```diff
@@ -149,6 +149,7 @@ def test_start(self):
         self.assertEqual(expected1, parse(text1))
         self.assertEqual(expected1, parse(text2))
         self.assertEqual(expected1, parse(text3))
+        self.assertEqual(expected2, parse(text4))
```
SqAtx, line 152: `expected2` and `text4` were unused. Looking at them, I figured it was because of this missing assertion.

tests/core/test_task.py
```diff
@@ -643,18 +643,6 @@ def test_simple_reverse_sort(self):
         self.assertEqual(task_store.data, expected)
 
 
-    def test_simple_reverse_sort(self):
```
SqAtx, line 646: The test was duplicated.

GTG/core/clipboard.py
diegogangl: Yeah, the clipboard is supposed to handle GTG content. That's probably an old bug

diegogangl:
Looks good, thanks!

PR #1125, opened 2024-07-09
chore: move snap to core24 and GNOME 46

sergiusens:
- move snap to core24 which brings GNOME 46
- simplify how liblarch is built
- setup gitignore for expected output


diegogangl:
LGTM, thanks!

Issue #1115, opened 2024-06-16
Cannot remove tags, selecting to delete the same tag twice induces crash


mattmcadoo:
```
Gtk-Message: 23:50:46.202: GtkDialog mapped without a transient parent. This is discouraged.
Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/GTG/gtk/browser/delete_tag.py", line 53, in on_response
    self.on_delete_confirm()
  File "/usr/lib/python3.12/site-packages/GTG/gtk/browser/delete_tag.py", line 44, in on_delete_confirm
    self.browser.app.ds.tags.remove(tag.id)
  File "/usr/lib/python3.12/site-packages/GTG/core/base_store.py", line 114, in remove
    item = self.lookup[item_id]
           ~~~~~~~~~~~^^^^^^^^^
KeyError: '11ce34cc-108b-4ef5-b0a0-1f8a00ad61da'
2024-06-15 23:50:47,946 - INFO - errorhandler:handle_response:137 - Going to exit because either of fatal error or user choice
Aborted (core dumped)
```

   repository:               https://github.com/getting-things-gnome/gtg.git
   at the commit:            c501ff800e4517f66b3254f373d41bf9afd861fa

Gentoo using an ebuild I modified to pull directly from Github.

Build info:
```
The Meson build system
Version: 1.4.1
Source dir: /var/tmp/portage/app-office/gtg-9999/work/gtg-9999
Build dir: /var/tmp/portage/app-office/gtg-9999/work/gtg-9999-build
Build type: native build
Project name: gtg
Project version: 0.6.0
Host machine cpu family: x86_64
Host machine cpu: x86_64
Program python3 found: YES (/var/tmp/portage/app-office/gtg-9999/temp/python3.12/bin/python3)
Configuring prefix-gtg.sh using configuration
Found pkg-config: YES (/usr/bin/x86_64-pc-linux-gnu-pkg-config) 2.2.0
Run-time dependency gtk4 found: YES 4.14.4
Run-time dependency glib-2.0 found: YES 2.78.6
Run-time dependency libsecret-1 found: YES 0.21.1
Run-time dependency gtksourceview-4 found: YES 4.8.4
Configuring gtg using configuration
Configuring local-gtg using configuration
Program git found: YES (/usr/bin/git)
Configuring info.py using configuration
Program msgfmt found: YES (/usr/bin/msgfmt)
Configuring org.gnome.GTG.appdata.xml.in using configuration
[...]

PR #1126, opened 2024-07-11
Fixes for 'filter_tag'

gycsaba96:
This PR has two main changes:
- Iterate over all tasks, not just the root elements.
- Consider all descendants of a tag, not just the immediate children.

Because of these bugs, it was impossible to get rid of the `money` tag in the default dataset:
1. Start the program with the default dataset. 
2. Delete the `money` tag in the sidebar.
3. Restart the application.
4. Observe that the `money` tag is absent from the sidebar.
5. Open the _"Learn How to Use Tags and Enable the Sidebar"_ task.
6. Observe the `money` tag reappearing in the sidebar. 

(I discovered this bug while investigating Issue #1115. They might be related, but the described anomaly is different.)

### DIFF
--- a/GTG/core/tags.py
+++ b/GTG/core/tags.py
@@ -179,6 +179,15 @@ class Tag(GObject.Object):
             ancestors.append(here)
         return ancestors
 
+
+    def get_matching_tags(self) -> List['Tag']:
+        """Return the tag with its descendants."""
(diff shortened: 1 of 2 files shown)


diegogangl:
Looks good, merging. Thanks!

PR #1127, opened 2024-07-23
Look for parent in due date binding

### LINKED ISSUE #1072: Effective due date isn't shown in main view
I have a task with no due date, but its parent has one.

The due date of the parent is used when sorting by due date (makes sense), but is not shown in the main view.

This is what I see after sorting my tasks by due date:

image

GTG is actually using 2024-01-31 as the due date of the first task, but isn't showing it.

This is happening on https://github.com/getting-things-gnome/gtg/commit/4ed9bc88067d5efdb48d4e262e7e6a7872095c9c, but the behaviour in 0.6 was correct.

diegogangl:
If the task's parent has a due date then we should show sub-tasks as having that due date.

Fix #1072

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -564,7 +564,7 @@ class Task(GObject.Object):
     @GObject.Property(type=bool, default=False)
     def has_date_due(self) -> bool:
         if not self._has_date_due and self.parent:
-            return self.parent._has_date_due
+            return self.parent.has_date_due
         else:
             return self._has_date_due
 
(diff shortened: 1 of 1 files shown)


GTG/core/tasks.py
```diff
@@ -563,7 +563,10 @@ def is_recurring(self) -> bool:
 
     @GObject.Property(type=bool, default=False)
     def has_date_due(self) -> bool:
-        return self._has_date_due
+        if not self._has_date_due and self.parent:
+            return self.parent._has_date_due
```
SqAtx, line 567:
```suggestion
            return self.parent.has_date_due
```
I think this works? It avoids accessing a protected member.
Same with the other one.

SqAtx (approved):
LGTM (please address my suggestion if it makes sense)

Thanks for fixing this!

GTG/core/tasks.py
diegogangl: Sounds good, changed

PR #1129, opened 2024-07-27
Improve version name

SqAtx:
Update base version from 0.6.0 to 0.7-dev.

The idea is that when we're ready to release 0.7, we'll create a 0.7  branch. We'll then update the version to "0.7" on that branch, and "0.8-dev" on master.

Append the short commit hash in development mode (as before)

Append the date in flatpak mode, to help keep track of versions when using the weekly flatpak. Tested the flatpak by running the GHA workflow on my fork.

### DIFF
--- a/GTG/core/meson.build
+++ b/GTG/core/meson.build
@@ -1,14 +1,19 @@
-# Append short commit hash and put it in
-# info.py when using development profile.
-# Use git rev-parse --short HEAD if available.
-
 git = find_program('git', required : false, disabler : true)
 
 if git.found() and get_option('profile') == 'development'
-    version = meson.project_version() + '-' + run_command(
+    base_version = meson.project_version()
+    commit = run_command(
             ['git', 'rev-parse', '--short', 'HEAD'],
             check: true
         ).stdout().strip()
+    version = base_version + '-' + commit
+elif get_option('profile') == 'flatpak_snapshot'
+    base_version = meson.project_version()
+    date = run_command(
+            ['date', '--utc', '+%Y%m%d'],
+            check: true
+        ).stdout().strip()
+    version = base_version + '-' + date
 else
     version = meson.project_version()
 endif
--- a/build-aux/org.gnome.GTG.Devel.json
+++ b/build-aux/org.gnome.GTG.Devel.json
@@ -66,7 +66,7 @@
     {
       "name": "gtg",
       "config-opts" : [
-          "-Dprofile=development"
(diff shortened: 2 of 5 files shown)


SqAtx:
Interestingly, this didn't fix what I originally set out to fix:
image
It still says 0.6 in the list of flatpaks.

But the About dialog is correct:
image
The dev version, for comparison:
image

diegogangl:
The version for flatpak comes from the appdata.xml file: https://github.com/getting-things-gnome/gtg/blob/master/data/org.gnome.GTG.appdata.xml.in.in

I guess we could add a 0.7-dev version while we are working on a new release

diegogangl (approved):

SqAtx:
Oh cool, I'll have a look at that file!

SqAtx:
Hm it doesn't seem to be quite as simple as that. I tried adding `<release version="0.7-dev" type="snapshot"></release>` or `<release version="0.7-dev" type="development"></release>` at the top of the list of releases (as per https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Releases.html).

But it always shows 0.6 when inside `flatpak list`.

There may be a place to tell flatpak that this is a dev or snapshot version - I'll experiment.

SqAtx:
Ah OK, I tried building the flatpak locally and I got "F: Ignoring release element without timestamp or date", which would explain it. And indeed, adding `<release version="0.7-dev"/>` (stable is the default) also results in `flatpak list` showing 0.6.

But there must be more to it. To be sure, I added
```
        <release version="0.7-dev" date="2022-03-05" type="development"/>
        <release version="0.7-snap" date="2022-03-04" type="snapshot"/>
        <release version="0.7-stable" date="2022-03-03" type="stable"/>
```
to the XML. Building the flatpak locally with `sudo flatpak-builder --force-clean --install-deps-from=flathub --repo=repo --install builddir build-aux/org.gnome.GTG.Devel.json` results in the version being 0.7-dev, as expected.

But pushing to a fork, triggering the workflow, downloading and installing the flatpak, still shows 0.6. I don't understand where it gets that 0.6 from. I'll keep digging.

SqAtx:
OK it works now
image
I think something was wrong with my test setup last time :)

I also renamed the `flatpak` meson profile to `flatpak_snapshot`. I think it makes more sense.

diegogangl:
LGTM! Thanks, and yeah dev snapshot looks better

PR #1132, opened 2024-08-10
Fix (almost) every typing error in saved_searches.py

gycsaba96:
- use the `_Element` class instead of the `Element` factory for typing
- convert tag IDs to `UUID` during XML parsing
- expect `None` values returned by lxml
- other minor typing fixes
- fix a related test case


diegogangl:
LGTM, thanks!

PR #1133, opened 2024-08-18
Fix most of the typing errors in datastore.py

gycsaba96:
The main changes are the following:
- use the `_Element`/`_ElementTree` class instead of the `Element`/`ElementTree` factory for typing,
- expect `None` values returned by lxml,
- expect `None` values in case the methods are used in the wrong order,
- fix the use of undefined variables in `do_first_run_versioning`.

(Please note that `do_first_run_versioning` is never called. Maybe it is a bug or something to be removed.)


diegogangl:
LGTM, thanks! Will make a note to look at that function too

PR #1134, opened 2024-09-01
Fix typing errors in firstrun_tasks.py

gycsaba96:
The changes are as follows:
- use the `_Element`/`_ElementTree` class instead of the `Element`/`ElementTree` factory for typing
- expect `None` values returned by lxml
- add type hints to the default tasks
- fix the return type of the `generate` function


diegogangl:
LGTM, thanks!

PR #1135, opened 2024-09-03
Sort Order

### LINKED ISSUE #1021: Sorting does not set an arrow/triangle (such as ▼/▲) to indicate the sorting order in the popover menu
Since the GTK4 port, we don't have a treeview header area anymore, but we have a sorting popover menubutton.

The state and direction of sorting should be reflected within that menu, on startup and when it is changed by the user.

diegogangl:
This PR adds a couple of buttons to reverse the sorting order (ascending or descending), also converts the sorting options to stateful actions which gives us radio buttons

closes #1021

### DIFF
--- a/GTG/core/sorters.py
+++ b/GTG/core/sorters.py
@@ -31,13 +31,45 @@ def unwrap(row, expected_type):
 
     return item
 
+class ReversibleSorter(Gtk.Sorter):
 
-class TaskTitleSorter(Gtk.Sorter):
+    def __init__(self) -> None:
+        self._reverse: bool = False
+        super().__init__()
+
+
+    @property
+    def reverse(self) -> bool:
+        return self._reverse
+
+
+    @reverse.setter
+    def reverse(self, value: bool) -> None:
+        self._reverse = value
+        self.changed(Gtk.SorterChange.INVERTED)
+
+
+    def reversible_compare(self, first, second) -> Gtk.Ordering:
+        """Compare for reversible sorters."""
+
+        if self._reverse:
+            if first < second:
+                return Gtk.Ordering.LARGER
+            elif first > second:
+                return Gtk.Ordering.SMALLER
+            else:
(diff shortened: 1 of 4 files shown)


GTG/core/sorters.py
```diff
@@ -31,14 +31,46 @@ def unwrap(row, expected_type):
...
 
-    def __init__(self):
-        super(TaskTitleSorter, self).__init__()
+    @reverse.setter
+    def reverse(self, value: bool) -> None:
+        self._reverse = value
+        self.changed(Gtk.SorterChange.INVERTED)
 
 
+    def reversible_compare(self, first, second) -> Gtk.Ordering:
```
SqAtx, line 52: An alternative is to have the equal test first:
```
        if first == second:
            return Gtk.Ordering.EQUAL

        if first < second:
            if self._reverse:
                return Gtk.Ordering.LARGER
            else:
                return Gtk.Ordering.SMALLER

        if self._reverse:
            return Gtk.Ordering.SMALLER
        else:
            return Gtk.Ordering.LARGER
```

The entire thing looks a bit clunky to me, but is apparently the GTK way to solve this problem, so LGTM.

diegogangl:
Used stateful actions for the sorting menu, it's looking a lot better now

Screenshot_20240914_190814

SqAtx (approved):
Well you updated the review while I was looking at it ^^'

It looks a lot better now!  The UI is clear and usable.

I do think we can iterate again on the UI later.  I don't have concrete suggestions right now, but it seems to me that there is potential to make it a bit more compact / require fewer clicks. But yeah, very happy with how it turned out :)

GTG/core/sorters.py
diegogangl: Ah that's a bit cleaner. Thanks! But yeah, I couldn't find a less clunky way of doing this

PR #1136, opened 2024-09-08
Fix calendar selection for repeating tasks

### LINKED ISSUE #1015: Setting yearly recurring tasks from the Task Editor's menu sets them as repeating monthly
Since the GTK4 port, setting yearly recurring tasks is broken in the Task Editor dialog: when clicking a date in the minicalendar in that dialog's menu, it sets them as monthly recurring.

gycsaba96:
Fixes #1015

After selecting the desired day in the yearly calendar, the `_update_calendar` method synced the monthly calendar, thus triggering the `_on_monthly_selected` callback. The bug also worked the other way around. 

Now, the `_update_calendar` blocks these callbacks using a `with` statement to prevent the issue.

### DIFF
--- a/GTG/gtk/data/recurring_menu.ui
+++ b/GTG/gtk/data/recurring_menu.ui
@@ -102,13 +102,11 @@
         <!-- without, it covers the section heading -->
         <property name="margin-top">6</property>
         <property name="show-heading">False</property>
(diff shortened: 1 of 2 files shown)


diegogangl:
LGTM, thanks!

PR #1140, opened 2024-09-20
correction README.md

ArimaBatista:
Correcting the dependency name from libsecret-1.0 to libsecret-1-dev, located in the "'Setting up & running the development version >>> Dependencies >>> You can get most of those from your distribution packages: >>> # On Debian/Ubuntu:' section."

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ sudo dnf install meson python3-cairo python3-gobject gtk3 itstool gettext python
 # On Debian 10 (buster), you need to install the backported version, activate it with:
 echo 'deb http://deb.debian.org/debian buster-backports main' | sudo tee -a /etc/apt/sources.list
 # On Debian/Ubuntu:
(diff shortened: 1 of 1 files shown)


Neui:
Why? From what I can see, it adds nothing for us (unless mesen wants the pkg-config files?). In fact, we could remove `libsecret-1.0` because `gir1.2-secret-1` already depends on it (according to Ubuntu packages).

(Also the PR title and commit message are super generic, something like "README: fix debian/ubuntu libsecret dependency" is way more descriptive, along with the why in the message body.)

diegogangl:
Definitely, if gir1.2-secret-1 is already going to drag libsecret-1 then we should get rid of that package in the list

pentux-GitHub:
Coming back to this while reviewing dormant PRs for 0.7. The Debian install line this was fixing no longer exists, the README was rewritten for the GTK4 port, and Neui and Diego concluded above that the right fix was removing the package rather than renaming it. I'd think it's safe to close this one. Thanks @ArimaBatista for the contribution!

PR #1141, opened 2024-09-20
Rename info.py.in to info.py

ArimaBatista:
File naming error resulting in test failures


ArimaBatista:
This causes the error 'GTG/core/meson.build:25:10: ERROR: File info.py.in does not exist'. To fix this, change 'info.py.in' to 'info.py' on line 26 of that file. After this change, 'bash launch.sh' will work with automated tests and 'precommit'.

Neui:
`.in` files are to be post-processed (usually by the build system). In this case, it is for the version (replacing `@VCS_TAG@` with the version).

I don't like this solution because it removes the indication that it is a file to be post-processed. If anything, tests should just work with the build-installed version since we don't support running the project in-place anyway (which sound what you are trying to do), unless we (or someone else) make the project runnable in-place.

PR #1142 depends on this commit, so you should just combine both into one PR (you could add the commit to this repo, but it seems you are editing from the Github Web UI (because of the branch name) and I am not sure if it can do that.)

SqAtx:
Yes, that's not actually a file naming error. The problem is that meson needs to be run before running the tests. That will create process the `.py.in` file in the output folder, and you can then run the tests there. But that's not documented except in the comments of #1036:

> The file `info.py.in` is a template that gets filled (also called configured) by the meson setup step.
> The resulting file is placed into the build directory.
> This is the file we actually want Python to find and use.

It's kind of annoying to have to do that. It gets in the way of quick iteration when writing code and tests in parallel. So on my local machine, after running meson once, I copied the resulting `info.py` into `GTG/core`. Now I can run the tests normally:
```
~/src/gtg$ pytest
==================================================================== test session starts =====================================================================
platform linux -- Python 3.12.3, pytest-8.2.2, pluggy-1.5.0
rootdir: /home/kevin/src/gtg
configfile: pyproject.toml
testpaths: tests
collected 160 items                                                                                                                                          

tests/backend/backend_caldav_test.py sssssss                                                                                                           [  4%]
tests/core/test_config.py ...................                                                                                                          [ 16%]
[...]

=============================================================== 153 passed, 7 skipped in 1.20s ===============================================================
```

This is probably worth a subsection under "Setting up a development environment" in the README. I'll look into that.

pentux-GitHub:
Reviewing dormant PRs for 0.7: as SqAtx explained above, info.py.in is a meson template filled at build time, so the rename would conflict with the build system. The underlying pain (running tests without a build step) is real and is being addressed by documentation in #1144. I'd think this one can be closed. Thanks @ArimaBatista!

diegogangl:
Hi @ArimaBatista, I feel bad for closing your PRs but as @pentux-GitHub said this is better addressed by changing the docs. Thanks again!

PR #1142, opened 2024-09-20
Update meson.build

ArimaBatista:
As a follow-up to issue #1141, we've successfully renamed info.py.in to info.py. This change has resolved the automated test failures we were experiencing.

PR #1144, opened 2024-09-21
Add README section to run the tests

SqAtx:
Describe a simple way to get the tests running. This should help clear the confusion around the `ModuleNotFoundError: No module named 'GTG.core.info'` type of errors you get when just running `pytest` (see #1141).

Adapted from the explanation at #1036

We can close #1141 and #1142

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,20 @@ them locally with `pre-commit run --all-files`.
 For more guidelines about contributing, see
 [CONTRIBUTING.md](./CONTRIBUTING.md).
 
+### Running the tests
+
+Tests are run using `pytest`.
+
+We first need some files to be pre-processed by meson, which is done by the
+following one-time step:
+```sh
+meson setup ./build
+cd GTG/core/ && ln -s ../../.build/GTG/core/info.py .
+```
+
+Then you can run the tests with `pytest` in the repository root.
+
+
 # "Where is my user data and config stored?"
 
 It depends:
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ We first need some files to be pre-processed by meson, which is done by the
 following one-time step:
 ```sh
 meson setup ./build
-cd GTG/core/ && ln -s ../../.build/GTG/core/info.py .
+cd GTG/core/ && ln -s ../../build/GTG/core/info.py .
 ```
 
 Then you can run the tests with `pytest` in the repository root.


Neui:
I think the chosen text puts a bit too much on the "why" rather than how to run it as a README file.
Also I don't think referencing the GitHub Actions workflow is useful if we already show the commands to use anyway here.
My suggestion:

> ### Running the tests
> 
> Tests are run using `pytest`.
> 
> Currently GTG doesn't support testing in-place, but you can copy the required files into your source tree for local development:
> ```sh
> meson setup ./build  # Generate required files
> cp .build/GTG/core/info.py GTG/core/  # Copy required files
> ```
> 
> Then you can run the tests with `pytest` in the repository root.


Also:
1. What about using `.local_build` created by the `launch.sh` script instead?
2. Using `ln -s` to create that link may also be useful so the file "auto-updates".
3. We should include how to run tests for the "built" version (I think it's `ninja -C .local_build test` or something, I would need to look into that again)

diegogangl:
I agree with @Neui 's remarks. 

> What about using .local_build created by the launch.sh script instead?

Not so sure about this. I think it's better to keep builds-to-run and builds-to-test separate, to avoid any mixups

SqAtx:
> I think the chosen text puts a bit too much on the "why" rather than "how" to run it in a README file, which should be straight to the point.

Fair point. Figured I'd put it here to remind myself of the "why" but you're probably right - there may be a place for such documentation but that place isn't here.

> Additionally, I don't think referencing the GitHub Actions workflow is useful if we already show the commands to use anyway here.

The thinking behind it was that if they ever drift apart (i.e. we update the workflow but not the README), the README will still point to an example of a setup that works. That said, the workflow actually does a lot more than this simple setup, so it might not be that relevant.

> We should include how to run tests for the built version (I think it's ninja -C .local_build test or something, I would need to look into that again)

Come to think of it, maybe that's what the GHA workflow should be running, even. I'm not too fussed about that right now, just trying to lower the barrier to entry.

README.md
```diff
@@ -182,6 +182,20 @@ them locally with `pre-commit run --all-files`.
...
 
+### Running the tests
+
+Tests are run using `pytest`.
+
+We first need some files to be pre-processed by meson, which is done by the
+following one-time step:
+```sh
+meson setup ./build
+cd GTG/core/ && ln -s ../../.build/GTG/core/info.py .
```
SqAtx, line 193: I don't like this, but my first idea didn't work:
```
$ ln -s .build/GTG/core/info.py GTG/core/
$ ls -l GTG/core/info.py
lrwxrwxrwx 1 kevin kevin 23 Nov  8 21:10 GTG/core/info.py -> .build/GTG/core/info.py
```

diegogangl:
Just tested this and I'm getting a `ModuleNotFoundError: No module named 'GTG.core.info'`.

picsel2:
Hi there! I have to decline the request to review this PR. My priorities have shifted away from GTG.

pentux-GitHub:
This still fills a real gap, the README doesn't document how to run the tests anywhere, and it's one character away from working.

The `ModuleNotFoundError: No module named 'GTG.core.info'` you hit, @diegogangl, is a path typo in the snippet. `meson setup ./build` generates the file at `./build/GTG/core/info.py`, but the symlink points at `../../.build/...` (leading dot), which doesn't exist. I reproduced it with a minimal meson project; the correct relative path from `GTG/core/` is `../../build/...`:

​```sh
meson setup ./build
ln -s ../../build/GTG/core/info.py GTG/core/info.py
​```

With that fix (and the shorter, how-focused wording @Neui suggested), this looks ready. It's also the right fix for the underlying issue behind #1141 and #1142, both of which try to drop the `.in` extension, which breaks the `@VCS_TAG@` substitution and, as noted there, breaks `./launch`. Documenting the meson step here, as @SqAtx pointed out, is the correct approach and lets those two be closed. @SqAtx if you'd like to finish it, otherwise the change is small enough for anyone to pick up.

README.md
pentux-GitHub:
```suggestion
cd GTG/core/ && ln -s ../../build/GTG/core/info.py .
```

`meson setup ./build` generates the file under `build/` (no leading dot), so the relative path from `GTG/core/` is `../../build/...`. Tested on current master: with this line the import works and the test suite runs.

pentux-GitHub (commented):
Verified on current master: with the one-character fix below, the recipe works end to end — `meson setup ./build`, the symlink, then `pytest` runs the suite without the `ModuleNotFoundError`. I also reproduced the error Diego hit using the current `.build` path, confirming the typo is the only thing standing between this PR and mergeable. Suggested change attached so it can be applied in one click.

PR #1145, opened 2024-09-21
Remove liblarch

SqAtx:
I think it has been unused since the move to GTK4.

Tested by creating and running the flatpak.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -48,7 +48,6 @@ parts:
       - python-deps
 
   python-deps:
-    source: https://github.com/getting-things-gnome/liblarch.git
     plugin: python
     build-packages:
       - python3-dbus
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -33,7 +33,6 @@ jobs:
         python -m pip install --upgrade pip
         sudo apt-get update
         sudo apt install -y meson gettext itstool libgirepository1.0-dev gir1.2-gtk-4.0 libgtksourceview-5-dev libportal-dev
-        pip install --user -e git+https://github.com/getting-things-gnome/liblarch.git#egg=liblarch
         pip install --user pytest pycairo PyGObject caldav lxml
 
     - name: Build and install GTG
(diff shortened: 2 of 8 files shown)


picsel2 (commented):
Just two nitpicks:

1. The snap manifest still uses liblarch: https://github.com/SqAtx/gtg/blob/cae22e1cbaa2859e972e4e615cf0113337de8c70/snap/snapcraft.yaml#L51
2. There is some incorrect type hint using liblarch in MainWindow: https://github.com/SqAtx/gtg/blob/cae22e1cbaa2859e972e4e615cf0113337de8c70/GTG/gtk/browser/main_window.py#L797

SqAtx:
Thanks; I've updated these two places.

liblarch is mentioned in a few other places. I thought we could remove these mentions when we come across them, as I'm not 100% sure how to deal with all of them. My goal here was mostly to remove it from the README and the list of dependencies.

SqAtx:
> I think you had merge rights?

I do not :) 

I think this can be merged. I haven't re-tested it, but the only changes I made since my last tests are to comments, type hints, and the Snap packaging.

snap/snapcraft.yaml
```diff
@@ -46,13 +46,3 @@ parts:
         -e 's|Icon=.*|Icon=${SNAP}/usr/share/icons/hicolor/scalable/apps/org.gnome.GTG.svg|'
     after:
       - python-deps
-
-  python-deps:
-    source: https://github.com/getting-things-gnome/liblarch.git
```
picsel2, line 51: I have absolutely no experience with snap yet I think you removed here to much. I guess only line 51 (`source: [...]` ) it should be removed here.

picsel2 (changes requested):

snap/snapcraft.yaml
picsel2: Maybe @sergiusens would be so kind to check the snapcraft file? He was the last one to contribute to it.

SqAtx: Hm yeah it's possible. The other lines kind of look like the dependencies of liblarch, but they may be dependencies of GTG.

I'm not super fussed about this file because I don't think we release snaps regularly, and I don't think we've done it since the GTK4 refactor. So there is a good chance it doesn't work at all.

Happy to go with your suggestion, but if there is further debate I'd like to defer this work and just not touch the file.

diegogangl:
LGTM, merging. Thanks!

PR #1146, opened 2024-09-22
Suppress mismatched signature errors in type-checking

gycsaba96:
The only purpose of the `BaseStore.new` and `BaseStore.from_xml` methods is to provide an overview of how to use the subclasses. However, the mismatched signatures caused typing errors. Suppressing these errors instead of simply deleting the base class methods can still provide this "educational" benefit.

Other options would be:
- to use an extra generic parameter in these methods (looks a bit over-complicated...)
- to use composition instead of inheritance (more elegant, but this would probably break existing code and established interfaces)


diegogangl:
This sounds like the best solution for me (without having to rewrite). Thanks!

PR #1147, opened 2024-09-29
Fix typing errors in system_info.py

gycsaba96:
- import `importlib.util` directly to please the type checker
- handle case when `flatpack-version` is not present
- fix the return type of `__get_gtk_version`


diegogangl:
Merged, thanks!

PR #1148, opened 2024-10-06
Small type fixes for core

gycsaba96:
- add simple missing type annotations
- suppress errors for untyped parts


diegogangl:
Merged, thanks!

PR #1149, opened 2024-10-17
Prevent reparenting a task to itself

gycsaba96:
It was possible to reparent a task to itself using drag-and-drop. `TaskPane.check_parent` now checks if the drop target is the same task we are moving.

1. Pick a task and start dragging it over another task.
2. Move it back over itself, and drop it.
3. Observe as the task disappears instead of not doing anything.

By repeating these steps, you can make all of your tasks disappear. They are still in the `gtg_data.xml`, but there is no way accesing them.


diegogangl:
LGTM, thanks!

Issue #1051, opened 2024-03-04
GTG does not use prepared datasets


kocio-pl:
Launching GTG with `-s` should invoke two pre-made datasets, as mentioned in the README: `bryce` and `screenshots`. GTG creates datasets with requested name, but does not fill it with prepared data.

picsel2:
How did you launch GTG? Did you use the `launch.sh` script?

kocio-pl:
Yes, this is how I'm testing current version.

picsel2:
It looks like current GTG is searching for a file named `gtg_data.xml`. The datasets in `data/test-data/` looks like they are in an older data format.

This leads to GTG executing `firstrun_tasks.generate` unexpectedly.

nekohayo:
Yeah, those datasets are in the old v1 (well, unversioned) format and haven't been fixed up and recommitted in clean v2 (0.5/0.6?) format,  nor "converted to what 0.7 is" (if 0.7 actually is a different format? I'm not sure). I would actually like them to be present in all versions of the format, for (regression) testing purposes, if possible. @diegogangl does current git `master` / 0.7 use a different file format, or it remains strictly the same thing as the "v2" format?

kocio-pl:
I almost finished converting script, but it looks like `bryce` dataset has multiple parents problems, which caused #855 warnings in the past and now trying to open this dataset fail because of them.

kocio-pl:
The initial script is here: 
https://github.com/kocio-pl/gtg/blob/old_datasets_conversion_script/scripts/convert_old_datasets.py

When run in the folder with old data files, it builds `gtg_data.xml` which is not too pretty, but both `bryce` and `screenshots` at least are opening with current app code. 

Known problems:
- the tag counters are not always accurate (which might be at least in part a problem with app code, who knows?)
- some tasks include unconverted subtasks and tags in the content field (`screenshots`)
- task with multiple parents in `bryce` are just mechanically deparented, so no task has more than 1 parent now, but it would be good to check that no task has lost all the parents it had before

There might be even more problems with old datasets, but it's hard to say how much we want to care for them.

pentux-GitHub:
Housekeeping: this is fixed on current master — `scripts/debug.sh` (lines 52-58) copies the named dataset from `data/test-data/` into `tmp/` on first use, and both `bryce` and `screenshots` are present there. Verified working (the bryce dataset was used extensively for the recent CalDAV and performance work). It seems safe to close.

PR #1150, opened 2024-10-19
Type fixes for versioning.py

gycsaba96:
The changes are as follows:
- use the `_Element`/`_ElementTree` class instead of the `Element`/`ElementTree` factory for typing,
- expect `None` values returned by lxml,
- other minor fixes.

Two unresolved lines still reference the `core/xml.py` file removed in e43eee0. This doesn't cause runtime errors because the code in `versioning.py` is currently unreachable (see PR #1133; this is possibly the reason behind issue #1051).

### DIFF
--- a/GTG/core/versioning.py
+++ b/GTG/core/versioning.py
@@ -37,7 +37,7 @@ tags_cache = {}
 tid_cache = {}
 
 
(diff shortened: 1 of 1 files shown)


diegogangl:
Merging, thanks!

PR #1151, opened 2024-11-03
Prevent reparenting a tag to itself

gycsaba96:
Similarly to PR #1149, it was possible to reparent a tag to itself using drag-and-drop. `Sidebar.check_parent` now checks if the drop target is the same tag we are moving.

1. Pick a tag and start dragging it over another tag.
2. Move it back over itself, and drop it.
3. Observe the program freezing and eventually running out of memory.

(If it is not against any current plans, I would be happy to experiment with a common `StoreItem` base class or something similar to prevent such duplicated bugs.)

### DIFF
--- a/GTG/gtk/browser/sidebar.py
+++ b/GTG/gtk/browser/sidebar.py
@@ -591,7 +591,10 @@ class Sidebar(Gtk.ScrolledWindow):
 
 
     def check_parent(self, value, target) -> bool:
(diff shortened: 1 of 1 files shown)


diegogangl:
Merged, thanks!

> (If it is not against any current plans, I would be happy to experiment with a common StoreItem base class or something similar to prevent such duplicated bugs.)

Sure, the only plan right now is to stabilize and release 0.7

PR #1152, opened 2024-11-08
Ensure consistent indentation when displaying trees

gycsaba96:
On Fedora 40 and 41, the GUI of the tree representation appears too flat. This is because the `indent` elements of the [`TreeExpander`](https://docs.gtk.org/gtk4/class.TreeExpander.html#css-nodes) are not wide enough. (It worked perfectly on Fedora 39.) 

gtg-indentation

Explicitly setting the sizes fixes the problem. (It also looks good on Fedora 39.)


diegogangl:
Thanks for catching this! Merged :)

PR #1156, opened 2024-11-11
Minor type fixes for core

gycsaba96:
The changes are as follows:
- Add some missing type hints. (It looks like `Plugin.missing_dbus` is never actually updated, maybe it can be removed once we are sure that we don't need it. The current type hint is based on an older code version.)
- Add a typing stub for `core/info.py.in`.
- Introduce `TMP_KEYWORDS` to avoid changing types.
- Simplify `TaskTagSorter.get_first_letter` to resolve a missing return error. (Based on my micro-benchmarks, it is still performant.)

### DIFF
--- a/GTG/backends/generic_backend.py
+++ b/GTG/backends/generic_backend.py
@@ -28,6 +28,7 @@ import os
 import pickle
 import threading
 import logging
+from typing import Dict, Any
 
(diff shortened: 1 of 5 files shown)


diegogangl:
LGTM, Thanks!

PR #1158, opened 2024-11-16
Introduce the StoreItem base class

gycsaba96:
This is a followup to PR #1151 removing duplicated functionality. The main benefits include:
- reduced code duplication,
- isolated testing of common functionality,
- the possibility of a more targeted testing of the `BaseStore` class (which I would be happy to do).

- introduced a new class that implements common functionality for items in a `BaseStore`,
- added corresponding unit tests,
- fixed a typo in the copyright text.

### DIFF
--- a/GTG/core/base_store.py
+++ b/GTG/core/base_store.py
@@ -42,8 +42,8 @@ class StoreItem(GObject.Object):
     def __init__(self,id: UUID):
         self.id: UUID = id
         self.parent: Optional[Self] = None
-        self.children: List[Self] = []
-        super(StoreItem, self).__init__()
+        self.children: list[Self] = []
+        super().__init__()
 
 
     @GObject.Property(type=int)
@@ -67,7 +67,7 @@ class StoreItem(GObject.Object):
         return ancestors
 
 
-    def check_possible_parent(self, target) -> bool:
+    def is_parentable_to(self, target) -> bool:
         """Check for parenting an item to its own descendant or to itself."""
         return self != target and self not in target.get_ancestors()
 
--- a/GTG/core/base_store.py
+++ b/GTG/core/base_store.py
@@ -25,7 +25,7 @@ from uuid import UUID
 import logging
 
 from lxml.etree import _Element
-from typing import Dict, List, Optional, TypeVar, Generic
+from typing import Dict, Optional, TypeVar, Generic
 from typing_extensions import Self
 
 
@@ -57,9 +57,9 @@ class StoreItem(GObject.Object):
         return len(self.children) > 0
 
 
-    def get_ancestors(self) -> List[Self]:
+    def get_ancestors(self) -> list[Self]:
         """Return all ancestors of this tag"""
-        ancestors: List[Self] = []
+        ancestors: list[Self] = []
         here = self
         while here.parent:
             here = here.parent
@@ -79,7 +79,7 @@ class BaseStore(GObject.Object,Generic[S]):
 
     def __init__(self) -> None:
         self.lookup: Dict[UUID, S] = {}
-        self.data: List[S] = []
+        self.data: list[S] = []
 
         super().__init__()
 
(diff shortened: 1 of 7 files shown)


GTG/core/base_store.py
```diff
@@ -25,17 +25,52 @@
...
+class StoreItem(GObject.Object):
+    """Base class for items in BaseStore."""
+
+    __gtype_name__ = 'gtg_StoreItem'
+
+
+    def __init__(self,id: UUID):
+        self.id: UUID = id
+        self.parent: Optional[Self] = None
+        self.children: List[Self] = []
```
SqAtx, line 45:
```suggestion
        self.children: list[Self] = []
```
Now that Python 3.8 is gone, we can write code that targets 3.9, which means we no longer need `typing.List` and the like.

GTG/core/base_store.py
```diff
@@ -25,17 +25,52 @@
...
+        super(StoreItem, self).__init__()
```
SqAtx, line 46:
```suggestion
        super().__init__()
```
Bad habits form Python 2 still in the code base :) Let's fix them as we see them.

GTG/core/base_store.py
```diff
@@ -25,17 +25,52 @@
...
+        """Return all ancestors of this tag"""
+        ancestors: List[Self] = []
+        here = self
+        while here.parent:
+            here = here.parent
+            ancestors.append(here)
+        return ancestors
+
+
+    def check_possible_parent(self, target) -> bool:
```
SqAtx, line 70:
```suggestion
    def is_parent_of(self, target) -> bool:
```
and inverse the conditions below.

I think it might be clearer to read `if not foo.is_parent_of(bar):` rather than `if foo.check_possible_parent(bar):`?

tests/core/test_store_item.py
```diff
@@ -0,0 +1,144 @@
...
+        self.assertEqual(item.id, uuid)
+
+
+    def test_children_count_default_value(self):
+        item = StoreItem(uuid4())
+        self.assertEqual(item.children_count, 0)
+        self.assertFalse(item.has_children)
+
+
+    def test_children_count_singel_child(self):
```
SqAtx, line 40:
```suggestion
    def test_children_count_single_child(self):
```

SqAtx (approved):
I think it's a good improvement! Just some minor comments.

GTG/core/base_store.py
gycsaba96: Done. (We should also update the readme and every other reference to python version.)

GTG/core/base_store.py
gycsaba96: Done.

tests/core/test_store_item.py
gycsaba96: Done.

GTG/core/base_store.py
gycsaba96: I agree that the name is not perfect. However, `is_parent_of` wouldn't correctly describe the method (e.g., you can not parent an item to itself.) I used `is_parentable_to` in the new version. I think it also reads much better than the original one.

GTG/core/base_store.py
```diff
@@ -25,17 +25,52 @@
...
+        """Read only property."""
+        return len(self.children)
+
+
+    @GObject.Property(type=bool, default=False)
+    def has_children(self) -> bool:
+        return len(self.children) > 0
+
+
+    def get_ancestors(self) -> List[Self]:
```
diegogangl, line 60: If we are moving away from typing.List, this can also be replaced with list

GTG/core/base_store.py
```diff
@@ -25,17 +25,52 @@
...
+        """Return all ancestors of this tag"""
+        ancestors: List[Self] = []
```
diegogangl, line 62: ...and this

diegogangl (approved):
There's a few new usages of typing.List left. Setting this to approved anyways, since it's not crucial to this PR. 
Let me know if you would like to change those, otherwise we can merge right away

gycsaba96:
> There's a few new usages of typing.List left. Setting this to approved anyways, since it's not crucial to this PR. Let me know if you would like to change those, otherwise we can merge right away

All replaced now in the `base_store.py` file. :)

diegogangl:
LGTM now, thanks!

PR #1159, opened 2024-11-16
Fix crash when deleting a task and its descendant together

gycsaba96:
When deleting a task and one of its descendants simultaneously, the program crashes. It happens because it tries to remove the descendant twice: once as a selected task and once as the descendant of the other task.

2. Expand the first task (_"Getting Started with GTG (read me first)"_).
3. Select the parent tag and its children.
4. Try to delete them using simultaneously.
5. Observe the crash.

When deleting multiple tasks simultaneously, we should check that a given task was not already removed as the descendant of a previously removed task.

### DIFF
--- a/GTG/core/base_store.py
+++ b/GTG/core/base_store.py
@@ -136,8 +136,8 @@ class BaseStore(GObject.Object,Generic[S]):
         self.emit('removed', str(item_id))
 
 
-    def batch_remove(self,item_ids: List[UUID]):
-        """Remove multiple items and their descendants at the same time. """
+    def batch_remove(self,item_ids: List[UUID]) -> None:
+        """Remove multiple items, ensuring nothing gets deleted twice"""
         for key in item_ids:
             if key in self.lookup:
                 self.remove(key)
--- a/GTG/core/base_store.py
+++ b/GTG/core/base_store.py
@@ -136,6 +136,13 @@ class BaseStore(GObject.Object,Generic[S]):
         self.emit('removed', str(item_id))
 
 
+    def batch_remove(self,item_ids: List[UUID]):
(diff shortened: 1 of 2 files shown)


GTG/core/base_store.py
```diff
@@ -136,6 +136,13 @@ def remove(self, item_id: UUID) -> None:
         self.emit('removed', str(item_id))
 
 
+    def batch_remove(self,item_ids: List[UUID]):
+        """Remove multiple items and their descendants at the same time. """
```
SqAtx, line 140:
```suggestion
    def batch_remove(self,item_ids: List[UUID]) -> None:
        """Remove multiple items, ensuring nothing gets deleted twice"""
```
The method actually doesn't do anything about descendants - what it actually does it making sure `remove()` doesn't crash. So I'd like the docstring to convey that if you want to delete several items you need to call this method instead of calling `remove()` in a loop. There is probably a better way to phrase it.

SqAtx (approved):
I guess you didn't want to do the obvious
```
item = self.lookup.get(item_id)
if not item:
  return
```
probably because it might potentially hide problems instead of making us fix them?

I'm fine with that idea. LGTM - I'd just clarify the docstring.

gycsaba96:
> I guess you didn't want to do the obvious
> 
> ```
> item = self.lookup.get(item_id)
> if not item:
>   return
> ```
> 
> probably because it might potentially hide problems instead of making us fix them?

Exactly. :) Moreover, doing this would also require us to change the specification of `BaseStore.remove`, which could potentially break other parts of the code.

SqAtx (approved):

diegogangl:
LGTM, thanks!

PR #1162, opened 2024-12-10
Remove old List use introduced by another branch from base_store.py

gycsaba96:
PR #1158 aimed to remove all outdated `typing.List` references form the `base_store.py` file. However, PR #1159 introduced a new one in the meantime that was not considered by 1158. When 1158 was merged, it caused all sorts of errors (e.g., failing the GHA tests) because of the now missing `typing.List` import.


diegogangl:
Merged, thanks!

PR #1165, opened 2024-12-20
Tests and related bugfixes for BaseStore.remove

gycsaba96:
I wrote unit tests for BaseStore.remove and discovered 2 bugs along the way:
- When removing an item, the code only considered the children instead of removing all descendants. This resulted in the store being left in an inconsistent state. (This seems related to the scattered use of `BaseStore.refresh_lookup_cache` method.)
- When removing a tree of items, only the root element was reported with a `removed` signal. (If I am not mistaken, the code never relied on this signal, but let's fix it anyway.)

- write unit tests for `BaseStore.remove`,
- ensure that all descendants are recursively removed,
- ensure that a `removed` signal is emitted for each removed item,
- make `SavedSearch` a subclass of `StoreItem` to keep things readable.

### DIFF
--- a/GTG/core/base_store.py
+++ b/GTG/core/base_store.py
@@ -150,23 +150,17 @@ class BaseStore(GObject.Object,Generic[S]):
         """Remove an existing item from the store."""
 
         item = self.lookup[item_id]
+        parent = item.parent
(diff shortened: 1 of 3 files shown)


diegogangl:
LGTM, thanks!

PR #1166, opened 2024-12-29
Tests and a fix for BaseStore.parent

gycsaba96:
I wrote unit tests for `BaseStore.parent`. The tests revealed that calling the method with a non-existing `parent_id` changes the list of root elements (`self.data`).

- add unit tests for `BaseStore.parent`,
- ensure that the store is unchanged when called with invalid IDs,
- add helpful error messages to the exceptions.


diegogangl:
Ah that was before I understood the importance of keeping try/excepts 1-line long :)
LGTM, thanks!

PR #1168, opened 2025-01-03
Remember unselecting tags + disable saved search auto-selection

### LINKED ISSUE #1130: Last selected tag remembered even after it should have been cleared
~From the Open or Actionnable tab, type something in the "Quickly create tasks here" box. No new task appears unless you change tabs and back again.~

~I will run a git bisect when I have a minute.~

The problem was actually different.

even if I close GTG like this, with All Tags selected:
image
When I reopen it (via `launch.sh`), `mytag` is selected:
image
If I select `mytag2` and restart GTG, `mytag2` comes up. After that, if I select All Tags and crestart GTG, `mytag2` comes up again.

It seems to just be a matter of clearing the config value when All Tags is selected, evne though db6392318d037ffc95dc7e6de6d0080407af5875 looks like it's doing the right thing.

I haven't confirmed whether this is specific to the dev environment, or if it also happens with the flatpak version.

I'm on f501dfbf6a618d0e0425ed25ae8dab37694e407e

gycsaba96:
Fixes #1130 

- updated the config both when a tag is selected and when tags are unselected,
- disable auto-selection for the saved searches.

The second point is unrelated to the original bug. While testing my solution, it turned out that there was no way of unselecting saved searchs because of the enabled auto-selection.

### DIFF
--- a/GTG/gtk/browser/sidebar.py
+++ b/GTG/gtk/browser/sidebar.py
@@ -405,6 +405,7 @@ class Sidebar(Gtk.ScrolledWindow):
 
     def unselect_tags(self) -> None:
         """Clear tags selection"""
+        self.browser.config.set("selected_tag", '')
 
         with signal_block(self.tag_selection, self.tag_handle):
             self.tag_selection.unselect_all()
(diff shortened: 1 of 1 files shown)


diegogangl:
Merged, thanks!

Issue #1164, opened 2024-12-13
Update AppData metadata for Flathub and packages to point to the new website URL


nekohayo:
As per https://github.com/getting-things-gnome/gtg/issues/1092#issuecomment-2430398335 the website URL changed from `https://wiki.gnome.org/Apps/GTG/` to https://getting-things-gnome.github.io/ and we should make sure to update that [throughout the repository here](https://github.com/search?q=repo%3Agetting-things-gnome%2Fgtg+wiki.gnome&type=code) so that the app (README, About dialog, etc.) and metadata (ex: on https://flathub.org/apps/org.gnome.GTG …) points to the correct location.

PR #1170, opened 2025-01-10
Update website link in README.md

gerbrent:


Neui:
There are other references to the wiki left (incuding at the bottom of the README.md file): https://github.com/search?q=repo%3Agetting-things-gnome%2Fgtg+wiki.gnome.org&type=code

Maybe you could also change these.

SqAtx (approved):
Agreed - that's #1164. But this is already an improvement (thanks!) so we can merge it while we work on the rest.

diegogangl:
Merged, thanks!

PR #1171, opened 2025-01-10
Test and refactor BaseStore.unparent

gycsaba96:
I added unit tests for `BaseStore.unparent` and did a bit of refactoring. The only minor thing it revealed was that a `KeyError` was raised even when both IDs existed, but the corresponding items were not in a child-parent relation. 

- add unit tests for `BaseStore.unparent`,
- use `ValueError` instead of `KeyError` when both parameters exist but are not in a child-parent relation,
- simplify the algorithm.

**Note:** `BaseStore.unparent` doesn't actually need the `parent_id` as an extra input since it can be derived as `child.parent.id`. Unparenting is used in around 20 places, so the argument should be relatively easy to remove. Are there any plans or design decisions that could influence this?

### DIFF
--- a/GTG/core/base_store.py
+++ b/GTG/core/base_store.py
@@ -194,19 +194,21 @@ class BaseStore(GObject.Object,Generic[S]):
 
     def unparent(self, item_id: UUID, parent_id: UUID) -> None:
         """Remove child item from a parent."""
+        if item_id not in self.lookup:
+            raise KeyError("item_id is not in store: "+str(item_id))
+        if parent_id not in self.lookup:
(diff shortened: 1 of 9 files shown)


diegogangl:
No plans or anything really, IIRC items didn't have a parent property (only children) in the very early versions of the new core. Less parameters is better though. Would you like to do it in this branch or another PR?

gycsaba96:
Let's do it in one PR.

I removed the `parent_id` argument and adjusted the code and the tests. If the item has no parent, `unparent` has no effect. (I also renamed some test classes to keep things consistent.)

diegogangl:
Looks good, thanks!

PR #1173, opened 2025-01-19
Fix GTK version numbering in system_info.py

gycsaba96:
The changes are as follows:
- present the GTK version numbers in "major.minor.micro" order,
- adjust a related docstring.


SqAtx (approved):
Good catch :) Odd mistake, too

diegogangl:
Thanks!

PR #1175, opened 2025-01-20
Save task before closing editor

### LINKED ISSUE #1138: New tasks sometimes don't get saved
```
rm -rf .local_build/
./launch.sh
```
Create a task using the + in the top left
Upon closing the modal, the task doesn't get created.

The Quick Add at the bottom works.

I'll `git bisect` to figure out when this broke.

GTG: 0.7-dev-8a8d7497
Flatpak: False
Snap: False
Display Name: wayland-0
Desktop: ubuntu:GNOME

lxml: True
caldav: True
liblarch: False
Cheetah3: False
dbus-python: False
pdflatex: False
pypdftk: False

OS: Ubuntu 24.04 LTS
Python: CPython 3.12.3
GLib: 2.80.0
PyGLib: 3.48.2
PyGObject: 3.48.2
GTK: 4.2.14

SqAtx:
Fixes #1138

When the task editor gets closed, it first calls
`self.textview.process()` which, after processing, calls `self.save_cb()`, which is set to `TaskEditor.light_save()`. This only calls `TaskEditor.save()` at most every 7 seconds. So if you crete a new task and close its editor within 7 seconds, the task doesn't get saved.

This commit makes `TaskEditor.desctruction()` call `TaskEditor.save()`, to make sure the task is saved properly before closing the task popup.

### DIFF
--- a/GTG/gtk/editor/editor.py
+++ b/GTG/gtk/editor/editor.py
@@ -84,7 +84,7 @@ class TaskEditor(Gtk.Window):
 
         self.app = app
         self.ds = app.ds
-        self.task = task
+        self.task: Task = task
         self.config = app.config_core
         self.task_config = self.config.get_task_config(str(task.id))
         self.time = None
@@ -778,7 +778,7 @@ class TaskEditor(Gtk.Window):
         self.time = time.time()
 
 
-    # light_save save the task without refreshing every 30seconds
+    # light_save save the task without refreshing every 30 seconds
     # We will reduce the time when the get_text will be in another thread
 
     def light_save(self):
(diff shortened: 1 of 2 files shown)


GTG/gtk/editor/editor.py
```diff
@@ -836,13 +836,16 @@ def destruction(self, _=None):
...
+        # isn't guaranteed to actually save (see #1138)
+        self.save()
+
         # self.pengine.onTaskClose(self.plugin_api)
         # self.pengine.remove_api(self.plugin_api)
 
         tid = self.task.id
 
-        if self.is_new():
+        if self.task.is_new():
```
SqAtx, line 848:
```
    def is_new(self) -> bool:
        return self.task.is_new()
```
I wouldn't mind getting rid of that method eventually.

SqAtx (commented):
I'd love to start adding tests for this type of thing, but we don't currently have tests for GTK components, so it's going to take some time to learn how and start writing them.

GTG/gtk/editor/editor.py
diegogangl: Definitely, but watch out for other uses. Check out application.py at line 436

SqAtx: Oh yeah I don't mean to just delete it now, just that it is now a wrapper around something that makes more sense, so we could replace the usages and eventually remove the method.

PR #1177, opened 2025-01-20
Small ruff fixes

SqAtx:
This fixes a new issues found by `ruff check`.

There are a few more, but they need to be looked into individually. They are instances of the code being left in an inconsistent state.

### DIFF
--- a/GTG/core/plugins/api.py
+++ b/GTG/core/plugins/api.py
@@ -206,7 +206,7 @@ class PluginAPI():
         NOTE: This function stronglye depend on browser and could be easily
         broken by changes in browser code
         """
-        browser = self.get_browser()
+        # browser = self.get_browser()
 
         # set default bgcolor?
         # if func is None:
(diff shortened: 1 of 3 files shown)


GTG/core/plugins/api.py
```diff
@@ -206,7 +206,7 @@ def set_bgcolor_func(self, func=None):
         NOTE: This function stronglye depend on browser and could be easily
         broken by changes in browser code
         """
-        browser = self.get_browser()
+        # browser = self.get_browser()
```
diegogangl, line 209: Why comment this?

SqAtx: Because Ruff is complaining about the unused variable. I thought the line went with the below commented out code, so I figured commenting it out made sense?

It may be the wrong thing to do - I didn't check `get_browser()` for side effects

diegogangl: It breaks hamster.py which uses `api.browser` (line 317) but it could just be replaced with get_browser()

SqAtx: The assignment being commented out is that of a variable local to `set_bgcolor_func()` - not of `PluginAPI.browser`.

The line being commented out isn't actually doing anything.

diegogangl:
Merged, thanks!

PR #1163, opened 2024-12-11
Bump runtime to 47

TheEvilSkeleton:
Fixes https://github.com/getting-things-gnome/gtg/issues/1155

Fixes https://github.com/getting-things-gnome/gtg/issues/1179

PR #1178, opened 2025-01-20
Add typing_extensions to flatpak requirement

SqAtx:
Without it, the weekly flatpak fails on startup:

```
  File "/app/bin/gtg", line 98, in <module>
    from GTG.gtk.application import Application
  File "/app/lib/python3.11/site-packages/GTG/gtk/application.py", line 30, in <module>
    from GTG.core.datastore import Datastore
  File "/app/lib/python3.11/site-packages/GTG/core/datastore.py", line 30, in <module>
    from GTG.core.tasks import TaskStore, Filter
  File "/app/lib/python3.11/site-packages/GTG/core/tasks.py", line 35, in <module>
    from GTG.core.base_store import BaseStore, StoreItem
  File "/app/lib/python3.11/site-packages/GTG/core/base_store.py", line 29, in <module>
    from typing_extensions import Self
ModuleNotFoundError: No module named 'typing_extensions'
```

The requirement has been necessary since 908da7d6ed11c01f99bcb31872ab96afad31ea83

### DIFF
--- a/build-aux/requirements.txt
+++ b/build-aux/requirements.txt
@@ -3,3 +3,4 @@ caldav==1.3.9
 pytest
 Cheetah3
 dbus-python==1.2.18
+typing_extensions


SqAtx:
Note that I haven't been able to test this yet

gycsaba96:
> Note that I haven't been able to test this yet

I am not yet that familiar with the flatpak build pipeline. However, it seems the _python3-requirements.yaml_ file should also be updated based on [this comment](https://github.com/getting-things-gnome/gtg/blob/3910d84e87040b40cf2926fc8f143f324810ee94/build-aux/python3-requirements.yaml#L1). Another option could be to leverage [`typing.TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) and treat `typing_extensions` as an optional dependency not included in the flatpak build.

(If I understand correctly, this will be a quick fix before [PR 1163](https://github.com/getting-things-gnome/gtg/pull/1163) is merged.)

SqAtx:
You might be right - I tried to build it and it still didn't work. I'll try and follow your suggestion tomorrow and update this if it works.

gycsaba96:
> Hm I haven't been able to make that command work.

Maybe we are facing the same error: `error: app/org.gnome.Sdk/x86_64/46 not installed`
Installing the SDK solved the problem: `flatpak install org.gnome.Sdk/x86_64/46`

Again, using [`typing.TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) seems also a viable solution that keeps the flatpak cleaner.

Or we can just wait until PR #1163 is merged. :)

SqAtx:
Yep, both solutions are good. Unfortunately, it doesn't seem that #1163 results in a working flatpak just yet.

If we can't get that resolved quick enough, we can go the `typing.TYPE_CHECKING` route.

Note that Python 3.11 also introduced `typing.Self`, which will remove the need for `typing_extensions` (at least our current usage). So that's another way of fixing the problem.
 
I don't think we have formalized how we pick which versions of Python we support, because we don't check explicitly. Maybe the most reasonable thing to do is "GTG has to run on whatever version of Python is in the latest version of the SDK", and just make sure GHA uses that version to run the tests (and make sure GTG actually launches, actually). GHA currently uses 3.9, but if we want to keep life simple we should just figure out what version the latest GNOME SDK runs, and use that. I'll do that one day. Maybe right now, that would have enabled us to get away with just dropping Python<3.11, and remove the need for that dependency altogether.

I don't know. I'm just typing things now :p I'll look into it later. I'm certainly interested in keeping the weekly flatpak in working order.

Closing this in favour of #1163, or a future PR that just fixes the `typing_extensions` problem.

gycsaba96:
> Maybe the most reasonable thing to do is "GTG has to run on whatever version of Python is in the latest version of the SDK", [...]

I really like this idea. Consider opening an RFC issue about it.

(Based on [this discussion](https://discourse.gnome.org/t/upgrading-python-version-in-flatpak/23290), the SDK already has Python 3.11 and 3.12 for version 46 and 47.)

Issue #1180, opened 2025-01-24
Children get marked undone after marking parent as done


SqAtx:
From the Actionable tab, click on the + to create a new task, with the following contents:
```
parent

- child1
- child2
```
I see this in the Actionable tab:

![Image](https://github.com/user-attachments/assets/500c4b95-52b4-49a3-b865-5788fa73fbaf)
which is not correct but that's a known problem - see #1039

After clicking on Open and back on Actionable:

![Image](https://github.com/user-attachments/assets/0590a02b-272c-4f9d-bb1c-f98d90c3912c)

Now click on the checkboxes to mark the children as done:

![Image](https://github.com/user-attachments/assets/4a08ac9c-ce7f-44e8-88c2-7405b07b3e37)

They don't disappear even though they should, but that's again #1039

After clicking on Open and back on Actionable again, `parent` is visible. Open it:

![Image](https://github.com/user-attachments/assets/23b6679f-e312-4461-8868-65c97e15f278)

Click on Mark as Done - it becomes

![Image](https://github.com/user-attachments/assets/e08201f2-1a7e-4a09-b36e-453d5f717c86)

Close task. It still looks undone

![Image](https://github.com/user-attachments/assets/16b005d3-91df-4d56-b5a5-3b21b61729e8)

Let's blame the view again and click Open->Actionable. Now the children are undone again:

![Image](https://github.com/user-attachments/assets/cd7f411a-4d22-48b5-837d-2d23625ecd00)

Notes:
* If you close the parent by clicking its textbox without opening the editor, it actually gets closed
* there may be a slightly simpler repro if you just stay on the Open tab
* I haven't actually reproed on master yet, but I have no doubt. This is my version:
```
GTG: 0.7-dev-20241206
Flatpak: 1.14.10
Snap: False
Display Name: wayland-0
Desktop: GNOME

lxml: True
caldav: True
Cheetah3: True
dbus-python: True
pdflatex: False
pypdftk: False
```

PR #1183, opened 2025-02-02
Use Task.set_status to mark tasks as done or active

gycsaba96:
Fixes GitHub issue #1180

Somehow, setting the task done was triggered twice: in `TaskEditor.change_status` and in `MainWindow.on_mark_as_done`. Since `MainWindow.on_mark_as_done` also used `Task.toggle_status`, the effect was reversed.

**Some remarks:**
- I don't clearly see why was the `set_status` triggered twice, but since fixing `MainWindow.on_mark_as_done` fixes the problem, I would say it is a question for another day.
- Dismissing tasks may have similar problems, but I couldn't yet reproduce the bug (see `MainWindow.on_dismiss_task`).

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -1389,7 +1389,7 @@ class MainWindow(Gtk.ApplicationWindow):
 
     def on_mark_as_done(self, widget=None):
         for task in self.get_pane().get_selection():
-            task.toggle_active()
+            task.set_status(Status.DONE)
 
     def on_dismiss_task(self, widget=None):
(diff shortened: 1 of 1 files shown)


SqAtx (approved):
Thanks for looking at my bug!

I'm not intimate with that part of the code and I'm just looking at the diff, but it looks like you're making it so that the code does what we mean! It was making assumptions that it didn't need to make. LGTM!

diegogangl:
TBH I don't remember why I used toggle_active() instead of setting the status. Probably porting something from a previous version? Anyways, this LGTM. Thanks!

PR #1188, opened 2025-02-15
Convenience functions for adding and removing children

gycsaba96:
The main motivation behind this PR is to simplify the cases when class A emits a signal on class B because class C did something. E.g., the `Task` class can manage the update of its `has_children` property alone. (Overriding these methods can also help realising when a task is no longer actionable.)

- introduce the `StoreItem.add_child` and `StoreItem.remove_child` methods (automatically trigger signals and potentially other functionality)
- add unit tests for the methods above
- leverage these methods in the `BaseStore`, `TaskStore` and `TagStore`

### DIFF
--- a/GTG/core/base_store.py
+++ b/GTG/core/base_store.py
@@ -57,6 +57,26 @@ class StoreItem(GObject.Object):
         return len(self.children) > 0
 
 
+    def add_child(self,child:Self):
+        """Add an item as child if not yet present."""
+        if child in self.children:
+            return
+        self.children.append(child)
+        self.notify('children_count')
+        if len(self.children) == 1:
+            self.notify('has_children')
+
+
+    def remove_child(self,child:Self):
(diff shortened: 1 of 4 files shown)


diegogangl:
I'm testing this with "Add Subtask" and "Add Parent Task" from the RMB menu, and it looks like the new tasks dissapear after closing the task editor?

gycsaba96:
I rebased the branch so that it has PR #1175 in it. It seems to solve the "Add Subtask" part. Unfortunately, I can not reproduce the problem with the "Add Parent Task" option. 

(However, the editor does not open when the task already has a parent. But this bug was present in earlier versions as well.)

diegogangl:
Hmmm I've nuked all the previous data, and now I can't reproduce the parent bug either. Maybe something got corrupted after all the testing. Anyways, this is an improvement and works correctly. Merged, thanks!

PR #1191, opened 2025-03-08
Introduce ActionRow widget and port to them

TheEvilSkeleton:
This makes the rows work better in the preferences dialog (they should now react when clicking on them)

### DIFF
--- a/GTG/gtk/action_row.py
+++ b/GTG/gtk/action_row.py
@@ -1,0 +1,74 @@
+# -----------------------------------------------------------------------------
+# Getting Things GNOME! - a personal organizer for the GNOME desktop
+# Copyright (c) 2025 - The GTG Team
+#
+# This program is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free Software
+# Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program.  If not, see <http://www.gnu.org/licenses/>.
+# -----------------------------------------------------------------------------
+
+from gi.repository import GObject, Gtk
+from typing import Any
+
+from GTG.gtk.browser import GnomeConfig
+
+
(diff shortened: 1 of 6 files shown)


TheEvilSkeleton:
Marking it as draft as there are a couple of bugs when opening the inspector.

diegogangl:
Looks interesting, let us know when we can review :)

TheEvilSkeleton:
> Looks interesting, let us know when we can review :)

I figured it out thanks to this comment: https://discourse.gnome.org/t/invoking-virtual-functions-defined-on-an-interface-via-pygobject/27453/2

I just needed to make it inherit `Gtk.Buildable`

nekohayo:
Did a quickie smoke-test: this does not change the visual appearance of those rows in the preferences dialog, it simply makes them work properly when you click on the row itself rather than the switch. From that point of view, this MR works.

However, why does this change a bunch of private (`_` prefixed) Python methods into public ones? (and if it's truly necessary, why not a separate commit for that before the rest?)

TheEvilSkeleton:
Uh, I'm not sure. I didn't even notice that change

TheEvilSkeleton:
> Interesting, so an automated replacement accident must have happened in the changes you committed in the 2nd commit, mainly in `gtk/general_preferences.py`

No, I likely changed them subconsciously

nekohayo:
The rebased version's diff is simpler now, as the 2nd commit no longer changes all those private methods to public. Functionally, it still works in practice. I haven't reviewed the other parts of the diff, but I bet others here will be able to do so more effectively :)

nekohayo:
Hmm, maybe I didn't look properly, trying this latest version of the branch again, I get this traceback on startup:

```python-traceback
  File "/app/lib/python3.12/site-packages/GTG/gtk/general_preferences.py", line 89, in _refresh_preferences_store
    self._preview_switch.set_active(show_preview)
AttributeError: 'NoneType' object has no attribute 'set_active'
```

TheEvilSkeleton:
> Hmm, maybe I didn't look properly, trying this latest version of the branch again, I get this traceback on startup:
> 
> ```
> Traceback (most recent call last):
>   File "/app/lib/python3.12/site-packages/GTG/gtk/general_preferences.py", line 89, in _refresh_preferences_store
>     self._preview_switch.set_active(show_preview)
>     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'NoneType' object has no attribute 'set_active'
> ```

Should be fixed now

Issue #608, opened 2021-04-02
Automatic night mode and dark styles support


nekohayo:
Since GTG 0.5 has a dark mode setting in the preferences, I guess I should point out that it ought to have an accompanying "Automatic" dark/night mode feature available, where the dark mode gets turned on automatically at sunset and off at sunrise. Because manually managing this is "not the user's job" (unless they _want_ to have the same theme no matter what the hour), in my opinion.

~~The basic/primitive way to do this would be to use the app's internal clock (since it already knows the time, using that to update the dates at midnight etc.), but~~ the "optimal" way would be, when available on the system, to use GNOME's night mode (connecting over DBus or something?) so that it happens with the rest of the desktop and benefits from its heuristics (as it takes into account timezone/DST/seasons/etc., I think) and/or plugging into the new Freedesktop dark mode standard (supported by GNOME 42+ and Elementary's Pantheon)

diegogangl:
Wait, gnome has a night mode?

nekohayo:
It doesn't control dark mode (I use my own silly script for that, which I've never taken the time to release publicly and announce, and it's a silly hack), but it does control what we used to call "red shift", so that's something I imagine you could plug into: in the Control Center, in the "Displays" panel, "Night Light" section.

diegogangl:
Ah ok, there was some talk about auto-changing apps to dark mode but I didn't know if it was eventually added.

For anyone attempting this, look for `toggle_darkmode()` in `application.py`. You also need to refresh the treeviews for the background colors, so look at the darkmode toggle code in `general_preferences.py`

azmeuk:
To add some context: there is (at least) [one GNOME extension](https://extensions.gnome.org/extension/2236/night-theme-switcher/) that switches themes at sunset/sunrise. Most of the GTK apps works very well with that, and some apps have an "automatic theme detection" so non-GTK content matches the global theme (for instance, the [dark reader](https://addons.mozilla.org/fr/firefox/addon/darkreader/) firefox extension con only awake when the global gtk theme is dark).

The extension can also trigger user-written scripts at theme switch, this is how I automatically switch my vim and alacritty themes.

nekohayo:
[...]

Issue #576, opened 2021-03-11
Hardcoded grey / gray color of description preview text doesn't play well with selected rows or themes


DonQuixoteI:
If the task is selected in the task browser, the grey color of description preview is difficult to read on top of blue background in the default white mode.

It will be easier to read if the description preview and the task counters in the sidebar (same permanent grey color) will have the same colors as tasks, tags and dates: black if not selected and white if selected.

The readability of description (for selected task) is a bit better in the dark mode, but the black icons are very difficult to see in case of dark grey background of dark mode.

nekohayo:
Is it hardcoded as grey in the code, or is it using transparency? (if possible, I would think transparency would work in all cases?)

I'd prefer automatic solutions (not options), if transparency can be used when not selected and full opacity when the row is selected... but then @diegogangl wanted to redo the whole widgetry so unless there's a trivial fix it would probably be moot to address that before then...

PR #1194, opened 2025-03-27
Adapt to user's color-scheme preference

TheEvilSkeleton:
Closes https://github.com/getting-things-gnome/gtg/issues/608


nekohayo:
This removes the "dark mode" toggle option entirely from the preferences. While I might miss it a bit, I can understand the choice to make it fully automatic only.

Overall it seems to work. The task editor's text view canvas is dark at all times, for some reason, but that is not specific to this branch, `master` has that too: #1196

I was not able to test whether this solves #576, because for some reason the "Show description preview" setting doesn't work in this version (and `master`), it never shows the text, so I don't know if the gray color would be correct in both variants...

GTG/gtk/editor/taskview.py
```diff
@@ -963,3 +961,20 @@ def format_update(cls, text: str) -> str:
...
+
+        if 'Adwaita' in scheme_ids:
+            scheme_name = 'Adwaita'
+        else:
+            scheme_name = 'classic'
+
+        if prefer_dark_theme:
+            scheme = manager.get_scheme(f'{scheme_name}-dark')
+        else:
+            scheme = manager.get_scheme(scheme_name)
```
subnut, line 978: Wouldn't it be simpler to write this like so -
```python
        scheme_name = 'Adwaita' if 'Adwaita' in scheme_ids else 'classic'
        scheme_name = f'{scheme_name}-dark' if prefer_dark_theme else scheme_name
        scheme = manager.get_scheme(scheme_name)
```

Or is this discouraged owing to possible incompatibility with older versions of Python?

TheEvilSkeleton: @diegogangl what are your thoughts on @subnut's suggestion?

SqAtx: Not Diego, but my 2 cents: I don't have a religious preference against ternary operators, but I prefer to only use them when just testing a boolean.

For example, this is fine:
```
scheme_name = f'{scheme_name}-dark' if prefer_dark_theme else scheme_name
```

But this:
```
scheme_name = 'Adwaita' if 'Adwaita' in scheme_ids else 'classic'
```
is harder to parse. I'd do a double take that I wouldn't do with a normal if statement.

So I prefer the version that's currently written, but it's not a strong preference.

nekohayo: I find the original 9-lines-long code much easier to read than the 3-lines version, and personally, code readability trumps compactness any time of the day. I personally would not have been able to easily understand the 3-lines version. In other words, [PEP-20](https://peps.python.org/pep-0020/)…

diegogangl: I agree with @nekohayo. While @subnut's code is cleaner, it's better to keep it as readable as possible

diegogangl:
LGTM, merging.
Thanks!

PR #1201, opened 2025-03-29
Rework search bar

TheEvilSkeleton:
The search bar's activation and deactivation are split into different methods, as the way the search bar should behave differently on a case-by-case basis.

This also makes use of modern APIs and methods provided by GTK, which should ease maintenance.

Fixes https://github.com/getting-things-gnome/gtg/issues/952

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -110,9 +110,6 @@ class MainWindow(Gtk.ApplicationWindow):
         self.config = app.config
         self.tag_active = False
 
-        # Timeout handler for search
-        self.search_timeout = None
-
         self.sidebar = Sidebar(app, app.ds, self)
         self.sidebar_vbox.append(self.sidebar)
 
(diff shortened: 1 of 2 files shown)


nekohayo:
This uses GTK4's new SearchEntry capabilities to do the same thing as our old GTK3-era search delay timer functions.

I've tested this branch now and it behaves exactly how I'd expect all of it to work. Functionally-speaking, it is perfect… everything… [down to the last minute details](https://www.youtube.com/watch?v=CpjH9M2SYsk)!

nekohayo:
Discussing with Hari a bit, it turns out there is a gotcha remaining to solve with the `Esc` key. If you press Escape, currently it clears the SearchEntry's contents on first press, and dismisses the searchbar on 2nd press… however:

* [ ] When the search entry is _not_ focused: `Esc` should not affect the searchentry and searchbar at all.

* [ ] When the search entry _is_ focused: we should toggle off search mode while still keeping the search query there for next use (to be recalled with the next `Ctrl+F`)

TheEvilSkeleton:
> * [ ]  When the search entry is _not_ focused: `Esc` should not affect the searchentry and searchbar at all.

This should be done in a separate PR, in my opinion. The bug is caused by the following line: https://github.com/getting-things-gnome/gtg/blob/6ad3da5a882f87e1b97189a859371ea58c845028/GTG/gtk/application.py#L304

However, removing that line will break many widgets and dialogs as they won't be able to close/obscure with the Esc key.

PR #1204, opened 2025-04-15
Make GTG look 10× better

TheEvilSkeleton:


diegogangl:
This does look better, but aren't window borders radius bigger in Gnome apps?

diegogangl:
That sounds reasonable, merged!

PR #1205, opened 2025-04-23
Fix automatic unit test dependencies

gycsaba96:
The GHA fails because of some dependency problem related to PyGObject. This can be avoided by specifying the exact version of PyGObject that matches the GNOME SDK.


diegogangl:
Looks good, thanks!

PR #1208, opened 2025-05-02
Do the filtering before building task trees for the UI 

### LINKED ISSUE #1207: Filtering of tasks doesn't play well with the tree structure
Tasks are organized in a tree structure where the deeper levels have a deeper indentation by default. However, when we filter this "full" tree, the program disregards that ancestors might have been filtered out. In particular:

- Tasks keep their original indentation, regardless of whether their ancestors are matching or not.
- Expanders are kept even when no descendants match the filtering.
- Deeply nested tasks are not matched if they are hidden before filtering.

A reasonable explanation is that the tree is first built and then filtered. It should be done the other way around: first filtering the tasks, and then building a tree from the matching tasks.

Detailed steps to reproduce the bug:

1. Create a deep task structure where only some of the tasks have a certain tag: `a @x > b > c > d @x > e`
2. Make sure that all expanders are open.
3. Filter on that tag.
4. Observe that task `d` keeps its original indentation instead of being aligned to the left.
5. Observe that task `d` still has an expander without matching children.
6. Hide and show the subtasks of `a` using its expander.
7. Observe that `d` is not shown anymore.

(These also work if you filter on the actionable items instead of a tag.)
 
https://github.com/user-attachments/assets/3eddda51-f715-4700-bada-961400dcf3eb

gycsaba96:
Fixes #1207 

The main idea is that the `TaskStore` provides pre-filtered `Gtk.TreeListModel` instances using the newly introduced `FilteredTaskTreeManager` class. This class performs the filtering before building the tree, thus producing correctly indented tree models. The `FilteredTaskTreeManager` also helps track when the expander should appear in the UI.

**Sidenotes:**
- When filtering based on a tag, the parents of matching tasks are also displayed. This behavior was already present before the changes (likely a bug in the filter).
- I believe it is possible to further simplify the `TaskStore` by relying on signals instead of direct calls to update the trees. But one step at a time. :)


gycsaba96:
I rebased the branch to trigger the updated GHA workflow.

diegogangl:
LGTM! And yes, signals would be more in line with GTK and probably more error-proof.
Thanks, merging this one

PR #1131, opened 2024-08-04
Fix sidebar updates for tags

gycsaba96:
Fixes #1115 

Thanks to this PR, the  `TagStore` keeps track of the models created for child tags and updates them as required triggering a GUI update (similarly to the `TaskStore`). This GUI update prevents the user from deleting a tag twice and crashing the program.

PR #1209, opened 2025-05-04
Remove deleted saved searches from the UI

### LINKED ISSUE #1005: Changes are not reflected in the Sidebar when creating or deleting tags outside the Task Editor, until restarting the app
Since the GTK4 port:

* If you create a task like `test @brand_new_tag` via the quick-add bar at the bottom of the task browser, it will not get added to the sidebar until restarting the app. It works correctly if typing the tag within a Task Editor window, so you probably just need to reuse that code.
* Right-click Deleting a tag from the sidebar doesn't have any visible effect after the confirmation dialog (removal throughout the GUI only happens after app restart)

Related: #1016

gycsaba96:
This closes #1005, which is originally about tags, and the saved search related part is only mentioned in the discussion). However, the tag related part was already fixed in PR #1131. 

The solution is simple: when a saved search is deleted, the corresponding `ListStore` is also updated.

### DIFF
--- a/GTG/core/saved_searches.py
+++ b/GTG/core/saved_searches.py
@@ -182,3 +182,14 @@ class SavedSearchStore(BaseStore):
         self.model.append(item)
 
         self.emit('added', item)
+
+
+    def remove(self, item_id: UUID) -> None:
+        """Remove an existing saved search."""
+
+        # Remove from UI
+        item = self.lookup[item_id]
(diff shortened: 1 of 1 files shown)


gycsaba96:
I rebased the branch to trigger the updated GHA workflow.

diegogangl:
Looks good, thanks!

PR #1210, opened 2025-05-23
Task filtering related fixes and simplifications

gycsaba96:
The original goal was to address some bugs in the task filtering functionality. The main changes are as follows:
-  Ensure that all required tags are matched separately (important  when filtering based on multiple tags with children tags).
- Allow filtering of untagged tasks in `SearchTaskFilter`.
- Do not count the tags of subtasks (causing to show immediate parents when filtering for tags).

After a bit of refactoring, it was clear that `TaskPaneFilter`'s functionality was a subset of `SearchTaskFilter`'s functionality. Keeping only `SearchTaskFilter` under a new name also allowed further simplifications in the `TaskPane` that constantly switched between different filters.

### DIFF
--- a/GTG/core/filters.py
+++ b/GTG/core/filters.py
@@ -61,60 +61,16 @@ class TagEmptyFilter(Gtk.Filter):
             return True
 
 
-class TaskPaneFilter(Gtk.Filter):
-    __gtype_name__ = 'TaskPaneFilter'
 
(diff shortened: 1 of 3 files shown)


diegogangl:
Brillant refactoring @gycsaba96! This simplies things a lot.
Merging, thanks :)

PR #1211, opened 2025-05-27
Disable edit button when multiple tags are selected

### LINKED ISSUE #1011: Ability to batch-delete the multiple selected tags in the sidebar
Since #123 resolved as part of the GTK4 port, it is possible to select multiple tags in the sidebar (very nice!), but _not_ to delete them in a batch.

It would be nice if we could do that, especially for cleaning the gazillion duplicate tags with "0 tasks" due to GTG 0.6's infamous #605.

### LINKED ISSUE #1012: When selecting multiple tags in the sidebar, the "Edit" action should be marked as insensitive
Now that multiple selections have been implemented in #123: 

When selecting multiple tags in the sidebar, the "Edit" action should probably be marked as insensitive, as I don't think we'd have a meaningful UI workflow (or usecase) for editing the properties of multiple tags at once

Related to #1011

diegogangl:
Fixes #1012 
Fixes #1011

### DIFF
--- a/GTG/gtk/browser/sidebar.py
+++ b/GTG/gtk/browser/sidebar.py
@@ -226,7 +226,7 @@ class Sidebar(Gtk.ScrolledWindow):
     def on_tag_RMB_click(self, gesture, sequence) -> None:
         """Callback when right-clicking on a tag."""
 
-        menu = TagContextMenu(self.ds, self.app, gesture.get_widget().tag)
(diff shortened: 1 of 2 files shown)


diegogangl:
@gycsaba96 Thanks for the review! I've moved the checks to the tag menu constructor and started passing the list of tags there.
We also got #1011 fixed for free

gycsaba96 (approved):

Issue #1206, opened 2025-04-25
Only the middle section of the task-row is colored with the tag color


gycsaba96:
In the task list, the tag color is applied only to the middle section of the given row. 
This behavior first appears in 6fcf137.

![Image](https://github.com/user-attachments/assets/9325d897-055c-4d62-a73f-460a835b1ce6)

PR #1212, opened 2025-05-28
Ensure that colouring fills the entire row in the task list

gycsaba96:
The solution mixes the old styling with the new `.rich-list` formatting while ensuring that spacing values are increments of 6 pixels.

This fixes GitHub issue #1206.

### DIFF
--- a/GTG/gtk/data/style.css
+++ b/GTG/gtk/data/style.css
@@ -24,6 +24,16 @@ indent, expander{
     -gtk-icon-size: 16px;
 }
 
+.task-list row {
+    padding: 0px;
+}
+
+.task-list .task-box {
+    min-height: 48px;
+    padding-left: 12px;
+    padding-right: 12px;
+}
+
 /* --------------------------------------------------------------------------------
  * ERROR HANDLER
  * --------------------------------------------------------------------------------
--- a/GTG/gtk/browser/task_pane.py
+++ b/GTG/gtk/browser/task_pane.py
@@ -39,6 +39,8 @@ class TaskBox(Gtk.Box):
         self.config = config
         super().__init__(valign=Gtk.Align.CENTER)
 
+        self.add_css_class('task-box')
+
         self.expander = Gtk.TreeExpander()
         self.expander.add_css_class('arrow-only-expander')
         self.expander.set_indent_for_icon(True)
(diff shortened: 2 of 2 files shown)


nekohayo (approved):
Tested in practice, I can confirm that the fix works as advertised!

GTG/gtk/data/style.css
```diff
@@ -24,6 +24,16 @@ indent, expander{
...
 
+.task-list row {
+    padding: 0px;
+}
+
+.task-list .task-box {
+    min-height: 48px;
+    padding-left: 12px;
+    padding-right: 12px;
+}
```
TheEvilSkeleton, line 35: Would it be possible to apply the color to its parent widget (GtkListItemWidget) instead? I think that approach would be safer than restyling and "fighting" with other CSS classes, which should also prevent us from adding more classes.

TheEvilSkeleton (changes requested):

GTG/gtk/data/style.css
gycsaba96: Unfortunately, coloring is not the only issue. Since drag-and-drop works on `TaskBox` instances, the draggable and droppable areas are also very thin without adjusting the `TaskBox` sizes.

drag-and-drop

gycsaba96: I did a bit more digging to investigate @TheEvilSkeleton's suggestion. The solution would require manipulating the `ListItemWidget` instead of the `TaskBox` from Python. However, the `ListItemWidget` is not visible to us when setting up drag-and-drop callbacks, we only see a `ListItem`. Moreover, I can not find `ListItemWidget` in the [docs](https://docs.gtk.org/gtk4/), suggesting that it is not part of the public GTK4 API and we shouldn't rely on it.

That said, I did try manipulating it, and I got various exceptions, errors, and segmentation faults...

While I understand that we want to avoid overriding defaults and built-in styles, I propose we stick with the CSS-based solution here.

TheEvilSkeleton: > While I understand that we want to avoid overriding defaults and built-in styles, I propose we stick with the CSS-based solution here.

Thanks for investigating this. I agree the CSS approach is better in this case.

I didn't expect it to require this much effort, but I will admit, I'm not familiar with GtkListView and its sister widgets. I get the feeling that GtkListView might be insufficient for our use case. I think GtkListBox and GtkListBoxRow are more suitable for our use case, because they're designed for more advanced widgetry – and GtkListBoxRow can be subclassed, which should allow us to have all the functionality we need on the row itself.

TheEvilSkeleton: @gycsaba96 age you open to port to GtkListBox in the future? No pressure of course. I haven't had the chance to test your MR yet, but we can merge this first before going substantial changes.

gycsaba96: @TheEvilSkeleton Thank you for your additional input. I mark this conversation as resolved then.

In principle, I would be happy to circle back to it. However, I am not sure how much effort and disruption is required to get rid of that single line of competing CSS code. Consider opening an issue so that others can also provide their insights or even do the work.

PR #1217, opened 2025-06-08
Improve the refiltering of all tasks

gycsaba96:
Switching from all tasks to a tag [can take an extremely long time with 5000+ tasks](https://github.com/getting-things-gnome/gtg/pull/1215#issuecomment-2952771547). This happens because manipulating the underlying `Gio.ListSotore` objects triggers many bind and unbind operations in the UI. Clearing the models before checking the items can massively reduce this time (from 295.3s to 0.8s in my case).


diegogangl:
LGTM, merging!

PR #1219, opened 2025-06-21
Speed up (and fix) TagCompletion initialisation

gycsaba96:
This is a follow-up of [my recent profiling](https://github.com/getting-things-gnome/gtg/pull/1215#issuecomment-2952771547) showing increased runtime in the `TagCompletion` class. While working, I also noticed a small bug in the code. 

- Fill up `TagCompletion.tags` with presorted values. (This reduced the runtime of the `Gtk.ListStore` population from 313.30s to 0.16s on my dataset containing 3621 tags.)
- Take into account that `Tag.name` does not have a leading `@` sign.

### DIFF
--- a/GTG/gtk/tag_completion.py
+++ b/GTG/gtk/tag_completion.py
@@ -63,6 +63,7 @@ class TagCompletion(Gtk.EntryCompletion):
         """ Initialize entry completion"""
         super().__init__()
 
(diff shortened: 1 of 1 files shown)


diegogangl:
That's a massive speed-up, thanks!
Everything looks good, merging

PR #1221, opened 2025-06-24
Fix automatic unit tests

gycsaba96:
The automatic GHA tests are failing because it does not find the *vobject* package. PR #1219 passed when I opened it, but the exact same code failed after the merge. The difference seems to be that the GHA moved to a *cladav* version not requiring *vobject* anymore.

**The changes are as follows:**
- Require the same *caldav* version as in the flatpak build.
- Add *vobject* as an explicit dependency.


diegogangl:
Thanks, merging!

PR #1224, opened 2025-07-29
Update task counters of tags when a task is deleted

gycsaba96:
The `DataStore` observes the 'removed' events of the `TaskStore` and updates the statistics of the tags.

2. Remove all tasks at the same time.
3. Observe that the counter in the sidebar are unchanged.


diegogangl:
LGTM and works as expected. Thanks!

Issue #1111, opened 2024-06-06
description preview shows &amp; in place of &


HeladoDeBrownie:
on version 0.6.0, when "Show description preview" is toggled on and a task's description has `&` written in it, its description preview shows `&amp;` instead

Issue #1013, opened 2024-02-27
"Compact" vs "Relaxed / Touch" density modes preferences for the Task Browser


nekohayo:
The new GTK4 UI is very nice, but it makes it less space-efficient to deal with my thousands of tasks (hundreds per tag, typically).

I would like the ability (through the Preference dialog) to set select between two density modes: the default touch-friendly one, and a "Compact" density mode where we would have only 3px padding above/below individual task item rows in the task browser.

There might also be a companion setting in the preferences dialog to auto-enable/disable compact vs cozy density in automatic light vs dark mode, because high-density textual contents are often harder to read in dark mode.

PR #1225, opened 2025-08-01
Show task previews based on the configuration

### LINKED ISSUE #1014: "Show description preview" Task Browser preference/setting broken
Since the GTK4 port, the "Show description preview" preference/setting doesn't really work anymore, as there is no gray description text showing next to the title text in the task browser window, only the tooltip on hovering the task...

gycsaba96:
Closes #1014 and #1111.

- Make `SectionConfig` a `GObject` with a bindable property. This can be reused for other similar tasks, e.g., #1013. (Maybe separate properties or a signal will be more suitable in the long run, but the current solution can be easily changed in any direction.)
- Add a new `GtkLabel` (holding the preview) to `TaskBox`.
- Bind the relevant properties of the `TaskBox` for automatic updates.


diegogangl:
Great work! We can iterate on configs in the future, the whole dialog could use some love.
Thanks :)

PR #1226, opened 2025-08-04
Fix quick add with a tag

gycsaba96:
I discovered two somewhat related bugs during testing.

**Steps to reproduce the first bug:**
2. Delete all tasks.
3. Select a tag.
4. Add a task via quick add and observe a crash.

This happens because GTG tries to select the last task in the task list even when it is empty. The original intention was probably to select the newly added task. (Note that it is still possible to have an empty list, e.g., when in the Closed or Actionable view.) I removed the `select_last` method since it wasn't used anywhere else.

**Fixing the crash revealed the second bug:** the newly added task is not visible in the task list. 

This is solved by the new `TaskStore.add_tags` method that emits a signal when the tags are added. It accepts a list of tags but emits only a single signal to keep things efficient.

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -886,6 +886,12 @@ class TaskStore(BaseStore[Task]):
         return f'Task Store. Holds {len(self.lookup)} task(s)'
 
 
+    def add_tags(self, task: Task, tags: list[Tag]):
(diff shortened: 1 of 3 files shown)


diegogangl:
Confirmed, thanks! Changes LGTM

PR #1229, opened 2025-08-15
Fix sidebar counters

### LINKED ISSUE #1023: Sidebar tags' dynamic task counters are sometimes buggy when switching between view modes
#170 was was implemented in 6127e1ce, dd4d1407, f336e44 and d0245fc20.

However, it is sometimes buggy. Testing with my 2021-11 anonymized dataset, I'm not entirely sure how to reproduce this in the clearest way (anyone has exact steps with a simpler dataset?).

What I observed there is that after going to the Actionable view within a task, and marking a bunch of tasks as done, task count does not match reality:

image

As shown in the screenshot above, there sure aren't 40 tasks shown in that current view combination (the counts for the Open and Closed views are different, too)

gycsaba96:
Closes #1023 

While considering different options, I realised that if a task is changed in a way that it influences at least one counter, it can possibly impact any other counters as well. The complication mainly revolves around actionability. E.g., if the user completes/removes/dismisses a task, it can impact: all the tags applied to that task, their ancestor tags, and the same can be true for the parent task, which can possibly become actionable.

**Based on the previous observation, I made the following changes:**
- Move stat-related calculations to its own class (`TagStats`). The goal is to prevent logic from being scattered and making other methods more complex. This could also help recognise bugs and optimise the calculations.
- Connected major task events with stat recalculations. This may raise performance concerns for large datasets. I plan to investigate it, but this PR is already big enough. `TagStats` is a good start for incremental updates, but in the worst case, we can consider enabling and disabling stat calculations with a keybinding.
- Use IDs instead of tag names. This avoids conflicting names when _"all"_ or _"untagged"_ names are used, and avoids complications when tags are renamed.

### DIFF
--- a/GTG/core/datastore.py
+++ b/GTG/core/datastore.py
@@ -54,6 +54,11 @@ class TaskCounts(GObject.Object):
     task_count_actionable = GObject.Property(type=int,default=0)
     task_count_closed = GObject.Property(type=int,default=0)
 
+    def reset(self) -> None:
+        self.task_count_open = 0
+        self.task_count_actionable = 0
+        self.task_count_closed = 0
+
 
 
 class TagStats:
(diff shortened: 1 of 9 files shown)


gycsaba96:
Good catch, thank you, @diegogangl 
 
A change in start date didn’t trigger a possible change in actionability.

diegogangl:
All set then, merging...
Thanks!

PR #1232, opened 2025-08-22
 Remove `flatpak_snapshot` profile

TheEvilSkeleton:


diegogangl:
Sounds good, thanks

SqAtx:
What was wrong with it? I'd added it in https://github.com/getting-things-gnome/gtg/commit/a34e8db2690878e8e9fbed9ab7ce94b82c34f2ac so that when using the weekly flatpak, I get a sense of how long ago it was released. Probably should have been called `weekly_snapshot`. Why remove it?

TheEvilSkeleton:
I removed it because, to be honest, we shouldn't give "special" favor for Flatpak just because we officially support it. It's best to use the development version for development purposes, whether it's within Flatpak or another environment. We can still change the development build to include the date it was built, if you prefer that

PR #1234, opened 2025-08-30
Use GLib.idle_add to schedule the recalculation of sidebar counters

gycsaba96:
PR #1233 did a good job speeding up the program startup. However, there are other scenarios that can trigger frequent recalculations (e.g., deleting many tasks, syncing backends, etc.), which can be slow with a large dataset. Constantly disabling and enabling recalculations would pollute the source code, so I decided to look for a different solution.

Luckily, [GLib.idle_add](https://docs.gtk.org/glib/func.idle_add.html) can schedule the recalculation to run after all of these changes have been made. Since these values are only used in the UI to inform the user, we shouldn't worry about any other functionality being affected.

### DIFF
--- a/GTG/core/datastore.py
+++ b/GTG/core/datastore.py
@@ -38,7 +38,7 @@ import GTG.core.info as info
 import GTG.core.dirs as dirs
 import GTG.core.versioning as versioning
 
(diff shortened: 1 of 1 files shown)


diegogangl:
Looks good, thanks!

PR #1238, opened 2025-09-26
Implement a preference for compact mode

### LINKED ISSUE #1013: "Compact" vs "Relaxed / Touch" density modes preferences for the Task Browser

gycsaba96:
Fixes #1013

- extend the default config,
- extend the preferences window with the new option,
- dynamically add and remove a CSS class to the `TaskBox`.

### DIFF
--- a/GTG/core/config.py
+++ b/GTG/core/config.py
@@ -34,6 +34,7 @@ log = logging.getLogger(__name__)
 DEFAULTS = {
     'browser': {
         "bg_color_enable": True,
+        "compact_mode": False,
         "contents_preview_enable": False,
         'tag_pane': True,
         "sidebar_width": 265,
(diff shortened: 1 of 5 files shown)


diegogangl:
This looks great, thanks! Merging right away

PR #1242, opened 2025-11-06
Catch potential errors in base_store.parent

diegogangl:
This can happen with very old datasets (like Bryce)


gycsaba96 (changes requested):
I was unable to load the Bryce dataset and reproduce the crash. It was probably caused by a task that had multiple parents (e.g., task 67@1).

Nevertheless, we can be extra cautious, but I think this deserves an error-level log message.

PR #1258, opened 2026-02-16
Apply sidebar filters to all panes

### LINKED ISSUE #1246: "Tasks with no tags" sidebar filter is not applied when switching between task browser's "Open" / "Actionable" / "Closed" views tabs
Tested with the current git `master` version, I noticed that if you select the "Tasks with no tabs" filter in the sidebar, it will initially correctly show only tasks without tags, but if you switch between one of the view mode tabs at the top (Open/Actionable/Closed) it will reset to showing all tasks, including those with tags.

Demonstration video:

https://github.com/user-attachments/assets/b54044a6-99e0-4342-9660-4152850176f3

gycsaba96:
Fixes #1246

The sidebar updated only the current pane instead of all of them. (The same bug was reproducible for saved searches, too.)

**The changes are as follows:**
- Introduce helper methods in `MainWindow` to change the filtering in all panes at once.
- Use these methods in `Sidebar` instead of updating only the current pane.
- Do not call `TaskPane.set_filter_tags` when changing panes. (This would undo the "untagged" filtering.)

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -1401,7 +1401,6 @@ class MainWindow(Gtk.ApplicationWindow):
         #       to actionable
         self.stack_switcher.get_stack().get_first_child().get_first_child().emit('expand-all')
 
-        self.get_pane().set_filter_tags(set(self.sidebar.selected_tags()))
         self.sidebar.change_pane(current_pane)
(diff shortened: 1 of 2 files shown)


diegogangl:
LGTM, thanks! (sorry for the massive delay)

PR #1260, opened 2026-02-21
fix(caldav): add default calendar fallback for tasks without calendar tag

pentux-GitHub:
## Problem
When a task has no calendar tag (@DAV_name) and no calendar_name 
attribute, GTG silently skips synchronization with the message 
"has no calendar to be synced with".

This affects all new users who set up CalDAV sync without knowing 
they need to manually tag every task with @DAV_<calendar_name>.

## Solution
Add a fallback in _get_todo_and_calendar() that uses a calendar 
named 'gtg' by default, or the first available calendar if none 
is named 'gtg'.

## Update
Added a configurable `default-calendar` parameter so users can set 
their preferred default calendar name, instead of relying on a 
hardcoded value. Defaults to 'gtg' for backward compatibility.

## Tested on
- GTG 0.6.0 / Debian 13
- Nextcloud CalDAV
- Synced with Tasks.org on Fairphone 4 (e/OS)

### DIFF
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -83,6 +83,9 @@ class Backend(PeriodicImportBackend):
         "is-first-run": {
             GenericBackend.PARAM_TYPE: GenericBackend.TYPE_BOOL,
             GenericBackend.PARAM_DEFAULT_VALUE: True},
+        "default-calendar": {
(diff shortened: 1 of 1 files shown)


diegogangl:
LGTM and sounds like a good solution but it looks like this isn't exposed to the UI?
(also, sorry for the delay!)

diegogangl:
Merging, just having a default calendar name is an improvement

PR #1262, opened 2026-02-25
fix: don't apply search filter when switching panes if searchbar is closed (#1244)

### LINKED ISSUE #1244: Dismissed search (invisible searchbar) silently filters the view when switching between Open / Actionable / Closed tabs
Somewhat related to #1243: in the current git `master` version:

1. `Ctrl+F`
2. Search for something that ends up filtering the view
3. Press `Esc` to dismiss the search; the view becomes unfiltered again
4. Switch to another view mode tab (ex: from "Open" to "Actionable")

Result: the view is silently filtered by the search term even though the search mode is disabled. Doing `Ctrl+F` again reveals what's going on behind the scenes, and pressing `Esc` to dismiss that search view unfilters that tab's view again.

Demonstration video:

https://github.com/user-attachments/assets/5bb1ac9d-b6c8-4cfe-b68e-8da0f6ed97e0

pentux-GitHub:
## Summary

Fixes #1244

When the search was dismissed (via `Esc` or the clear button) and the 
user then switched to another view tab (Open / Actionable / Closed), 
the new tab's view was silently filtered by the previous search term, 
even though the searchbar was visually closed.

## Steps to reproduce (before this fix)

1. `Ctrl+F` to open the search bar
2. Type something that filters the task list
3. Press `Esc` to dismiss the search — the current view becomes unfiltered
4. Switch to another tab (e.g. from "Open" to "Actionable")
5. **Bug**: the new tab is silently filtered by the previous search term
6. Pressing `Ctrl+F` again reveals the hidden search term still active

A second variant also existed: clearing the search field via the X 
button (without pressing `Esc`) and then switching tabs produced the 
same silent filtering.

## Root cause

In `on_pane_switch()`, the search query was applied to the new pane 
unconditionally:
```python
if search_query := self.search_entry.get_text():
    self.get_pane().set_search_query(search_query)
```

The text entry retains its content even after the searchbar is 
dismissed. So when switching panes, the stale query was silently 
re-applied to the new pane.

## Fix

One line change in `on_pane_switch()`: only apply the search query 
when the searchbar is actually active (`get_search_mode()` returns `True`):
```python
if self.searchbar.get_search_mode() and (search_query := self.search_entry.get_text()):
    self.get_pane().set_search_query(search_query)
```

## Testing

Verified manually on the development version (`./launch.sh`) by 
reproducing both variants described above. After this fix:
- Switching tabs after dismissing search no longer silently filters the view
- Switching tabs while search is active still correctly applies the query
- Re-opening the searchbar (loupe button or `Ctrl+F`) still correctly 
  recalls the previous term and re-filters — this is expected behavior
- No regressions observed on normal search and tab switching workflows

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -1406,7 +1406,7 @@ class MainWindow(Gtk.ApplicationWindow):
         self.get_pane().sort_btn.set_popover(None)
         self.get_pane().sort_btn.set_popover(self.sort_menu)
 
-        if search_query := self.search_entry.get_text():
+        if self.searchbar.get_search_mode() and (search_query := self.search_entry.get_text()):
             self.get_pane().set_search_query(search_query)
 
         self.notify('is_pane_open')
--- a/GTG/gtk/editor/editor.py
+++ b/GTG/gtk/editor/editor.py
@@ -169,6 +169,11 @@ class TaskEditor(Gtk.Window):
 
         self.textview.set_text_from_task()
 
+        # Refresh editor if a subtask is deleted externally (fix #544)
+        self._removed_handler = self.ds.tasks.connect(
+            'removed', self._on_task_removed
+        )
+
         # Connect search field to tags popup
         self.tags_tree.set_search_entry(self.tags_entry)
         self.tags_tree.set_search_equal_func(self.search_function, None)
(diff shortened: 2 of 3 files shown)


diegogangl:
Looks good, thanks!

PR #1264, opened 2026-02-25
fix: replace deprecated task.get_id() with task.id in generic_backend

pentux-GitHub:
## Summary

In the new core, `Task` IDs are UUID objects accessed directly via 
`task.id`. The old `task.get_id()` method no longer exists, causing 
an `AttributeError` when the CalDAV backend tries to sync tasks.

## Error
```
AttributeError: 'UUID' object has no attribute 'get_id'
```

## Changes

Two occurrences fixed in `GTG/backends/generic_backend.py`:

- Line 660: in `_do_set_task_thread()` 
- Line 680: in `queue_set_task()`
```python
# Before
tid = task.get_id()

# After
tid = task.id
```

## Testing

Verified with a live CalDAV backend — the `AttributeError` no longer 
appears in the terminal after this fix.

### DIFF
--- a/GTG/backends/generic_backend.py
+++ b/GTG/backends/generic_backend.py
@@ -657,7 +657,7 @@ class GenericBackend():
                 task = self.to_set.pop()
             except IndexError:
                 break
-            tid = task.get_id()
+            tid = task.id
             if tid not in self.to_remove:
                 self.set_task(task)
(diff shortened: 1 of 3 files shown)


diegogangl:
LGTM, thanks!

Issue #1176, opened 2025-01-20
_get_calendar_tasks() references undefined 'uid' variable


SqAtx:
```
    def _get_calendar_tasks(self, calendar: iCalendar):
        """Getting all tasks that has the calendar tag"""
        for task in self.datastore.tasks.data:
            if CATEGORIES.has_calendar_tag(task, calendar):
                yield uid, task
```
https://github.com/getting-things-gnome/gtg/blob/3910d84e87040b40cf2926fc8f143f324810ee94/GTG/backends/backend_caldav.py#L344-L348

Found it running Ruff.

I'm not sure how this translates into an actual bug, because I don't use that backend. I don't think it has been reported before.

The backend was added in https://github.com/getting-things-gnome/gtg/commit/64d535aee2c59e78fdc269ddd4baaecf257b64b4, and this was broken in https://github.com/getting-things-gnome/gtg/commit/74bd3f447b0e198f6420ddde378e4d5ae7177785

pentux-GitHub:
This was fixed by #1265 (merged). `_get_calendar_tasks` no longer references the undefined `uid`:

```python
def _get_calendar_tasks(self, calendar: iCalendar):
    """Getting all tasks that has the calendar tag"""
    for task in self.datastore.tasks.data:
        if CATEGORIES.has_calendar_tag(task, calendar):
            yield str(task.id), task
```

The `# noqa: F821` that was suppressing the Ruff warning is gone too. Safe to close.

PR #1241, opened 2025-11-06
Add missing infobar for backends

diegogangl:
fixes #998

⚠️ This needs help testing!

PR #1240, opened 2025-11-06
Ensure we have an added date to task in caldav

diegogangl:
Syncing a task that has no CREATED field leads to a state where the task's 'date_added' attribute is not set.

The xml task element that is subsequently written to disk has then no added tag, although the schema requires it to be always present.

fixes #1033

⚠️ This needs help testing!

PR #1265, opened 2026-02-25
fix(caldav): restore CalDAV backend compatibility with the new Task core API

pentux-GitHub:
## Summary

The CalDAV backend was written against the old GTG core API, where `Task` 
exposed a set of getter/setter methods (`get_title()`, `set_due_date()`, 
`get_uuid()`, etc.). Following the core rewrite, `Task` now exposes these 
values as Python properties (`task.title`, `task.date_due`, `task.id`, etc.).

This PR restores full compatibility between the CalDAV backend and the new 
core, fixing a cascade of errors that prevented the backend from syncing at 
all.

---

## Errors fixed

**`backend_caldav.py`**

| Error | Root cause | Fix |
|---|---|---|
| `TypeError: 'str' object is not callable` | `Field.get_gtg()` and `write_gtg()` called old getter/setter methods that no longer exist as callables | Added `PROPERTY_GET_MAP` and `PROPERTY_SET_MAP` dictionaries to bridge old method names to new property access |
| `KeyError` on task lookup | `tasks.lookup[uid]` raises if key is absent | Changed to `tasks.lookup.get(UUID(uid))` |
| `TypeError: Task.__init__() missing arguments` | `Task()` now requires `id` and `title` | Changed to `Task(id=UUID(uid), title='')` |
| `AttributeError: 'set' object has no attribute 'sort'` | `task.tags` is now a `set`, not a list | Removed `task.tags.sort()`, which has no meaning on a set |
| `NameError: name 'uid' is not defined` in `_get_calendar_tasks` | `uid` was referenced but never defined in this scope (broken since commit 74bd3f44, see #1176) | Changed to `str(task.id)` |
| `KeyError` in `set_task` | `datastore.tasks.add()` raises if task already exists | Added a guard: only call `add()` if task is not already in the store |

**`generic_backend.py`**

| Error | Root cause | Fix |
|---|---|---|
| `AttributeError: 'UUID' object has no attribute 'get_id'` | `datastore` now passes `task.id` (a UUID) to `queue_set_task()`, but the method tried to call `.get_id()` on it | Refactored `queue_set_task()` and `launch_setting_thread()` to work with UUID directly, resolving the Task object from the store before calling `set_task()` |

---

## Behavior after this fix

- CalDAV backend connects and syncs successfully ✅
- Tasks from the remote CalDAV server are imported into GTG ✅  
- Terminal is clean — no exceptions during normal sync operation ✅

## Known remaining issue

Tasks created in GTG are not yet pushed back to the CalDAV server. 
This appears to be a separate issue in the outbound sync path and is 
not introduced by this PR.

## Testing

Tested against a live Nextcloud CalDAV server. Tasks created remotely 
are correctly imported into GTG on the next sync cycle.

### DIFF
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -23,7 +23,7 @@ Backend for storing/loading tasks in CalDAV Tasks
 import logging
 import re
 from collections import defaultdict
-from uuid import UUID
+from uuid import NAMESPACE_URL, UUID, uuid5
 from datetime import date, datetime
 from gettext import gettext as _
 from hashlib import md5
@@ -50,6 +50,23 @@ DAV_IGNORE = {'last-modified',  # often updated alone by GTG
               'percent-complete',  # calculated on subtask and status
               'completed',  # GTG date is constrained
               }
+
+
+# CalDAV UIDs are opaque unique strings (RFC 5545), while GTG task ids
+# are UUIDs. Valid UUID strings are used as-is; anything else is mapped
+# to a stable name-based UUID (RFC 4122, uuid5) so that the same remote
+# UID always yields the same task id across sync runs.
+UUID_NAMESPACE = uuid5(NAMESPACE_URL, 'gtg://backend_caldav')
+
+
+def uid_to_task_id(uid) -> UUID:
+    """Map a CalDAV UID (any unique string) to a stable GTG task UUID."""
+    if isinstance(uid, UUID):
+        return uid
+    try:
+        return UUID(uid)
+    except (TypeError, AttributeError, ValueError):
+        return uuid5(UUID_NAMESPACE, str(uid))
 
 
 class Backend(PeriodicImportBackend):
@@ -205,11 +222,11 @@ class Backend(PeriodicImportBackend):
                              '%r => %r', task, new_todo)
             return
         uid = UID_FIELD.get_dav(todo=new_todo)
-        self._cache.set_todo(new_todo, uid)
+        self._cache.set_todo(new_todo, str(uid_to_task_id(uid)))
 
     def _remove_todo(self, uid: str, todo: iCalendar) -> None:
         logger.info('SYNCING removing todo for Task(%s)', uid)
-        self._cache.del_todo(uid)  # cleaning cache
+        self._cache.del_todo(str(uid_to_task_id(uid)))  # cleaning cache
         try:  # deleting through caldav
             todo.delete()
         except caldav.lib.error.DAVError:
@@ -288,7 +305,9 @@ class Backend(PeriodicImportBackend):
     def _import_calendar_todos(self, calendar: iCalendar,
                                import_started_on: datetime, counts: dict):
         todos = calendar.todos(include_completed=not self._cache.initialized)
-        todo_uids = {UID_FIELD.get_dav(todo) for todo in todos}
+        todo_uids = {str(uid_to_task_id(uid))
+                     for uid in (UID_FIELD.get_dav(todo) for todo in todos)
+                     if uid}
 
         # browsing all task linked to current calendar,
         # removing missed ones we don't see in fetched todos
@@ -301,11 +320,15 @@ class Backend(PeriodicImportBackend):
 
         for todo in self.__sort_todos(todos):
             uid = UID_FIELD.get_dav(todo)
-            self._cache.set_todo(todo, uid)
+            if not uid:  # RFC 5545 requires a UID, but stay safe
+                logger.warning('Skipping todo without UID: %r', todo)
+                continue
+            tid = uid_to_task_id(uid)
+            self._cache.set_todo(todo, str(tid))
             # Updating and creating task according to todos
-            task = self.datastore.tasks.lookup.get(UUID(uid))
+            task = self.datastore.tasks.lookup.get(tid)
             if not task:  # not found, creating it
-                task = Task(id=UUID(uid), title='')
+                task = Task(id=tid, title='')
                 Translator.fill_task(todo, task, self.namespace, self.datastore)
                 self.datastore.tasks.add(task)
                 counts['created'] += 1
@@ -339,7 +362,8 @@ class Backend(PeriodicImportBackend):
                 parents = PARENT_FIELD.get_dav(todo)
                 if (not parents  # no parent mean no relationship on build
                         or parents[0] in known_todos  # already known parent
-                        or self.datastore.tasks.lookup.get(UUID(uid))):  # already known uid
+                        or self.datastore.tasks.lookup.get(
+                            uid_to_task_id(uid))):  # already known uid
                     yield todo
                     known_todos.add(uid)
             if loop >= MAX_CALENDAR_DEPTH:
@@ -815,7 +839,9 @@ class Description(Field):
                 if new_line:
                     result += new_line + '\n'
             elif line.startswith('{!') and line.endswith('!}'):
-                subtask = task.req.get_task(line[2:-2].strip())
+                tid = line[2:-2].strip()
+                subtask = next((child for child in task.children
+                                if str(child.id) == tid), None)
                 if not subtask:
                     continue
                 if subtask.status == TaskStatus.DONE:
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -23,6 +23,7 @@ Backend for storing/loading tasks in CalDAV Tasks
 import logging
 import re
 from collections import defaultdict
+from uuid import UUID
 from datetime import date, datetime
 from gettext import gettext as _
 from hashlib import md5
@@ -119,7 +120,9 @@ class Backend(PeriodicImportBackend):
             logger.warning("not loaded yet, ignoring set_task")
             return
         with self.datastore.mutex:
-            self.datastore.tasks.add(task)
+            if task.id not in self.datastore.tasks.lookup:
+                self.datastore.tasks.add(task)
+            # If task already exists, it was updated in-place by the backend
 
     @interruptible
     def remove_task(self, tid: str) -> None:
@@ -300,10 +303,9 @@ class Backend(PeriodicImportBackend):
             uid = UID_FIELD.get_dav(todo)
             self._cache.set_todo(todo, uid)
             # Updating and creating task according to todos
-            task = self.datastore.tasks.lookup[uid]
+            task = self.datastore.tasks.lookup.get(UUID(uid))
             if not task:  # not found, creating it
-                task = Task()
-                task.id = uid
+                task = Task(id=UUID(uid), title='')
                 Translator.fill_task(todo, task, self.namespace, self.datastore)
                 self.datastore.tasks.add(task)
                 counts['created'] += 1
@@ -337,7 +339,7 @@ class Backend(PeriodicImportBackend):
                 parents = PARENT_FIELD.get_dav(todo)
                 if (not parents  # no parent mean no relationship on build
                         or parents[0] in known_todos  # already known parent
-                        or self.datastore.tasks.lookup[uid]):  # already known uid
+                        or self.datastore.tasks.lookup.get(UUID(uid))):  # already known uid
                     yield todo
                     known_todos.add(uid)
             if loop >= MAX_CALENDAR_DEPTH:
@@ -348,8 +350,7 @@ class Backend(PeriodicImportBackend):
         """Getting all tasks that has the calendar tag"""
         for task in self.datastore.tasks.data:
             if CATEGORIES.has_calendar_tag(task, calendar):
-                # This line was broken in commit 74bd3f44 - see #1176
-                yield uid, task # noqa: F821
+                yield str(task.id), task
 
     #
     # Utility methods
@@ -407,9 +408,48 @@ class Field:
     def _is_value_allowed(self, value):
         return value not in self.ignored_values
 
+    # Mapping: old method names -> Task property names (new core)
+    PROPERTY_GET_MAP = {
+        'get_title':               'title',
+        'get_excerpt':             'content',
+        'get_uuid':                'id',
+        'get_start_date':          'date_start',
+        'get_due_date_constraint': 'date_due',
+        'get_closed_date':         'date_closed',
+        'get_added_date':          'date_added',
+        'get_modified':            'date_modified',
+    }
+
+    PROPERTY_SET_MAP = {
+        'set_uuid':        None,
+        'set_text':        'content',
+        'set_start_date':  'date_start',
+        'set_due_date':    'date_due',
+        'set_closed_date': 'date_closed',
+        'set_added_date':  'date_added',
+        'set_modified':    'date_modified',
+        'set_tags':        None,
+        'set_parent':      None,
+        'remove_parent':   None,
+    }
+
     def get_gtg(self, task: Task, namespace: str = None):
         "Extract value from GTG.core.task.Task according to specified getter"
-        return getattr(task, self.task_get_func_name)()
+        name = self.task_get_func_name
+        if name in self.PROPERTY_GET_MAP:
+            value = getattr(task, self.PROPERTY_GET_MAP[name])
+            if name == 'get_uuid':
+                return str(value)
+            return value
+        if name == 'get_tags_name':
+            return [t.name for t in task.tags]
+        if name == 'get_parents':
+            return [str(task.parent.id)] if task.parent else []
+        if name == 'get_children':
+            return [str(c.id) for c in task.children]
+        # Fallback: call as method (e.g. get_status)
+        attr = getattr(task, name)
+        return attr() if callable(attr) else attr
 
     def clean_dav(self, vtodo: iCalendar):
         """Will remove existing conflicting value from vTodo object"""
@@ -440,7 +480,18 @@ class Field:
 
     def write_gtg(self, task: Task, value, namespace: str = None):
         """Will write new value to GTG.core.task.Task"""
-        return getattr(task, self.task_set_func_name)(value)
+        name = self.task_set_func_name
+        if name in self.PROPERTY_SET_MAP:
+            prop = self.PROPERTY_SET_MAP[name]
+            if prop is not None:
+                setattr(task, prop, value)
+            return
+        # Fallback: call as method (e.g. set_title, set_status)
+        attr = getattr(task, name)
+        if callable(attr):
+            return attr(value)
+        else:
+            setattr(task, name, value)
 
     def set_gtg(self, todo: iCalendar, task: Task,
                 namespace: str = None) -> None:
@@ -648,7 +699,7 @@ class Categories(Field):
             task.add_tag(to_add)
         for to_delete in local_tags.difference(remote_tags):
             task.remove_tag(to_delete)
-        task.tags.sort(key=remote_tags.index)
+        # task.tags is a set in the new core, ordering is not supported
 
     def get_calendar_tag(self, calendar: iCalendar) -> str:
         return self.to_tag(calendar.name, DAV_TAG_PREFIX)
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -35,7 +35,7 @@ from GTG.backends.generic_backend import GenericBackend
 from GTG.backends.periodic_import_backend import PeriodicImportBackend
 from GTG.core.dates import LOCAL_TIMEZONE, Accuracy, Date
 from GTG.core.interruptible import interruptible
-from GTG.core.tasks import Task, Status
+from GTG.core.tasks import Task, Status as TaskStatus
 from vobject import iCalendar
 
 logger = logging.getLogger(__name__)
@@ -122,7 +122,7 @@ class Backend(PeriodicImportBackend):
         with self.datastore.mutex:
             if task.id not in self.datastore.tasks.lookup:
                 self.datastore.tasks.add(task)
-            # If task already exists, it was updated in-place by the backend
+        self._set_task(task)
 
     @interruptible
     def remove_task(self, tid: str) -> None:
@@ -133,7 +133,7 @@ class Backend(PeriodicImportBackend):
             logger.warning("no task id passed to remove_task call, ignoring")
             return
         with self.datastore.mutex:
-            return self.datastore.tasks.remove(tid)
+            return self.datastore.tasks.remove(UUID(tid) if isinstance(tid, str) else tid)
 
     #
     # real main methods
@@ -240,7 +240,7 @@ class Backend(PeriodicImportBackend):
         fact that it's missing"""
         task, do_delete = None, False
         task = calendar_tasks[uid]
-        if import_started_on < task.get_added_date():
+        if import_started_on < task.date_added:
             return
         # if first run, we're getting all task, including completed
         # if we miss one, we delete it
@@ -248,7 +248,7 @@ class Backend(PeriodicImportBackend):
             do_delete = True
         # if cache is initialized, it's normal we missed completed
         # task, but we should have seen active ones
-        elif task.get_status() == Task.STA_ACTIVE:
+        elif task.status == TaskStatus.ACTIVE:
             __, calendar = self._get_todo_and_calendar(task)
             if not calendar:
                 logger.warning("Couldn't find calendar for %r", task)
@@ -264,7 +264,7 @@ class Backend(PeriodicImportBackend):
         if do_delete:  # the task was missing for a good reason
             counts['deleted'] += 1
             self._cache.del_todo(uid)
-            self.datastore.tasks.remove(uid)
+            self.datastore.tasks.remove(UUID(uid))
 
     @staticmethod
     def _denorm_children_on_vtodos(todos: list):
@@ -518,7 +518,7 @@ class Field:
     @classmethod
     def _browse_subtasks(cls, task: Task):
         yield task
-        for subtask in task.get_subtasks():
+        for subtask in task.children:
             yield from cls._browse_subtasks(subtask)
 
 
@@ -615,11 +615,11 @@ class UTCDateTimeField(DateField):
 
 
 class Status(Field):
-    DEFAULT_STATUS = (Status.ACTIVE, 'NEEDS-ACTION')
-    _status_mapping = ((Status.ACTIVE, 'NEEDS-ACTION'),
-                       (Status.ACTIVE, 'IN-PROCESS'),
-                       (Status.DISMISSED, 'CANCELLED'),
-                       (Status.DONE, 'COMPLETED'))
+    DEFAULT_STATUS = (TaskStatus.ACTIVE, 'NEEDS-ACTION')
+    _status_mapping = ((TaskStatus.ACTIVE, 'NEEDS-ACTION'),
+                       (TaskStatus.ACTIVE, 'IN-PROCESS'),
+                       (TaskStatus.DISMISSED, 'CANCELLED'),
+                       (TaskStatus.DONE, 'COMPLETED'))
 
     def _translate(self, gtg_value=None, dav_value=None):
         for gtg_status, dav_status in self._status_mapping:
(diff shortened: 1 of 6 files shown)


pentux-GitHub:
Update: after further testing, the fix also restores the GTG → CalDAV 
sync direction (tasks created in GTG are now pushed to the remote server).

Two additional fixes were needed beyond the initial commit:

1. `datastore.py`: connect `task-filterably-changed` signal to 
   `queue_set_task()` so backends are notified when a task is modified.
   Without this, changes in GTG were never pushed to the backend.

2. `backend_caldav.py`: replace remaining old-API calls 
   (`get_added_date()`, `get_subtasks()`, `Task.STA_ACTIVE`, 
   `tasks.remove(str)`) with new core equivalents.

Tested: tasks created in GTG with the correct CalDAV calendar tag 
(`@DAV_Personnel` in my case) are now correctly pushed to the 
Nextcloud CalDAV server.

diegogangl:
Testing this. I came across this error when sync-ing:

```
Exception in thread Thread-1 (__backend_startup):
  File "/usr/lib64/python3.14/threading.py", line 1082, in _bootstrap_inner
    self._context.run(self.run)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/usr/lib64/python3.14/threading.py", line 1024, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/datastore.py", line 646, in __backend_startup
    backend.start_get_tasks()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/interruptible.py", line 38, in new
    return fn(*args)
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/periodic_import_backend.py", line 79, in start_get_tasks
    self._start_get_tasks()
    ~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/periodic_import_backend.py", line 98, in _start_get_tasks
    self.do_periodic_import()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/interruptible.py", line 38, in new
    return fn(*args)
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/backend_caldav.py", line 112, in do_periodic_import
    self._do_periodic_import()
    ~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/backend_caldav.py", line 148, in _do_periodic_import
    self._import_calendar_todos(calendar, start, counts)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/backend_caldav.py", line 299, in _import_calendar_todos
    for todo in self.__sort_todos(todos):
                ~~~~~~~~~~~~~~~~~^^^^^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/backend_caldav.py", line 339, in __sort_todos
    or self.datastore.tasks.lookup.get(UUID(uid))):  # already known uid
                                       ~~~~^^^^^
  File "/usr/lib64/python3.14/uuid.py", line 219, in __init__
    raise ValueError('badly formed hexadecimal UUID string')
ValueError: badly formed hexadecimal UUID string

```

pentux-GitHub:
Thanks a lot for testing! That's a real gap in my fix: I assumed CalDAV UIDs
are always valid UUIDs. My Nextcloud generates UUID-shaped UIDs so I never
hit it, but the spec only requires a unique string, so `UUID(uid)` can crash
on other servers or clients like yours.

I'll rebase the branch (it now conflicts with #1264 which got merged), add a
proper guard for non-UUID UIDs, re-test against Nextcloud in both directions,
and ping you here.

pentux-GitHub:
Rebased on current master (the conflict was with my own #1264, which this
branch's refactor supersedes), and fixed the crash you found. Thanks again
for testing, your server caught what mine couldn't!

## Root cause of your ValueError

My fix assumed CalDAV UIDs are always valid UUIDs. My Nextcloud generates
UUID-shaped UIDs so I never hit it, but RFC 5545 only requires an opaque
unique string, which is what your server produces.

## Fix

New `uid_to_task_id()` helper: a valid UUID string is used as-is, anything
else is mapped to a stable name-based UUID (RFC 4122 uuid5), so the same
remote UID always yields the same task id across sync runs. It is applied
at every boundary between remote UIDs and the GTG task store: the import
loop, the `__sort_todos` lookup that crashed for you, the cache keys, and
the reconciliation set.

Canonicalizing the cache keys and the reconciliation set also fixes a
latent bug for valid-but-non-canonical UUIDs (e.g. uppercase, which
Nextcloud's own web UI produces): without it, such tasks were flagged
"missing from backend" on every cycle (delete + reimport churn) and pushes
could create duplicates. I caught one live during testing: an
uppercase-UID todo now updates in place.

Also removed the last old-core API call left in the file
(`task.req.get_task()` in `_extract_plain_text`), which crashed when
pushing any task whose content references subtasks (the tutorial task
does).

## Tests

Added a new, unskipped test class with 3 tests: importing a non-UUID UID
(reproduces your exact ValueError on master), mapping stability, and the
subtask-reference extraction. The old skipped suite still targets the
pre-rewrite core API; modernizing it is a separate follow-up I'd be happy
to tackle.

Re-tested against a live Nextcloud in both directions, including cross
modifications updating the same .ics file (no duplicates).

## Known limitation (for a follow-up)

For non-UUID-UID servers, `_clean_task_missing_from_backend` queries the
server by the GTG task id (`todo_by_uid`), which can't match the original
remote UID, so the graceful "fetch and update" recovery degrades to
delete + reimport on the next cycle. The proper fix is storing the remote
UID as a namespaced task attribute (the backend already does this for
calendar name/url); that felt out of scope here.

Could you give it another run against your server?

diegogangl:
It's moving along now, I'm getting this error (through the gui):


```python-traceback
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/gtk/browser/main_window.py", line 1725, in remove_backend_infobar
    self.vbox_toolbars.foreach(self.__remove_backend_infobar, backend_id)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Box' object has no attribute 'foreach'
```
and I'm seeing this in the command line:

```
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/tasks.py", line 807, in _model_expand
    self.tid_to_subtask_model[item.id] = self._create_model_for_children(item)
                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/tasks.py", line 815, in _create_model_for_children
    if self.task_filter.match(child):
       ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
TypeError: argument item: Expected GObject.Object, but got str
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/gtk/browser/task_pane.py", line 507, in <lambda>
    ,lambda _,x: not self.filter_manager.has_matching_children(item)),
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/tasks.py", line 709, in has_matching_children
    return any(self.task_filter.match(c) for c in task.children)
           ^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/tasks.py", line 709, in <genexpr>
    return any(self.task_filter.match(c) for c in task.children)
               ~~~~~~~~~~~~~~~~~~~~~~^^^
TypeError: argument item: Expected GObject.Object, but got str
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/filters.py", line 141, in do_match
    return (self.is_task_matched_by_pane(task)
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/filters.py", line 121, in is_task_matched_by_pane
    return task.is_actionable
           ^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.14/site-packages/gi/_propertyhelper.py", line 246, in __get__
    value = self.fget(instance)
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/tasks.py", line 123, in is_actionable
    active_children = all(t.status != Status.ACTIVE for t in self.children)
                      ^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/tasks.py", line 123, in <genexpr>
    active_children = all(t.status != Status.ACTIVE for t in self.children)
                          ^^^^^^^^
AttributeError: 'str' object has no attribute 'status'
```

pentux-GitHub:
Second layer peeled! Reproduced your `str`-in-`task.children` with the
project's own RELATED-TO fixtures (`UID:ROOT` / `UID:CHILD`): the generic
field pipeline was feeding raw UID strings to `Task.add_child()`, which
appends anything without a type check. The same pattern sat one layer
deeper for CATEGORIES: `task.add_tag()` was given plain strings, which the
new core rightfully refuses, so any todo with categories crashed the
import too.

Fix, in the last commit:
- `fill_task` no longer routes CATEGORIES / RELATED-TO through the generic
  string pipeline; categories become real Tag objects via
  `datastore.tags.new()`, keeping the same add/remove diff logic.
- Hierarchy is wired explicitly in the import loop with the store API
  (`datastore.tasks.parent()`), relying on `__sort_todos` importing
  parents first. Conservative on purpose: links are only created, never
  re-parented nor removed from the DAV side.

Two more regression tests (parent/child import, categories import), suite
at 5 passing. Re-validated against a live Nextcloud: a real parent with
two subtasks and tagged todos land as real Task/Tag objects, zero
anomalies, and the hierarchy renders correctly in the browser.

Unrelated heads-up while testing: task store mutations from the sync
thread occasionally froze or crashed the UI (GLib ref-count assertion),
independently of this branch's changes. I have a mechanism map and will
file it as a separate core issue rather than widen this PR.

Another run on your side would be gold. If a third layer shows up, send
it over, this onion is getting peeled either way.

diegogangl:
Seems like this is blocked by #1241

pentux-GitHub:
Unblocked: the infobar port discussed in #1241 is up as #1284 (replace
with the number) — with it, backend errors show a banner instead of
freezing, so this PR can be re-tested.

pentux-GitHub:
Pushed an additional commit that fixes a CalDAV sync loop I hit while
testing this backend against a real Nextcloud task calendar: `SEQUENCE`
was bumped on every `_set_task`, even when `should_sync()` reported no
change, so each imported remote edit echoed back a higher `SEQUENCE` and
the same todos were rewritten on every cycle. The increment now happens
only when a real change is actually saved (RFC 5545 semantics).

Full analysis and reproduction in #1290. Verified against Nextcloud: an
unchanged task keeps a stable `SEQUENCE` and the loop is gone.

Happy to split this into its own PR if you'd rather review it
separately, but since the CalDAV backend only runs with this PR applied,
it seemed most useful to include it here.

diegogangl:
It seemed to work the first time, but afterwards GTG crashes on startup with the following:

```
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/gtk/application.py", line 121, in do_startup
    self.ds.find_and_load_file(data_file)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/datastore.py", line 478, in find_and_load_file
    self.load_file(filepath)
    ~~~~~~~~~~~~~~^^^^^^^^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/datastore.py", line 212, in load_file
    self.load_data(self.xml_tree)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/datastore.py", line 194, in load_data
    self.tasks.from_xml(tasks_xml, self.tags)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/tasks.py", line 917, in from_xml
    assert added_element.text is not None, 'Added text not found in task '+str(tid)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Added text not found in task 945ff268-6976-5b56-9824-68ada97d91ec
Gtk-Message: 09:52:28.885: GtkDialog mapped without a transient parent. This is discouraged.
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/gtk/application.py", line 161, in do_activate
    if self._check_exception():
       ~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/gtk/application.py", line 216, in _check_exception
    dialog.set_application(self)  # Keep application alive, for now
    ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'set_application'
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/gtk/application.py", line 672, in do_shutdown
    self.ds.save()
    ~~~~~~~~~~~~^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/datastore.py", line 250, in save
    assert path is not None, "Failed to determine save location."
           ^^^^^^^^^^^^^^^^
AssertionError: Failed to determine save location.

(gtg:679816): GLib-GIO-CRITICAL **: 09:52:28.916: GApplication subclass 'GTG+gtk+application+Application' failed to chain up on ::shutdown (from end of override function)

```

pentux-GitHub:
Thanks, that's a real bug and the traceback has everything needed —
reproduced and fixed in the last commit.

What happens: `CREATED` is optional in RFC 5545, and the field is in
`DateField`'s ignored-values list, so a VTODO without it never writes
to the task. The task therefore keeps the `Date.no_date()` it gets
from `Task()`. `str(Date.no_date())` is the empty string, so `to_xml`
writes `<added></added>`, lxml reads that back as `None`, and the
assert at tasks.py:917 aborts the whole load. Hence "worked the first
time, crashes afterwards": the damage is done at save time.

The fix gives imported tasks a fallback added date. I used
`date_modified` rather than `now()` on purpose: `now()` would claim
the task was created during the import, whereas LAST-MODIFIED is
always set and is a truthful upper bound. New test
`test_import_todo_without_created_date` fails on the previous commit,
passes on this one.

One thing this does *not* fix: your current file is already written
with an empty `<added>`, so it will still refuse to load. To recover
it:

(back up the file first, obviously)

    python3 - <<'EOF'
    from lxml import etree
    path = 'gtg_data.xml'   # adjust
    tree = etree.parse(path)
    fixed = 0
    for dates in tree.iter('dates'):
        added, modified = dates.find('added'), dates.find('modified')
        if added is not None and not added.text \
                and modified is not None and modified.text:
            added.text = modified.text
            fixed += 1
    tree.write(path, encoding='UTF-8', xml_declaration=True)
    print(fixed, 'tasks repaired')
    EOF

That leads to the wider point: the core currently writes a file it
then refuses to read. A task with no added date is serializable but
not loadable, and the assert takes the entire database down rather
than the one task. That's the same family as #1289, and probably
deserves its own fix in `from_xml` — happy to take it on separately if
you think that shape is right.

pentux-GitHub:
Following up on my own comment: I found your #1240 ("Ensure we have an
added date to task in caldav", July 2025) — same diagnosis, a year
before mine. It does exactly the right thing:

    if not task.get_added_date():
        task.set_added_date(datetime.now())

It just predates the core rewrite: `get_added_date()` /
`set_added_date()` are gone from `Task`, so it would `AttributeError`
on current master. That's presumably why it stayed open.

So I've moved my guard to your spot — end of `fill_task`, right after
the calendar tag — instead of the import loop. Same place, new API,
and it now also covers `_update_task`, which goes through `fill_task`
too. The only thing I did differently is falling back on
`date_modified` rather than `now()`: `now()` would date the task from
the import rather than from its actual life. Force-pushed as an amend
since the commit was minutes old.

That should make #1240 redundant — but it's your patch and your call,
so tell me if you'd rather keep yours and I'll drop mine.

Test `test_import_todo_without_created_date` still red before / green
after. Full local suite green, CI green, no conflicts with master.

diegogangl:
This is almost there!

I found 3 problems:

1. Some tasks have tags like this: `@DAV_Deck:_Server`. The `:` are not recognized as part of tags, so maybe they should be replaced by something else? 

2. Opening a task with these tags makes it so that everytime I open the task the tag is added again

3. Subtasks are syncronized from ths server, but not _to_ the server

4. I found this traceback in the terminal

```
2026-07-16 10:36:23,948 - ERROR - backend_caldav:_set_task:203 - Something went wrong while updating Task "List : Metrics" with id "55662985-81ea-52f8-a309-7f2c4e482361".Status: Status.ACTIVE, tags: DAV_Deck:_Server, DAV_Deck => Todo([REDACTED URL].ics)
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/backend_caldav.py", line 201, in _set_task
    todo.save()
    ~~~~~~~~~^^
  File "/home/januz/.local/lib/python3.14/site-packages/caldav/calendarobjectresource.py", line 1163, in save
    self._create(id=self.id, path=path)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/januz/.local/lib/python3.14/site-packages/caldav/calendarobjectresource.py", line 1004, in _create
    self._put()
    ~~~~~~~~~^^
  File "/home/januz/.local/lib/python3.14/site-packages/caldav/calendarobjectresource.py", line 975, in _put
    return self._put(False)
           ~~~~~~~~~^^^^^^^
  File "/home/januz/.local/lib/python3.14/site-packages/caldav/calendarobjectresource.py", line 977, in _put
    raise error.PutError(errmsg(r))
caldav.lib.error.PutError: PutError at '404 Not Found

<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\NotFound</s:exception>
  <s:message>Node with name 'stack-13.ics' could not be found</s:message>
</d:error>
', reason no reason
```

pentux-GitHub:
All four turned out to be real, and two had causes I had completely
missed. Rebased on current master (so #1284 is in: backend errors now
show a banner instead of tracebacks) and pushed four separate commits,
one per cause, so you can take, drop or reorder them independently.

**1 & 2 — tags with `:` — `only build tags the core can parse back`**

One bug. `to_tag()` only replaced spaces, so "Deck: Server" produced
`DAV_Deck:_Server`. GTG's tag parsers don't agree on what a tag may
contain: `core.tags` accepts `:`, while the editor's `TAG_REGEX` and
`core.tasks`' `TAG_LINE_REGEX` stop at anything outside `\w` and `-`.
So the editor read back `@DAV_Deck`, truncated, and re-added it as a
second tag on every open — exactly what you saw. The backend now only
builds tags every parser agrees on. (The parsers disagreeing is a core
issue of its own; I didn't want to widen this PR with it.)

**3 — subtasks not going up — two commits, two causes**

- `port the sort-order field to the new core API`: `OrderField` called
  `parent.get_child_index(uid)`, removed by the rewrite, so **every**
  push of a subtask raised AttributeError. That alone explains "from
  the server but not to the server".
- `send the server's own UID in RELATED-TO`: `get_parents` returned
  `str(task.parent.id)` — for a server whose UIDs aren't UUIDs, that's
  the uuid5 *I* derived from the UID, a value the server never issued,
  so it drops the link. The remote UID is now kept as a task attribute
  at import and used in RELATED-TO. This also closes the "known
  limitation" I flagged earlier about `_clean_task_missing_from_backend`.

**A fourth one you didn't report, found while testing this**

`let a rename reach the backends and the filters`: renaming a task
emitted no signal at all. `TaskStore.add()` only wires
`notify::is-actionable`, `notify::is-active` and `tags-changed` to
`task-filterably-changed` — the title, which is also what search
filters on, was missing. So the server kept whatever title the task
had when some other signal last fired: my subtasks reached Nextcloud
named "so", "sou" and "s", the first letters I had typed. This one is
in the core rather than the backend; it's the only reason the sync
looks right now. No storm of pushes either: the editor already
debounces its saves (`GnomeConfig.SAVETIME`).

**4 — the PutError**

Consequence rather than cause, I think. The failing task ("List :
Metrics") carried both `DAV_Deck:_Server` and the phantom `DAV_Deck`;
the truncated tag matched no calendar, so the backend fell back to the
default one and Deck refused. With the tag fixed it should resolve to
its own calendar and stop being marked modified on every open. If Deck
still refuses a genuine edit, that's a read-only remote and GTG can
only log it — which it does. I get the same class of refusal here on a
share I can't write to; #1291 is about that noise.

**Verified live against Nextcloud**: a parent, two subtasks and a
grand-child created in GTG now appear correctly nested on the server,
with full titles, tags and dates. 236 tests passing locally, CI green,
each commit red before / green after.

Still open, for later: only a title change notifies backends — editing
just the body of a task doesn't (`content` isn't a GObject property).
Worth its own fix, out of scope here.

If four commits at once is too much to take, rolling back to `always
give imported tasks an added date` and merging that as a baseline is a
perfectly reasonable call — I'd reopen the rest as its own PR. Your
repo, your rhythm.

diegogangl:
Almost everything works. Tested with a task that a subtask which also had a subtask. The middle subtask got a tag in the title (`@DAV_Personal`) and the content of the root task was broken.
Also we get the tag repeating every time I open the task. And "Otra subtask" is also repeated using markdown syntax

```Una nueva task para sync
Let's see how it takes sync, @DAV_Personal, @DAV_Personal

[ ] Otra subtask
[ ] Otra subtask
Otra subtask

```

diegogangl:
I also got a traceback here:

```
Exception in thread Thread-1 (__backend_startup):
  File "/usr/lib64/python3.14/threading.py", line 1082, in _bootstrap_inner
    self._context.run(self.run)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/usr/lib64/python3.14/threading.py", line 1024, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/datastore.py", line 703, in __backend_startup
    backend.start_get_tasks()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/interruptible.py", line 38, in new
    return fn(*args)
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/periodic_import_backend.py", line 79, in start_get_tasks
    self._start_get_tasks()
    ~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/periodic_import_backend.py", line 98, in _start_get_tasks
    self.do_periodic_import()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/core/interruptible.py", line 38, in new
    return fn(*args)
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/backend_caldav.py", line 114, in do_periodic_import
    self._do_periodic_import()
    ~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/backend_caldav.py", line 148, in _do_periodic_import
    self._import_calendar_todos(calendar, start, counts)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/backend_caldav.py", line 292, in _import_calendar_todos
    calendar_tasks = dict(self._get_calendar_tasks(calendar))
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/backends/backend_caldav.py", line 352, in _get_calendar_tasks
    yield uid, task # noqa: F821
          ^^^
NameError: name 'uid' is not defined. Did you mean: 'id'?

```

pentux-GitHub:
Thanks for the new round! Before anything else: the traceback in your
last comment can't come from this branch. The line it quotes,
`yield uid, task  # noqa: F821`, only exists on master (the noqa was
added there in Jan 2025); this branch rewrote that line in its very
first commit (now `yield str(task.id), task`, line 407). The line
numbers agree: backend_caldav.py:352 and datastore.py:703 are current
master, and :703 only shifted to that position with today's #1293
merge. The branch itself hasn't moved since my last push (76ff3b3).

So for the crash: checking out fix/caldav-new-core-compat and
rebuilding should make it disappear, since it's literally the #1176
bug this branch fixes.

For the content damage in your previous comment: since it's most
likely the same build, I can't tell residue from regression yet. Your
data file and the Nextcloud calendar also carry scars from the earlier
rounds (including the truncated-tag era). A clean pass would be:
branch checked out, rebuilt, and ideally a fresh data file, otherwise
we're measuring the wreckage of past tests.

One thing I'm not waving away: `@DAV_Personal` has no character the
tag parsers disagree on, so if the re-tagging on open still happens on
the branch with clean data, my truncation explanation from the 16th
doesn't cover it. I'm reading the editor side of that right now either
way, and I'll report here.

pentux-GitHub:
Update on the re-tagging, as promised: it's explained, and it's not this
branch. It's a pre-existing race in the editor, on master.

Short chain: `Task.add_tag()` (tasks.py:303) puts the tag in the task's
set but never touches the content. On every editor open,
`set_text_from_task` (taskview.py:166) re-inserts all the task's tags
into the text. Its "already in the buffer" guard checks `task_tags`,
which is only filled by `process()` on a 250 ms timer, so at open time
it checks an empty set and filters nothing. `get_text()` then saves the
duplicate. One more copy per open/close cycle.

CalDAV is just the main way tags reach the model without being typed, so
you'll still see the re-tagging after checking out this branch: it's not
a signal against the PR. I'll file it as its own core issue.

The duplicated subtasks: I suspect the same family
(`set_text_from_task` also re-inserts model subtasks it didn't detect in
the text, lines 188-192), but the full mechanism depends on what earlier
pushes wrote into the DESCRIPTION, so that part is not established and
I'm not claiming it. I need time to dig, and I'll be less available over
the coming days, so it may take a little while. The re-test on the
branch with fresh data is still the most useful next step meanwhile.

pentux-GitHub:
One scope note before your next test pass, so expectations are right: **this PR does not cover pushing local deletions to the server.**

While re-reviewing the branch I traced the removal path: nothing ever calls `queue_remove_task`, so deleting a task in GTG never reaches the DAV server. Predicted symptom (stated, not yet tested): a locally deleted task should reappear at the next periodic import. This is not a regression — current master pushes nothing at all, deletions included — but I'd rather state it explicitly than have you find it mid-test.

I'd propose keeping it out of this PR, which is about restoring compatibility, and wiring the deletion path in a follow-up: it needs the datastore's removal signal connected to `queue_remove_task`, plus the backend's `remove_task`/`_remove_task` pair sorted out (right now the public one touches the local store and the DAV delete has no caller). That deserves its own focused review. If you'd rather have it fixed in this PR before merging, I can do that instead — your call.

diegogangl:
Fair enough. This PR is getting a bit long to review too. I will merge this one for now (since it LGTM)

PR #1266, opened 2026-03-02
Fix a few unexpected task folding/unfolding

### LINKED ISSUE #1245: When closing a task editor viewer window, all tasks in the main window task browser's "Open" view get folded
Testing with the current git `master` version, I found a somewhat minor issue (i.e. it presumably doesn't cause data loss, but would still be annoying). With the default welcome dataset that has a tutorial task with subtasks:

1. Unfold the task, so that you can see all its children in the "Open" view
2. Create a new top-level task
3. Open that newly created top-level task's editor
4. Close that task editor window

Result: the tutorial task gets folded, so the children are no longer visible. I presume if there were multiple parent tasks present, they would probably all get folded.

Demonstration video:

https://github.com/user-attachments/assets/b37119aa-4ccc-4697-9537-a4628a4fef3b

gycsaba96:
Fixes #1245 

The source of the bug was that the underlying list models were always cleared and rebuilt when a task refiltering was triggered. Now, it only updates what really needs to be moved. While testing, I also noticed another similar bug where tasks get expanded when switching between panes.

- Update the underlying lists within `FilteredTaskTreeManager` instead of rebuilding them in a folded state. (I also added some tests so that we can change things more confidently. The helper functions make it easy to add more test cases.) 
- Remove an old hack from `MainWindow` that caused unexpected unfolding when switching between panes.

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -716,16 +716,8 @@ class FilteredTaskTreeManager:
 
 
     def _refilter_all_tasks(self) -> None:
-        self._clear_models()
         for t in self.store.data:
             self._update_with_descendants(t)
-
-
-    def _clear_models(self):
-        self.tid_to_containing_model = dict()
-        self.root_model.remove_all()
-        for model in self.tid_to_subtask_model.values():
-            model.remove_all()
 
 
     def _on_changed(self,*args):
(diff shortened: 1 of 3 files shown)


diegogangl:
Changes LGTM but it has some conflicts

gycsaba96:
Thank you for letting me know, @diegogangl!

I rebased the changes, and everything seems to be in order now.

diegogangl:
Nice, looks good to merge now. Thanks!

PR #1270, opened 2026-04-05
fix: now you can close preference window with Esc key

LucianoXHDX:
The preferences windows now can close with Esc key like all windows, for that adds a key-press-event handler that 
calls hide() when Escape is pressed, with that its must be close.


diegogangl:
LGTM, thanks for also fixing the whitespace issue @LucianoXHDX !

PR #1271, opened 2026-04-16
fix: Escape key only closes searchbar, does not reopen it

### LINKED ISSUE #1243: `Escape` key reopens the closed searchbar, and can be called from anywhere in the main window
This is a very minor bug I have found in the git `master` version today, related to #1003 and #952. It seems that #1106 (or something else afterwards) introduced a slight oversight where you can reopen the searchbar by pressing the `Esc` key.

To reproduce:
1. `Ctrl+F`
2. Press `Esc` to dismiss the searchbar
3. Either press `Esc` immediately, or select something else in the main window (ex: click in the quick-add entry) and press `Esc`

Result: the searchbar reopens (essentially, `Esc` acts as a toggle). That's not really a normal/expected behavior, I've never seen that in other applications. When the searchbar is not focused, the `Esc` key would normally be used to cancel or defocus whatever else is focused (if applicable), so it's probably best if the searchbar doesn't interfere with that shortcut when it doesn't have focus.

impsislegobatman:
## Summary

Fixes #1243

The Escape key was mapped to `toggle_search`, which would reopen the searchbar after closing it with Ctrl+F. 

## Changes

- Added a `close_search()` method that only dismisses the searchbar when it is currently active (no-op when already closed)
- Mapped the Escape key action to `close_search` instead of `toggle_search`
- Refactored `toggle_search` to use `close_search` internally

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -223,7 +223,7 @@ class MainWindow(Gtk.ApplicationWindow):
             ('focus_sidebar', self.focus_sidebar, ('win.focus_sidebar', ['<ctrl>B'])),
             ('toggle_search', self.toggle_search, ('win.toggle_search', [])),
             ('search', self.activate_search, ('win.search', ['<ctrl>f'])),
-            ('close_search', self.toggle_search, ('win.close_search', ['Escape'])),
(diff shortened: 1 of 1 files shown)


diegogangl:
Sorry for the delay. Merged, thanks!

PR #1275, opened 2026-07-08
fix: don't crash when closing the date picker before it was shown (#1195)

### LINKED ISSUE #1195: Opening the "Synchronization Services" window causes a traceback with AttributeError: 'Window' object has no attribute '_GTGCalendar__sigid'
Trying out #1191 which is only 2 commits atop of current git `master`, I accidentally clicked the "Synchronization Services" menu item in the main hamburger menu, and I got this empty window as a result:

Image

… with this accompanying traceback:

```python-traceback
  File "/app/lib/python3.12/site-packages/GTG/gtk/editor/calendar.py", line 157, in _esc_close
    self.close_calendar()
  File "/app/lib/python3.12/site-packages/GTG/gtk/editor/calendar.py", line 110, in close_calendar
    if self.__sigid is not None:
       ^^^^^^^^^^^^
  File "/app/lib/python3.12/site-packages/GTG/gtk/editor/calendar.py", line 151, in __getattr__
    return getattr(self.__window, attr)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Window' object has no attribute '_GTGCalendar__sigid'
```

I wonder if it's somehow related to @SqAtx's issue #1176, since both are about a "calendar" of some sort.

**GTG:** 0.7-dev-5bd604bb
**Flatpak:** 1.16.0
**Snap:** False
**Display Name:** wayland-0
**Desktop:** GNOME

**lxml:** True
**caldav:** True
**Cheetah3:** True
**dbus-python:** True
**pdflatex:** False
**pypdftk:** False

pentux-GitHub:
## Summary

Fixes #1195

`GTGCalendar.close_calendar()` crashes with `AttributeError: 'Window' object has
no attribute '_GTGCalendar__sigid'` when it runs before the calendar was ever shown.

## Analysis

- `self.__sigid` is only created in `show()`, when the "day-selected" signal
  gets connected.
- `close_calendar()` already guards with `if self.__sigid is not None:`, but if
  `show()` was never called, the attribute does not exist at all, so the guard
  itself raises.
- The class delegates unknown attributes to the wrapped window through
  `__getattr__`, which explains the unusual error message mentioning the
  'Window' object.

The close paths (`close_calendar`, `_esc_close`) can be reached independently of
`show()`: the Escape shortcut is registered at construction time.

## Changes

Initialize `self.__sigid = None` in `__init__`, next to the other state
attributes. The existing guard in `close_calendar()` then works as intended.
One line changed.

## Testing

Deterministic reproduction, no UI interaction needed (run from the repo root):

    import gi
    gi.require_version('Gtk', '4.0')
    gi.require_version('GtkSource', '5')
    from gi.repository import Gtk
    Gtk.init_check()

    from GTG.gtk.editor.calendar import GTGCalendar

    cal = GTGCalendar()      # show() never called
    cal.close_calendar()     # AttributeError on master, harmless with this fix
    cal._esc_close()         # the other close path
    print("No crash: closing before show() is harmless")

On master this reproduces the exact traceback from #1195. With this fix, both
calls are harmless no-ops.

Also manually tested the custom date picker in the task browser (the only place
GTGCalendar is instantiated): picking a date, reopening, closing with Escape,
everything works and the terminal stays clean.

### DIFF
--- a/GTG/gtk/editor/calendar.py
+++ b/GTG/gtk/editor/calendar.py
@@ -45,6 +45,7 @@ class GTGCalendar(GObject.GObject):
         self.__builder.add_from_file(GnomeConfig.CALENDAR_UI_FILE)
         self.__date_kind = None
         self.__date = Date.no_date()
+        self.__sigid = None
         self.__init_gtk__()
 
     def __init_gtk__(self):


diegogangl:
LGTM, thanks!

Issue #403, opened 2020-07-09
Clicking the "Now", "Soon", "Someday" buttons in the date picker widgets should dismiss the date pickers


nekohayo:
Not a huge deal, but 0.4 doesn't dismiss the date picker popovers (at least not in the Task Editor) when clicking the preset date buttons; it would make sense if it did (even for the Clear button I guess, because if the user wanted to replace the date they could simply click a date in the calendar).

diegogangl:
Sort of related to #314
TBH I don't think they should close. It only takes clicking away to make it pop down and you can change your mind while you have focus in the input field.

nekohayo:
Well, a symptom/problem with that at least, is that when I hit the "Escape" key to try to dismiss the date picker widget, it closes the whole task editor, which is jarring and much more unexpected than a popover being dismissed upon a successful choice  being made with an activation (?) mouse or keyboard event; in theory, the popover closing is kind of visual feedback that the command has been "acknowledged". Also, I've never changed my mind within seconds of choosing a start/due date; wanting to continue editing the rest of the task's contents, however, _that_ I want to do ;)

However,  looking at #314, if GTK creates a bunch of trouble with that and the calendar widget, then I guess I could settle for "at least temporarily remap the Escape key to apply only to the popover"...

pentux-GitHub:
Six years later, this is still reproducible on current master (252a8fca, 0.7-dev, Debian testing, GTK 4). I dug into the mechanism behind the Escape part, since it is the most jarring one. Full analysis below, PR coming.

## Why Escape closes the whole editor

Three pieces combine:

1. **A global accelerator.** `app.close` is bound to Escape at the application
   level, in `GTG/gtk/application.py`:
   `('close', self.close_context, ('app.close', ['Escape']))`

2. **The router never asks about popovers.** `close_context()` checks whether a
   task editor is active, or whether the search entry has focus. If an editor is
   active, it closes the task. There is no "is a popover open?" question in
   between.

3. **The date popovers do not consume Escape themselves.** In `task_editor.ui`,
   `start_popover`, `due_popover` and `closed_popover` are all declared with
   `autohide=False`. A non-autohide popover does not grab focus and does not
   handle Escape, so the key event travels up and lands in the global
   accelerator, which closes the editor.

Instrumentation confirms it: with a temporary `print()` at the top of
`close_context()`, pressing Escape while a date popover is open triggers the
print, and the editor closes.

[...]

PR #1276, opened 2026-07-08
fix: Escape closes the open date popover instead of the whole task editor (#403)

pentux-GitHub:
## Summary

Addresses the Escape key part of #403 (not using the "Fixes" keyword on
purpose: the original report also covers the preset buttons behavior, which
this PR deliberately leaves untouched).

Full root-cause analysis in
https://github.com/getting-things-gnome/gtg/issues/403#issuecomment-4910225357 (my comment
from today). Short version below.

## Root cause

Three pieces combine:

- `app.close` is bound to Escape at the application level
  (`GTG/gtk/application.py`), so the accelerator fires anywhere in GTG.
- The date popovers (`start_popover`, `due_popover`, `closed_popover`) are
  declared `autohide=False` in `task_editor.ui`, so they do not grab focus and
  do not consume Escape themselves. This is a deliberate design inherited from
  `modal=False` in the GTK 3 UI files (converted in 90f8ec5b): it lets the user
  keep typing in the date entry while the calendar stays visible. Flipping
  autohide would also change focus and click-outside behavior, hence this PR
  does not touch the .ui files.
- `close_context()` only asks "is an editor active?" and closes the whole
  task, never asking whether a popover is open.

## Changes

- New `TaskEditor.close_open_date_popover()`: pops down the currently visible
  date popover, if any, and reports whether it did.
- `close_context()` asks that question first: if a date popover was open, the
  first Escape only dismisses it; the second Escape closes the editor, as
  before.

This is the behavior @nekohayo suggested in the issue ("remap the Escape key
to apply only to the popover"), and follows the same pattern as the recent
Escape fixes #1270 and #1271.

## Testing

Manually tested on Debian (GTK 4, 0.7-dev):

- Editor open, no popover, Escape: editor closes (historical behavior kept).
- Editor open, start/due/closed date popover open, first Escape: only the
  popover closes, the editor survives. Second Escape: the editor closes.
- Recurrence popover: still closes itself on Escape (GtkPopoverMenu,
  autohide default), unaffected.
- Main window search bar and Escape: unaffected (no change to that branch of
  `close_context`).

Terminal stays clean during all scenarios.

### DIFF
--- a/GTG/gtk/application.py
+++ b/GTG/gtk/application.py
@@ -439,6 +439,8 @@ class Application(Gtk.Application):
         search = self.browser.search_entry.is_focus()
 
         if editor:
+            if editor.close_open_date_popover():
+                return
             self.close_task(editor.task.id)
         elif search:
             self.browser.toggle_search(action, params)
--- a/GTG/gtk/editor/editor.py
+++ b/GTG/gtk/editor/editor.py
@@ -247,6 +247,23 @@ class TaskEditor(Gtk.Window):
 
         self.closed_popover.popup()
 
+    def close_open_date_popover(self) -> bool:
+        """Close the currently open date popover, if any.
+
+        Returns True if a popover was closed. Used by the application's
+        Escape accelerator so that a first Escape only dismisses the
+        popover instead of closing the whole editor (see issue #403).
+        """
+
+        for popover in (self.start_popover,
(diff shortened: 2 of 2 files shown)


diegogangl:
LGTM, thanks!

PR #1278, opened 2026-07-09
fix: close a freshly created task by id and harden Task.__eq__ (#1273)

### LINKED ISSUE #1273: Traceback (AttributeError: 'UUID' object has no attribute 'id') when trying to immediately delete a newly created task
Tested with **GTG:** 0.7-dev-bef9282a (built with GNOME Builder on git commit 252a8fca3)

---

## To reproduce

Build and launch the latest `master` git version, and:
1. `Ctrl+N` to create a new task, and don't touch anything in its contents
2. Click the kebab menubutton ("Task action menu")
3. Notice that the "Delete" action is available; click it

## Actual Results

**Context:** Global generic exception

```python-traceback
  File "/app/lib/python3.12/site-packages/GTG/gtk/application.py", line 461, in delete_editor_task
    self.close_task(task)
  File "/app/lib/python3.12/site-packages/GTG/gtk/application.py", line 609, in close_task
    editor = self.open_tasks[tid]
             ~~~~~~~~~~~~~~~^^^^^
  File "/app/lib/python3.12/site-packages/GTG/core/tasks.py", line 647, in __eq__
    return self.id == other.id
                      ^^^^^^^^
AttributeError: 'UUID' object has no attribute 'id'
```

## Expected Results

* No traceback; either safely discard the no-op newly created task, or don't let me use the delete function?
* More broadly: should any of the three core actions (Mark as done vs Dismiss vs Delete) be sensitive/available at all, until the user has actually modified the task in a way that makes it saveable?

pentux-GitHub:
## Summary

Fixes #1273

Deleting a freshly created, still-untouched task from its editor crashed
with `AttributeError: 'UUID' object has no attribute 'id'`.

## Root cause

`delete_editor_task()` was the only one of the four `close_task()` call
sites passing the `Task` object instead of `task.id`. The failure mode
is sneaky: `Task.__hash__` delegates to `hash(self.id)`, so the stray
Task lands in exactly the right bucket of the `open_tasks` dict (which
is keyed by UUID), the dict then checks equality, and `Task.__eq__` does
`other.id` on a bare UUID.

## Fix (two one-liners)

- `application.py`: pass `task.id` like every other call site.
- `tasks.py`: make `Task.__eq__` follow the Python contract and return
  `NotImplemented` for foreign types, so this whole class of traps
  degrades to a clean `False`/`KeyError` instead of a crash. (Same
  type-safety family as the `str`-in-`children` issue being fixed in
  #1265.)

## Tests

New `tests/core/test_task_equality.py`: the foreign-type comparison test
fails with the exact reported AttributeError on master and passes with
the fix; Task-vs-Task equality is covered too. Also verified in the UI:
Ctrl+N then immediate delete now closes the editor and removes the task
cleanly, no traceback.

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -644,6 +644,9 @@ class Task(StoreItem):
     def __eq__(self, other) -> bool:
         """Equivalence."""
 
+        if not isinstance(other, Task):
+            return NotImplemented
+
         return self.id == other.id
 
 
--- a/GTG/gtk/application.py
+++ b/GTG/gtk/application.py
@@ -458,7 +458,7 @@ class Application(Gtk.Application):
         task = editor.task
 
         if editor.is_new():
-            self.close_task(task)
+            self.close_task(task.id)
         else:
             self.delete_tasks([task], editor)
 
--- a/tests/core/test_task_equality.py
+++ b/tests/core/test_task_equality.py
@@ -1,0 +1,21 @@
+from unittest import TestCase
+from uuid import uuid4
+
+from GTG.core.tasks import Task
+
+
+class TaskEqualityTest(TestCase):
+    """Task.__eq__ must follow the Python contract: comparing with a
+    foreign type (e.g. a bare UUID used as a dict key) must not raise
+    (regression test for #1273)."""
+
+    def test_equality_with_task(self):
+        tid = uuid4()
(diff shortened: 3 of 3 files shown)


diegogangl:
Merged, thanks!

Issue #109, opened 2015-05-24
Very slow to display ~250-500 tasks


nekohayo:
Essentially, https://bugs.launchpad.net/gtg/+bug/587307 still affects the current (0.3.1) version. If this was solved, there would be no need for issue #43.

sametmax:
+1

mkorvas:
After a crash, `gtg` has been eating 100% CPU for whopping 40 *minutes* that it's been launching so far. There is no indication of how much longer it's going to take. I normally don't stop `gtg` exactly for this reason -- by the time it has launched, half of my working day is gone. It's doing something really stupid, my tasks file has some 750 `<task>` elements, really not something to justify so much CPU time.

mkorvas:
(Not sure why Github says that I unassigned @izidormatusov ; I didn't mean it that way.)

diegogangl:
Of course, this needs to be profiled but IMO python + xml = performance poison. I bet most of that time is spent on disk IO and _sloooowly_ looping over all the xml tags. There are probably some quick fixes to improve performance a little, but the real deal would be switching to sqlite or something like that.

ksamuel:
Not at all, using lxml, most of the XML processing is done using C. As for I/O, no language is ever the bottleneck: the slowest one is a 100 times faster than disk access. The solution is simply to work in memory, and dump the result only when necessary, and in a separated thread.

All in all, the biggest problem of GTG is not performance, although I'm pretty sure the code is suboptimal give the small loads it has and how slow it is, but the fact it's doing the work in the same thread as the UI: because of this, any slow down freeze the UI and make it feels sluggish.

ploum:
The process to read the XML file was really optimised and in its own thread. The whole thing prompted a huge rewrite of liblarch to be multi-thread safe, something which was really hard because GTK was not multi-thread safe.

But this could be a simple bug preventing thread to operate in parallel due to a lock somewhere.

diegogangl:
My bad guys, I got this mixed up with #43 which involves reading/writing to xml. 
So yeah, besides profiling and finding that possible bug moving to lxml would surely improve performance.

diegogangl:
I profiled this with cProfile, looks like the main culprit is in in liblarch. I don't really know much about liblarch but it seems to be the `__sort_func` being called everytime a row is inserted in the model.

We should revisit this for 0.5

Here's the prof in case anyone wants to take a stab:
[...]

PR #1281, opened 2026-07-09
fix: resurrect and wire the 0.6 data migration (#855)

### LINKED ISSUE #855: Starting with bryce set raises parent/child exceptions
There are a few different errors when starting with bryce task set:
- Exception: 378be418-0570-40da-9ee5-e676975c77fa is not children of 91315d09-30b2-4317-af35-a8b7621cbcd0
- Exception: 5e80bbd9-a169-4e7b-9bc6-77e0029ed87c is not children of 417363d0-fac4-4e61-aeaf-064f26f87e3d
- Exception: 7e2fe2f5-f54f-4102-bac9-062f0030b25b is not children of 2b9fb936-6426-4f59-8a73-b9b59437d013
- Exception: 7e2fe2f5-f54f-4102-bac9-062f0030b25b is not children of b165f772-740c-4694-b7e4-96343febb1e6
- Exception: 60d983be-7897-4619-824d-bfddbbf79a67 is not children of 3fba3405-9e77-49f5-8008-02275671a239
- Exception: 5fd4a96d-ac52-49a7-a6ae-163dbeb37b52 is not children of 36807b16-3108-4fec-8598-f129aaf10347

It's possible to just ignore them and continue, but when switching between tags, they raise again.

pentux-GitHub:
## Summary

Fixes #855, and unblocks #109 (the bryce dataset becomes loadable, so
startup performance can finally be measured).

The 0.6 → 0.7 data migration was doubly dead on master:

- `versioning.py` called `xml.open_file()`, but that module was deleted
  in e43eee0a ("Remove old core code", Oct 2022): the converter would
  NameError on its first line if anything reached it.
- Nothing reached it: the file-not-found branch of
  `Datastore.find_and_load_file()` called `first_run()` directly,
  bypassing `do_first_run_versioning()`, which was written for exactly
  this job (find old file, convert, rename to `.imported`, else fall
  back to first run).

Net effect: any 0.6 user upgrading to 0.7 silently got the "Getting
started" tutorial instead of their data (left untouched but ignored
forever, and shadowed by the freshly saved empty file).

## Fix

- A minimal `_open_file()` inside `versioning.py`, modeled on the
  deleted one (strict parse + root tag check). The original's
  backup-rotation logic is intentionally not reimplemented: migration
  sources are existing files, and failing loudly beats silently
  ignoring user data.
- The file-not-found branch now goes through
  `do_first_run_versioning()`.

## Tests and verification

Two regression tests, both red on master: `convert()` NameErrors, and
the startup test's failure message is a portrait of the bug (the nine
tutorial titles where 'old parent' should be).

End-to-end on a real machine with the in-repo bryce dataset
(`./launch.sh -s bryce`): 1771 tasks and 659 subtask links migrated in
under a second, old file renamed to `.imported`, app fully usable. The
~50 `item not found in data` error logs during load are dangling
RELATED references inside that dataset, handled gracefully by
`base_store.parent()` — these are the "parent/child exceptions" this
issue originally reported, now demoted to log noise.

- `datastore.py:247` logs "Error renaming temp file" on any first save
  into a virgin data dir (pre-existing, also fires on plain first run).
- A cosmetic `Child name 'default' not found in GtkStack` warning at
  startup.

### DIFF
--- a/GTG/core/datastore.py
+++ b/GTG/core/datastore.py
@@ -496,9 +496,11 @@ class Datastore:
         # We couldn't open any file :(
         if not self.xml_tree:
             try:
-                # Try making a new empty file and open it
-                self.first_run(path)
+                # No data file yet: migrate old-format
+                # data if present, otherwise create the
+                # first-run file (see #855)
                 self.data_path = path
+                self.do_first_run_versioning(path)
 
             except IOError:
                 raise SystemError(f'Could not write a file at {path}')
--- a/GTG/core/versioning.py
+++ b/GTG/core/versioning.py
@@ -37,10 +37,26 @@ tags_cache = {}
 tid_cache = {}
 
 
+def _open_file(path: str, root_tag: str) -> et._ElementTree:
+    """Open an XML file and check its root tag.
+
+    Minimal replacement for the old core's _open_file(), which was
+    removed in e43eee0a together with the rest of the old core. The
+    backup-rotation logic of the original is intentionally not
+    reimplemented: migration sources are existing files, and failing
+    loudly beats silently ignoring user data.
+    """
+    tree = et.parse(path)
+    if tree.getroot().tag != root_tag:
+        raise ValueError(f'{path}: expected root tag {root_tag!r}, '
+                         f'found {tree.getroot().tag!r}')
+    return tree
+
+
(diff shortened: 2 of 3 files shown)


diegogangl:
Looks good, thanks!

PR #1282, opened 2026-07-10
fix: keep sidebar tags sorted

### LINKED ISSUE #1251: Tags are not kept sorted alphabetically in sidebar
In the current git `master` version, since the port to GTK4, tags are not sorted alphabetically, ever.

* [ ] Newly created tags (ex: by typing a whole new tag in the contents of a task, then closing the task editor) are not inserted sorted alphabetically into the sidebar
* [ ] Even after a restart, tags in the sidebar remain in the order of creation, rather than a more natural alphabetical order

The 2nd point tells me that they are surely not sorted within the tags list at the top of the file format either. That said, perhaps the sorting should be done only at the UI level, and kept manual in the file format, so that it is technically possible someday to toggle alphabetical-vs-manual sorting on/off at the GUI level for #523.

---

Note: when fixing this, the tag's hierarchy levels need to be taken into account, because tags can be nested as parents vs children (see also #1249 for example).

So if I have tags like these:

* a
* c
  * d
  * e
  * f
* g

…and I insert tag "b", if it's a top-level tag it would go here:

* a
* b ← new tag!
* c
  * d
  * e
  * f
* g

…but if I make it a child of "c", then it would look like this:

* a
* c
  * b ← new tag!
  * d
  * e
  * f
* g

aouxwoux:
## Summary

- keep root tags alphabetically ordered in the GTK4 sidebar model
- keep child tags sorted within their existing parent hierarchy
- add regression coverage for newly created root and nested tags

## Why

Tags currently appear in creation/XML order, which makes the sidebar difficult to scan. This addresses #1251 without changing the serialized tag order or parent relationships.

## Validation

- `python -m compileall -q GTG/core/tags.py tests/core/test_tag_store.py`
- `git diff --check`
- targeted pytest could not run because pytest is not installed in this environment

Fixes #1251

### DIFF
--- a/GTG/core/tags.py
+++ b/GTG/core/tags.py
@@ -166,11 +166,24 @@ class TagStore(BaseStore[Tag]):
 
         # open the first one
         if item.children:
-            for child in item.children:
+            for child in sorted(item.children, key=lambda child: child.name.casefold()):
                 model.append(child)
 
         self.tid_to_children_model[item.id] = model
         return Gtk.TreeListModel.new(model, False, False, self.model_expand)
+
+
+    @staticmethod
+    def _insert_sorted(model: Gio.ListStore, item: Tag) -> None:
+        """Insert a tag in alphabetical order without changing its parent."""
+        name = item.name.casefold()
+        for position in range(model.get_n_items()):
+            current = model.get_item(position)
+            if name < current.name.casefold():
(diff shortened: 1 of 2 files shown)


diegogangl:
Great idea! LGTM, thanks!

PR #1283, opened 2026-07-10
fix: use available icon for all tasks

### LINKED ISSUE #1274: "All tasks" icon is broken in master when running with GNOME Builder
I grabbed the latest version of git master (commit 252a8fc), built and ran it using GNOME Builder, and encountered this broken icon in the top-left corner, for the "All tasks" special tag:

Image

It's not the end of the world, but if we can fix that in time for 0.7 it would look nicer :)

aouxwoux:
## Summary

- replace the unavailable `emblem-documents-symbolic` icon used by the All Tasks sidebar row
- use the existing `view-list-symbolic` icon already supported by the sidebar

## Why

The current icon renders as broken when GTG is run from GNOME Builder because it is not bundled and is not available in that icon theme. This fixes #1274 with a one-line, theme-compatible change.

## Validation

- `python -m compileall -q GTG/gtk/browser/sidebar.py`
- `git diff --check`

Fixes #1274

### DIFF
--- a/GTG/gtk/browser/sidebar.py
+++ b/GTG/gtk/browser/sidebar.py
@@ -181,7 +181,7 @@ class Sidebar(Gtk.ScrolledWindow):
         button.set_margin_start(6)
         button.set_margin_end(6)
 
-        tags_icon = Gtk.Image.new_from_icon_name('view-list-symbolic')
+        tags_icon = Gtk.Image.new_from_icon_name('tag-symbolic')
         tags_icon.set_margin_end(6)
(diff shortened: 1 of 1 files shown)


diegogangl:
Hi, thanks for the PR. This is a good idea, but now the icon for tags and all tasks are the same

diegogangl:
In the end, it was better to use another icon for tags since the one you picked was perfect for "all tasks".
Took the liberty of pushing that here. Thanks for the PR!

Issue #1057, opened 2024-03-08
AttributeError: 'str' object has no attribute 'get_id' when syncing CalDAV


cpitclaudel:
I'm trying to setup CalDAV sync, and I got the following issue:

```
2024-03-07 07:55:30,052 - INFO - backend_caldav:_do_periodic_import:137 - Running periodic import
2024-03-07 07:55:30,237 - INFO - backend_caldav:_do_periodic_import:144 - Fetching todos from '<redacted-1>'
2024-03-07 07:55:30,714 - INFO - backend_caldav:_do_periodic_import:144 - Fetching todos from '<redacted-2>'
Exception in thread Thread-3 (_internal_flush_all_tasks):
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.10/threading.py", line 953, in run
2024-03-07 07:55:31,120 - INFO - backend_caldav:_do_periodic_import:137 - Running periodic import
    self._target(*self._args, **self._kwargs)
  File "/home/cpc/git/gtg/.local_build/install/local/lib/python3.10/dist-packages/GTG/core/datastore.py", line 664, in _internal_flush_all_tasks
    backend.queue_set_task(task.id)
  File "/home/cpc/git/gtg/.local_build/install/local/lib/python3.10/dist-packages/GTG/backends/generic_backend.py", line 680, in queue_set_task
    tid = task.get_id()
AttributeError: 'str' object has no attribute 'get_id'
```

I added a `print(task)` right before the faulty line, and I got:

```
33d9760d-2e07-4ae4-9c02-1fcdaeb46325
```

This is the faulty function:

```
    def queue_set_task(self, task):
        """ Save the task in the backend. In particular, it just enqueues the
        task in the self.to_set queue. A thread will shortly run to apply the
        requested changes.

        @param task: the task that should be saved
        """
        tid = task.get_id()
        if task not in self.to_set and tid not in self.to_remove:
            self.to_set.appendleft(task)
            self.__try_launch_setting_thread()
```

Looks like a simple mistake in the caller?

```
        def _internal_flush_all_tasks():
            backend = self.backends[backend_id]
            for task in self.tasks.data:
                if self.please_quit:
                    break
                backend.queue_set_task(task.id) # ← Should be `task`?
```

[...]

PR #1285, opened 2026-07-10
fix: persist backend configuration across restarts (#930, #845)

### LINKED ISSUE #845: CalDAV backend saves incomplete configuration on Chromebook
Running GTG 0.6 stable flatpak on my Chromebook and using CalDAV sync.

After configuring the backend an initial sync happens and synced task appear as expected. After closing GTG and re-openning, startup fails with the following error:

**Context:** Startup

```python-traceback
  File "/app/lib/python3.9/site-packages/GTG/gtk/application.py", line 110, in do_startup
    for backend_dic in BackendFactory().get_saved_backends_list():
  File "/app/lib/python3.9/site-packages/GTG/backends/__init__.py", line 143, in get_saved_backends_list
    module = self.get_backend(settings.get('module'))
  File "/app/lib/python3.9/site-packages/GTG/core/config.py", line 164, in get
    raise ValueError(
ValueError: No valid configuration value or default value was found for %s in %s
```

**Software versions:**
* Getting Things GNOME! v0.6
* CPython 3.9.9 (main, Nov 10 2011, 15:00:00)   [GCC 11.2.0]
* GTK 3.24.31, GLib 2.70.2
* PyGLib 3.42.0, PyGObject 3.42.0
* Linux-5.4.163-17384-g99ca1c60d20c-aarch64-with-glibc2.33

Inspecting the `backend.conf` I notice that a number values are missing or incorrect (see below). Adding and correcting also causes startup to fail because the password had not been stored.

```
...
[backend_caldav]
first_run = True
default = False
human-friendly-name =
user =
pid = 58a617ea-7590-41ff-8bfa-58ec6c5f83de
enabled = True
attached-tags = gtg-tags-all
period = 1
username = steve
```

### LINKED ISSUE #930: GTG allows creating multiple backends of type CalDAV, but they do not persist across restarts

pentux-GitHub:
## Summary

Fixes #930, Fixes #845

On the new core, backend configuration was never written back:
`CoreConfig.get_backend_config()` (the only writable handle) had a
single caller, the *reader* at startup. Adding a backend went through
`ds.register_backend()` + `ds.save()` (which only saves task data), so
`backends.conf` stayed empty and every backend vanished on the next
launch. This also subsumes the original 0.6 report (#845): sections
used to be keyed by module name, so a second backend of the same type
overwrote the first.

## Design

The Datastore already owns backend lifecycle, so it now also owns their
persistence. A single documented entry point,
`Datastore.save_backend_config(backend)`, serializes every static
parameter via the existing `cast_param_type_to_string()` (passwords go
to the keyring, as before), one section keyed by **pid** — which fixes
the #845 collision — and drops any leftover module-named legacy section
so a backend isn't loaded twice after upgrading.

It is called at the four mutations the Datastore already handles:

- **registration** — only on a genuinely new backend (`first_run`),
  never when reloading from disk at startup;
- **removal** — the section is deleted;
- **state toggle & parameter edits** — via a `BACKEND_STATE_TOGGLED`
  listener. This is deliberate: disabling spins a quit thread that
  flips the state afterwards, so persisting synchronously would race
  it; the signal is marshalled to the main loop (`idle_add`), so we
  save the settled state. The configuration panel commits parameter
  edits right before toggling, so the same listener captures those too.

The UI layer learns nothing new — it already commits and toggles at one
point.

## Tests

New `tests/core/test_backend_persistence.py` (registration persists,
state toggle persists fresh parameters, removal deletes the section,
legacy section cleanup), all red before / green after, using a fake
keyring so the real one is never touched.

Also, a suite-wide isolation guard in `tests/conftest.py`: since real
Datastores now persist config, the whole suite is pointed at a scratch
XDG dir so tests can never write into a contributor's real
`~/.config/gtg`. Full suite: 190 passed, 7 skipped locally.

## Follow-up (out of scope)

A "parameters changed" signal with a single listener would be the
tidiest long-term shape, but it's more than this fix needs today.

### DIFF
--- a/GTG/core/config.py
+++ b/GTG/core/config.py
@@ -254,3 +254,9 @@ class CoreConfig:
             self._backends_conf[backend],
             DEFAULTS['backend'],
             self.save_backends_config)
+
+    def delete_backend_config(self, backend):
+        """Remove the configuration section of a backend, if any."""
+        if backend in self._backends_conf:
+            self._backends_conf.remove_section(backend)
+            self.save_backends_config()
--- a/GTG/core/datastore.py
+++ b/GTG/core/datastore.py
@@ -34,6 +34,7 @@ from GTG.core import firstrun_tasks
 from GTG.core.dates import Date
 from GTG.backends.backend_signals import BackendSignals
 from GTG.backends.generic_backend import GenericBackend
+from GTG.core.config import CoreConfig
 import GTG.core.info as info
 import GTG.core.dirs as dirs
 import GTG.core.versioning as versioning
@@ -145,6 +146,9 @@ class Datastore:
         self._mutex = threading.Lock()
         self.backends: Dict[str,GenericBackend] = {}
         self._backend_signals = BackendSignals()
+        self._backend_signals.connect(
+            self._backend_signals.BACKEND_STATE_TOGGLED,
+            self._on_backend_state_toggled_persist)
 
         # When a backup has to be used, this will be filled with
         # info on the backup used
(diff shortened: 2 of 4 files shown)


diegogangl:
Works as expected, thanks!

cpitclaudel:
Probably makes https://github.com/getting-things-gnome/gtg/pull/1285 redundant?

pentux-GitHub:
I think that's a typo for another number — this PR is about backend
config persistence, not a sync crash. I had a look at your #1057
though and left an update there.

cpitclaudel:
Indeed a typo, sorry: I meant https://github.com/getting-things-gnome/gtg/pull/931

pentux-GitHub:
No worries! And your intuition is right: #931 is now effectively superseded by
this PR (together with the follow-up fix in #1293). Master indexes backend
config sections by pid, migrates the legacy module-named sections (the #845
double-load case), and there's a dedicated test for the `backend_caldav` → pid
migration in `tests/core/test_backend_persistence.py`.

Same goal, converging implementations — yours just sat in the queue too long. :)
Feel free to close #931, or I can ping Diego about it.

One caveat: only the config layer of #930 is done. Actually *running* two
backends of the same kind still needs work on the new core (UI to add a second
instance, the sync races Diego mentioned back then), so #930 itself probably
deserves to stay open.

PR #1286, opened 2026-07-11
fix: never let a plugin failure brick startup, verify declared DBus deps

### LINKED ISSUE #683: Activating the Hamster plugin without Hamster present (or missing permissions in Flatpak) will cause tracebacks on startup, crashing / preventing GTG's future launches with `DBus.Error.NameHasNoOwner` and `DBus.Error.ServiceUnknown`
Hi, I have installed GTG 0.5 via flapak, I am getting the following error the first time I run the app with `flatpak run org.gnome.GTG`

```python-traceback
  File "/app/lib/python3.8/site-packages/dbus/bus.py", line 175, in activate_name_owner
    return self.get_name_owner(bus_name)
  File "/app/lib/python3.8/site-packages/dbus/bus.py", line 359, in get_name_owner
    return self.call_blocking(BUS_DAEMON_NAME, BUS_DAEMON_PATH,
  File "/app/lib/python3.8/site-packages/dbus/connection.py", line 650, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NameHasNoOwner: org.freedesktop.DBus.Error.NameHasNoOwner
```

During handling of the above exception, another exception occurred:

```python-traceback
  File "/app/lib/python3.8/site-packages/GTG/gtk/application.py", line 123, in do_activate
    self.init_shared()
  File "/app/lib/python3.8/site-packages/GTG/gtk/application.py", line 162, in init_shared
    self.init_plugin_engine()
  File "/app/lib/python3.8/site-packages/GTG/gtk/application.py", line 188, in init_plugin_engine
    self.plugin_engine.activate_plugins()
  File "/app/lib/python3.8/site-packages/GTG/core/plugins/engine.py", line 201, in activate_plugins
    plugin.instance.activate(api)
  File "/app/lib/python3.8/site-packages/GTG/plugins/hamster/hamster.py", line 155, in activate
    self.hamster = dbus.SessionBus().get_object('org.gnome.Hamster',
  File "/app/lib/python3.8/site-packages/dbus/bus.py", line 239, in get_object
    return self.ProxyObjectClass(self, bus_name, object_path,
  File "/app/lib/python3.8/site-packages/dbus/proxies.py", line 248, in __init__
    self._named_service = conn.activate_name_owner(bus_name)
  File "/app/lib/python3.8/site-packages/dbus/bus.py", line 180, in activate_name_owner
    self.start_service_by_name(bus_name)
  File "/app/lib/python3.8/site-packages/dbus/bus.py", line 275, in start_service_by_name
    return (True, self.call_blocking(BUS_DAEMON_NAME, BUS_DAEMON_PATH,
  File "/app/lib/python3.8/site-packages/dbus/connection.py", line 650, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.ServiceUnknown: org.freedesktop.DBus.Error.ServiceUnknown
```

This happen the first time I run the app, if I use a different console to run the app it opens without the error. The same happen if I click in the app's icon, the first time gnome just wait for the app to open but no window shows up, if I click again in the icon then the window shows up properly. 

I have a GNOME Shell 3.38.4 on a Debian sid. Let me know if you need more info.

### LINKED ISSUE #998: Exception when adding a CalDAV provider: no attribute 'set_interaction_request'
Clicking the "Enable syncing" button on the CalDAV provider screen raises this exception:

**Context:** Global generic exception

```python-traceback
  File "/home/eloi/dev/gnome/gtg/.local_build/install/lib/python3.11/site-packages/GTG/gtk/browser/main_window.py", line 1554, in on_backend_needing_interaction
    infobar.set_interaction_request(description, interaction_type, callback)
AttributeError: 'NoneType' object has no attribute 'set_interaction_request'
```

**Software versions:**
* GTG (debug version) v0.6-367-g66f91627
* CPython 3.11.7 (main, Jan 29 2024, 16:03:57) [GCC 13.2.1 20230801]
* GTK 4.12.5, GLib 2.78.0
* PyGLib 3.46.0, PyGObject 3.46.0
* Linux-6.7.6-arch1-1-x86_64-with-glibc2.39

Here is the faulty line:
https://github.com/getting-things-gnome/gtg/blob/66f9162780c3f09b70ce6f26e0438606d6c751bb/GTG/gtk/browser/main_window.py#L1553-L1554

This seems to be due to the `_new_infobar` method returning a `None` value:

https://github.com/getting-things-gnome/gtg/blob/66f9162780c3f09b70ce6f26e0438606d6c751bb/GTG/gtk/browser/main_window.py#L1584-L1593

Indeed, the vbox_toolbars variable never gets another value assigned than `None`.

https://github.com/getting-things-gnome/gtg/blob/66f9162780c3f09b70ce6f26e0438606d6c751bb/GTG/gtk/browser/main_window.py#L86

### LINKED ISSUE #1248: Sorting order not restored on startup with 0.7 (take 3)
Testing git `master` now with `./launch.sh -s some_dataset`, I must sadly point out that #1020 (the successor to #353) is still not fixed, in any of the views.

When I open `gtg_dev_folder/tmp/the_dataset/xdg/config/gtg/gtg.conf` in gedit on the side, and then change the sort order with the menubutton in GTG's main window, I can see that it sets the value for the `sort_mode_open` variable in the gtg.conf file... however it does not seem to actually read/use/restore it on next startup: GTG remembers and restores which of the 3 views was used, but does not set the sorting order in the views / the sorting popovermenu.

I also noticed that it only saves that one variable in the file, it doesn't seem to save a variable for the sorting direction (ascending or descending)?

---

P.S.: I also saw that there were leftover variables like

```ini
tasklist_sort_column = 6
tasklist_sort_order = 1
```

…in that file. However, GTG doesn't remove / clean them if present, so I guess it ignores them too; if I delete those lines, they don't come back.

pentux-GitHub:
## Summary

Fixes #998, Fixes #683, Fixes #1248

Enabling the Hamster plugin without the Hamster service present could
make GTG unlaunchable: tracebacks at startup, and since the plugin
stayed enabled in the configuration, every subsequent launch died the
same way. Two mechanisms combined into that brick:

- `activate_plugins()` ran `plugin.instance.activate(api)` with no
  protection whatsoever, at startup (application.py), so any exception
  in any plugin killed the whole application, forever.
- The `dbus-dependencies` manifest field (which Hamster declares) was
  parsed into `Plugin.dbus_depends`... and never checked anywhere. The
  safety net was designed but never woven. Meanwhile
  `dbus.SessionBus().get_object()` can succeed lazily (Flatpak,
  orphaned service files), letting activation complete and the first
  real method call blow up later with ServiceUnknown /
  NameHasNoOwner — the exact errors in these reports.

## Fix (engine-level, protects every plugin, present and future)

- New `PluginEngine._check_dbus_depends()`: before activating, verify
  each declared DBus name with `name_has_owner()`. Missing (or no
  python-dbus, or no session bus — conservative on purpose) means the
  plugin is marked errored and skipped, with a log warning. The plugins
  dialog already knew how to display `missing_dbus` as (name, path)
  pairs; it had just never been fed.
- The activation loop itself is now guarded: a plugin raising during
  `activate()` gets logged (`log.exception`), marked errored and
  deactivated for the session, and GTG keeps starting.
- A refused or crashing plugin is also un-flagged as `enabled`, so the
  dialog tells the truth: the row shows unchecked and insensitive with
  the missing service named, instead of a frozen checked box. A stale
  `enabled` entry left in the configuration just produces one warning
  line at boot.

## Tests and verification

New `tests/core/test_plugin_engine_safety.py`: a raising plugin must
not propagate and must end up disabled, and a plugin with an
unsatisfied DBus dependency must be blocked before activation, with
`missing_dbus` in the exact tuple shape the dialog unpacks. Red on
master, green after.

Verified live on a machine with no Hamster at all: a dataset with the
plugin already enabled in its config — the exact brick scenario — now
boots normally, with one clean warning in the log, and the plugins
dialog shows the row unchecked with "please start
org.gnome.Hamster:/org/gnome/Hamster".

hamster.py itself still speaks the old core in places:
`on_task_modified` references an undefined `task_id` (NameError on
task completion), `get_id()` / `get_status()` / `STA_*` usages, and an
unguarded `GetTodaysFacts()` call. Happy to send a follow-up PR
porting the plugin internals if there's interest.

### DIFF
--- a/GTG/core/plugins/engine.py
+++ b/GTG/core/plugins/engine.py
@@ -192,26 +192,65 @@ class PluginEngine(Borg):
     def remove_api(self, api):
         self.plugin_apis.remove(api)
 
+    def _check_dbus_depends(self, plugin):
+        """Check the DBus services a plugin declared as dependencies.
+
+        The `dbus-dependencies` manifest field was parsed but never
+        verified anywhere, so a plugin talking to an absent service
+        (Hamster, typically) blew up at activation or later instead of
+        being cleanly marked unavailable (#998, #683, #1248).
+        Conservative on purpose: no python-dbus module or no session
+        bus counts as missing.
+        """
+        plugin.missing_dbus = []
+        for dep in plugin.dbus_depends:
+            # the plugins dialog displays these as (name, path) pairs
+            bus_name, _, obj_path = dep.partition(':')
+            try:
+                import dbus
+                owned = dbus.SessionBus().name_has_owner(bus_name)
+            except Exception:
+                owned = False
+            if not owned:
+                plugin.missing_dbus.append((bus_name, obj_path))
+                plugin.error = True
+
     def activate_plugins(self, plugins=[]):
-        """Activate plugins."""
-        assert(isinstance(plugins, list))
+        """Activate plugins, never letting one break startup."""
+        assert isinstance(plugins, list)
         if not plugins:
             plugins = self.get_plugins("inactive")
         for plugin in plugins:
             # activate enabled plugins without errors
             if plugin.enabled and not plugin.error:
+                self._check_dbus_depends(plugin)
+                if plugin.error:
+                    plugin.enabled = False
+                    log.warning(
+                        "Not activating plugin %s, missing DBus "
+                        "service(s): %s", plugin.module_name,
+                        plugin.missing_dbus)
+                    continue
                 # activate the plugin
                 plugin.active = True
-                for api in self.plugin_apis:
-                    if hasattr(plugin.instance, "activate"):
-                        plugin.instance.activate(api)
-                    if api.is_editor():
-                        if hasattr(plugin.instance, "onTaskOpened"):
-                            plugin.instance.onTaskOpened(api)
-                        # also refresh the content of the task
-                        tv = api.get_ui().get_textview()
-                        if tv:
-                            tv.on_modified(None)
+                try:
+                    for api in self.plugin_apis:
+                        if hasattr(plugin.instance, "activate"):
+                            plugin.instance.activate(api)
+                        if api.is_editor():
+                            if hasattr(plugin.instance, "onTaskOpened"):
+                                plugin.instance.onTaskOpened(api)
+                            # also refresh the content of the task
+                            tv = api.get_ui().get_textview()
+                            if tv:
+                                tv.on_modified(None)
+                except Exception:
+                    log.exception(
+                        "Plugin %s failed to activate, disabling it "
+                        "for this session:", plugin.module_name)
+                    plugin.error = True
+                    plugin.active = False
+                    plugin.enabled = False
 
     def deactivate_plugins(self, plugins=[]):
         """Deactivate plugins."""
--- a/tests/core/test_plugin_engine_safety.py
+++ b/tests/core/test_plugin_engine_safety.py
@@ -1,0 +1,55 @@
+from types import SimpleNamespace
+from unittest import TestCase
+
+from GTG.core.plugins.engine import PluginEngine
+
+
+class _RaisingInstance:
+    def activate(self, api):
+        raise RuntimeError('boom')
+
+
+class _RecordingInstance:
+    def __init__(self):
+        self.activated = False
+
+    def activate(self, api):
+        self.activated = True
+
+
+def _stub_plugin(instance, dbus_depends=None):
+    return SimpleNamespace(enabled=True, error=False, active=False,
+                           instance=instance, module_name='stub',
+                           dbus_depends=dbus_depends or [],
+                           missing_dbus=[])
+
+
+def _engine():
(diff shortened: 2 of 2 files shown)


pentux-GitHub:
Follow-up field check, both directions. Installed hamster-time-tracker
(3.0.3, Debian) and started it, then relaunched GTG on the same
dataset: the plugin row ungreys, activates cleanly, and its
preferences dialog opens fine. Then killed the Hamster process and
relaunched: the row greys back out, unchecked, with the "please
start..." hint, and the dataset's tasks are untouched. The guard
blocks when the service is absent, steps aside when it's there, and
steps back in when it disappears — round trip verified live.

diegogangl:
LGTM, thanks!

PR #1287, opened 2026-07-11
fix: port the Hamster plugin internals to the new core and GTK 4

pentux-GitHub:
## Summary

Follow-up to #1286, delivering the "noticed in passing" list there:
with the engine now refusing to activate Hamster when the service is
absent, this makes the plugin actually work when the service *is*
there. First full GTG↔Hamster loop verified on the new core.

## What was broken

The plugin still spoke the old core and GTK 3 throughout:

- `on_task_modified` referenced an undefined `task_id` (NameError the
  moment any task was completed), plus `get_id()` / `get_status()` /
  `STA_*` everywhere; helper.py used `get_title()`, `get_excerpt()`,
  `get_tags_name()`.
- The editor menu block built a `Gio.MenuItem` and then treated it
  like an old Gtk widget (`props.text`, `show_all()`,
  `connect('clicked')`); the change-label helpers indexed the stored
  item as a tuple (`[0]` TypeError).
- The headerbar button handler passed the clicked widget where the
  plugin api was expected (pre-existing), so toggling from the browser
  crashed as soon as the task's editor had been opened once. The
  relabel helpers also now only touch an editor's menu — called with
  the browser api they would silently append items to the burger menu.
- The records table used `Gtk.Viewport.add()` / `ShadowType`, both
  gone in GTK 4.
- `GetTodaysFacts()` was unguarded (the exact ServiceUnknown /
  NameHasNoOwner emitter from #998); `deactivate()` could pass a None
  widget id and never disconnected its datastore handlers; dead
  liblarch subscription code lingered.
- The selection callback still expected a `GtkTreeSelection`
  (`count_selected_rows()`).

## Bonus core fix

The plugin turned out to be the first real caller of
`PluginAPI.get_selected()` on the new core, and it was broken:
`MainWindow.get_selected_tasks()` still called the old tree's
`get_selected_nodes()`, which no longer exists. It now honors its
documented contract (a list of task ids) through the new pane's
`get_selection()`. Neighboring old-tree calls remain on the tag side
(~1446/1562/1748) — left for a follow-up.

## Tests and verification

New `tests/plugins/test_hamster_plugin.py`: FactBuilder builds a fact
from a real new-core Task, completing a task no longer NameErrors, and
`get_active_id()` survives a vanishing service. Red on master, green
after.

Verified live against hamster-time-tracker 3.0.3: headerbar button
enables on selection, starting from GTG creates the running activity
in Hamster, the records table renders in the task editor, toggling
works on a task whose editor was previously opened, and completing the
task stops the tracking automatically (with the historical one-minute
buffer). No tracebacks.

- The fact string sends empty fields as bare commas when description
  or tags are empty (cosmetic in Hamster's UI).
- The old-tree calls on the tag side of main_window, listed above.

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -1531,24 +1531,22 @@ class MainWindow(Gtk.ApplicationWindow):
             return None
 
     def get_selected_tasks(self, tree_view: str = ''):
-        """
-        Returns a list of 'uids' of the selected tasks, and the corresponding
-        iters
+        """Return the ids of the selected tasks, if any.
 
-        @param tree_view: The tree view to find the selected task in.
-                          Defaults to the task_tview.
+        @param tree_view: The pane to look in. Defaults to the
+                          currently selected one, falling back to the
+                          first pane with a selection.
         """
-
-        selected = []
         if tree_view:
-            selected = self.panes[tree_view].get_selected_nodes()
+            tasks = self.panes[tree_view].get_selection()
         else:
-            current_pane = self.get_selected_pane()
-            selected = self.panes[current_pane].get_selected_nodes()
-            for i in self.panes:
-                if not selected:
-                    selected = self.panes[i].get_selected_nodes()
-        return selected
+            tasks = self.panes[self.get_selected_pane()].get_selection()
+            if not tasks:
+                for pane in self.panes.values():
+                    tasks = pane.get_selection()
+                    if tasks:
+                        break
+        return [task.id for task in tasks]
 
     # If nospecial=True, only normal @tag are considered
     def get_selected_tags(self, nospecial=False):
--- a/GTG/plugins/hamster/hamster.py
+++ b/GTG/plugins/hamster/hamster.py
@@ -26,7 +26,7 @@ from gettext import gettext as _
 import dbus
 from gi.repository import Gtk, Gio
 
-from GTG.core.tasks import Task
+from GTG.core.tasks import Status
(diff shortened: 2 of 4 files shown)


pentux-GitHub:
One more field datapoint: on current master (while testing the
unrelated #1288 branch), simply selecting a task with the plugin
enabled crashes with AttributeError: 'list' object has no attribute
'count_selected_rows' (hamster.py:333, via the api's selection
callback). The selection_changed port in this PR fixes exactly that
path.

diegogangl:
This works correctly but I get a traceback when enabling:

```
  File "/home/januz/code/gtg/.local_build/install/lib/python3.14/site-packages/GTG/gtk/editor/taskview.py", line 206, in _masked_processing
    self.process(mask_current_word=buffer.get_modified())
```

pentux-GitHub:
Thanks for testing! Could you paste the full traceback? The snippet cuts off at the first frame (`taskview.py:206`, `_masked_processing`), so I can't see what actually raises.

Worth noting: that line is in the editor's text-processing path, not in anything this PR touches (the diff is `main_window.py`, `hamster.py`, `helper.py` + tests — `taskview.py` is untouched). So this may be a pre-existing editor issue that enabling the plugin happens to surface rather than something introduced here. The full stack will tell us which — I'll chase it down either way.

diegogangl:
I've rebased and tried again, but couldn't find the traceback

diegogangl:
LGTM, thanks!

PR #1284, opened 2026-07-10
fix: port the backend infobar to GTK 4 and the new core

pentux-GitHub:
## Summary

Fixes #784, Fixes #901

The backend infobar is GTG's historical way (since the 0.x era) of
telling the user that a synchronization backend failed or needs
attention. It never survived the GTK 4 / new core transition: the
container was restored in 1e93bbde, but the code paths behind it were
still speaking GTK 3 and the old core, so any backend error crashed
before anything was shown, leaving a frozen window and an orphan modal.
I hit this five times while testing #1265 — and it is what currently
blocks re-testing it (see the "blocked by #1241" note there; #1241's
content is already on master, this PR is the follow-up discussed
there).

## What was dead, now ported

- `main_window.py`: two `Gtk.Container.foreach()` calls (removed in
  GTK 4) replaced by a sibling-walking helper; `Box.add()` → `append()`;
  `BackendInfoBar(self.req, ...)` (old-core Requester) → the datastore.
- `backend_infobar.py`: constructor now takes the datastore route
  (`app.ds.get_backend()`); `get_content_area()` (gone in GTK 4) →
  `add_child()`; `Label.set_alignment()` → `set_xalign/set_yalign`;
  `show_all()` → `set_revealed(True)`; `hide()` on responses →
  `set_revealed(False)` (plain hide doesn't collapse a GTK 4 InfoBar);
  the textual-interaction dialog rebuilt without `Gtk.Alignment`
  (removed in GTK 4) and shown with `present()`.

Behavior is deliberately preserved as-is (this is a resurrection, not a
redesign).

## Tests and verification

New `tests/gtk/test_backend_infobar.py`: builds the error banner and
the interaction banner against a fake datastore, and checks that
responding collapses the bar. Red on master (TypeError on the old-core
constructor), green after. The tests skip cleanly when no display is
available, and pass under xvfb.

Verified live with a CalDAV backend armed with a wrong password: the
banner now appears instead of the historical freeze, the window stays
responsive, and OK collapses the banner.

## Honest notes

- `Gtk.InfoBar` is deprecated since GTK 4.10 (the test run shows the
  warnings). GTG doesn't depend on libadwaita, so migrating this to
  `Adw.Banner` (or a small custom revealer) is left as a maintainer
  decision / follow-up, together with possible UX simplifications
  (the CalDAV backend already uses the info-only INTERACTION_INFORM
  path, whose message text embeds a raw `repr()` worth cleaning).
[...]

PR #1288, opened 2026-07-11
chore: remove the dead old-tree tag selection code from main_window

pentux-GitHub:
## Summary

Follow-up to the note in #1287: main_window still carried the whole
GTK3-era tag-tree selection layer. `self.tagtreeview` was set to None
at init and never assigned again, so all of it was either unreachable
(handlers never connected, `self.tagpopup` never even created) or
silently no-oping behind `if self.tagtreeview:` guards. This removes
265 lines of dead code from the biggest file of the repo.

## Removed

- The old tag-treeview right-click and key handlers (never connected).
- `on_delete_tag`, `on_select_tag`, `get_selected_tags`,
  `get_selected_search`, `find_value_in_treestore`, `reset_cursor`,
  `set_target_cursor`, `has_any_selection`, `get_selected_tree`,
  `close_all_task_editors`: no callers anywhere in the repo.
- The `tagtree` / `tagtreeview` / `tag_active` init attributes.

## Kept / adjusted

- `on_delete_tag_activate` stays (the sidebar context menu calls it,
  always passing the tags explicitly) but loses its dead fallback:
  the signature is now `(self, tags)`.
- `select_on_sidebar` had one live caller (`Application.select_tag`)
  and silently did nothing; it now actually works, in one line,
  through the modern sidebar's `select_tag()`.

## Verification

- Full test suite green (186 passed, 7 skipped), no new pycodestyle
  issue (differential check against master).
- Live smoke: tag filtering by clicking sidebar tags works, and
  right-click → Delete on a tag still targets the right tag.

## Left for later (still speak the old core)

- `have_same_parent` (called from the browser sensitivity code) still
  uses `self.req`, which doesn't exist.
- The backend infobar construction uses `self.req` too — that line is
  replaced by #1284.
- The collapsed-paths restore still references `vtree_panes`.

### DIFF
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -108,7 +108,6 @@ class MainWindow(Gtk.ApplicationWindow):
         # Object prime variables
         self.app = app
         self.config = app.config
-        self.tag_active = False
 
         self.sidebar = Sidebar(app, app.ds, self)
         self.sidebar_vbox.append(self.sidebar)
@@ -133,8 +132,6 @@ class MainWindow(Gtk.ApplicationWindow):
         self._set_actions()
 
         # Tags
-        self.tagtree = None
-        self.tagtreeview = None
 
         self.sidebar_vbox.connect('notify::visible', self._on_sidebar_visible)
         self.add_action(Gio.PropertyAction.new('sidebar', self.sidebar_vbox, 'visible'))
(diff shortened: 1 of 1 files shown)


pentux-GitHub:
Pushed a second commit extending the sweep to the task side: the four
old-tree task/closed-pane context-menu handlers (their connections
were already commented out), `have_same_parent` (only ever called from
one of them — so the "same parent" sensitivity check for `add_parent`
was already dead in practice, the modern context menu never ran it),
and the now-orphaned `show_popup_at` helpers. 87 more lines gone.
`self.req` now appears exactly once in the whole file, in the backend
infobar line that #1284 replaces.

diegogangl:
Nothing feels as good as removing old dead code. Thanks!

Issue #939, opened 2022-11-06
"End of life" in Pop Shop


BarbzYHOOL:
It says it's end of life in the software installer "pop shop", where can i read about it?

Neui:
Probably it wants to say the *runtime* used (Gnome Platform v41) is End of Life, not GTG itself.
Upgrading to a newer Gnome Platform runtime [currently doesn't work](https://github.com/getting-things-gnome/gtg/issues/846) because of an icon we are using being removed, and it seems [someone](https://github.com/flathub/org.gnome.GTG/pull/7) on the "flathub side" created a patch to fix that particular issue. It probably just needs some more testing to make sure no other error occurs, then maybe it could be merged (and GTG be updated to an non-EOL runtime version).

BarbzYHOOL:
so it will work if i install or not? looks rad

Neui:
It should work, assuming it is using Flatpak/Flathub.

BarbzYHOOL:
thx

sebasvila:
> Probably it wants to say the _runtime_ used (Gnome Platform v41) is End of Life, not GTG itself. Upgrading to a newer Gnome Platform runtime [currently doesn't work](https://github.com/getting-things-gnome/gtg/issues/846) because of an icon we are using being removed, and it seems [someone](https://github.com/flathub/org.gnome.GTG/pull/7) on the "flathub side" created a patch to fix that particular issue. It probably just needs some more testing to make sure no other error occurs, then maybe it could be merged (and GTG be updated to an non-EOL runtime version).

Just copying here messages when updating via flatpak in Debian:

```
Looking for updates…
Info: org.gnome.Platform//41 is end-of-life, with reason:
   The GNOME 41 runtime is no longer supported as of September 17, 2022. Please ask your application developer to migrate to a supported platform.
Applications using this runtime:
   org.gnome.GTG
Nothing to do.

```

nekohayo:
So far the runtime [hasn't](https://github.com/flathub/org.gnome.GTG/issues/5) been pulled from under our feet and will continue to work, so if we can get #737 completed and the codebase in a good enough shape to release version 0.7, it would be preferrable to do that with a fix included, because we don't really have the spare resources to be making parallel releases or testing GTG 0.6 with newer runtimes (we don't really know what issues—if any—might arise from a newer runtime...).

[...]

PR #1292, opened 2026-07-16
build: move the development Flatpak to the GNOME 50 runtime (#939)

pentux-GitHub:
Refs #939.

The development manifest targeted GNOME 48, which reached end of life
on 24 March 2026 — flatpak now prints the EOL warning when installing
the nightly. Flathub currently offers 49 and 50.

Moving to 50 needs exactly one extra thing: `six`. The Python modules
are installed offline (`--no-index --find-links`), and `six` is a
dependency of `python-dateutil` that used to come from the runtime;
under 50 (Python 3.13) it doesn't, so `python3-caldav` fails to build.
Adding the wheel is the whole fix.

Built and ran locally against runtime 50: the app starts, icons are
fine, and the CalDAV backend syncs against a live Nextcloud. Details
and the wider runtime picture are in #939.

Not touched here, but worth knowing: `.github/workflows/weekly_flatpak.yml`
still uses the `gnome-46` builder image while the manifest asks for a
different runtime. It seems to work (flatpak-builder pulls what the
manifest declares), so I left it alone rather than guess — happy to
follow up if you'd rather align it.

### DIFF
--- a/.github/workflows/weekly_flatpak.yml
+++ b/.github/workflows/weekly_flatpak.yml
@@ -12,7 +12,7 @@ jobs:
     # again
     runs-on: ubuntu-24.04
     container:
-      image: bilelmoussaoui/flatpak-github-actions:gnome-46
+      image: ghcr.io/flathub-infra/flatpak-github-actions:gnome-50
       options: --privileged
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


diegogangl:
Yes, please let's align the action too

pentux-GitHub:
Done: the weekly job now uses `ghcr.io/flathub-infra/flatpak-github-actions:gnome-50` (image published this week; also moved from the legacy Docker Hub name to the ghcr.io path the project documents).

Verified by a manual run in my fork: https://github.com/pentux-GitHub/gtg/actions/runs/29684930247 — build succeeded in 5m29s and produced the bundle.

Note the workflow itself is cron-only from the default branch, so upstream it will first run on the Friday after merge.

diegogangl:
LGTM, thanks!

PR #1293, opened 2026-07-16
fix: save the parameters a backend has, not the ones it declares

pentux-GitHub:
Fixing my own #1285.

It saves the parameters a backend **declares**, while
`get_saved_backends_list()` only provides the ones it managed to
**read** — it deliberately drops the rest (`# Parameter not found in
config`, there since 2020). The rest of the code copes with that: the
parameter panels check `if parameter_name in backend_parameters`
before reading, and the caldav backend falls back on its own default
(`.get('default-calendar', 'gtg')`). My function is the only place
that demands otherwise.

**The trigger is mundane**: the day a parameter is added to a backend,
every config already written lacks it. That happened in February with
`default-calendar`. The caldav backend has shipped since 0.6, so those
configs are real, and nothing had rewritten them in between — which
was precisely the bug #1285 fixed.

**The damage isn't the crash.** `SectionConfig.set()` writes to disk
on every call. So the `KeyError` lands after eleven writes and
**before** the line that drops the legacy section. The config ends up
with two sections for the same backend, and the next startup **loads
that backend twice**.

**The fix separates the three things the function was conflating.**

1. *Migrate first*: carry the old section over verbatim, without
   interpreting it. A keyring reference is useless to us when the
   keyring is down, but it belongs to the user and dropping it would
   lose it for good.
2. *Then persist*: write what the backend actually holds — which is
   how 0.6 did it (`for key, value in b.get_parameters().items()`)
   before that code was removed in 2022. Anything we don't write keeps
   the value carried over above.
3. *Clean up last*.

Checked against: a pre-February config, an unavailable keyring, two
backends of the same type (#845 still holds, no bleed between them),
and a config already duplicated by the current code — that one gets
repaired on the way through.

228 tests passing.

### DIFF
--- a/GTG/core/config.py
+++ b/GTG/core/config.py
@@ -255,6 +255,16 @@ class CoreConfig:
             DEFAULTS['backend'],
             self.save_backends_config)
 
+    def get_backend_raw_items(self, backend):
+        """Key/value pairs of a backend section, exactly as stored.
+
+        No typing, no defaults, no interpretation: this is for moving a
+        section without needing to understand it.
+        """
+        if backend not in self._backends_conf:
+            return []
+        return list(self._backends_conf[backend].items())
+
     def delete_backend_config(self, backend):
         """Remove the configuration section of a backend, if any."""
         if backend in self._backends_conf:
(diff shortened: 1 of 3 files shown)


diegogangl:
LGTM, thanks!

Issue #929, opened 2022-09-19
Manual synchronization


cpitclaudel:
Currently it seems that the CalDAV service only supports periodic synchronization.  It would be very nice to have a button to trigger immediate synchronization without waiting for the next sync.

Tagging @jaesivsm for guidance on the implementation.

MagicFab:
As a new user I found it a bit strange not to have *any* feedback on synchronization. Once it worked I kind of trusted it - but it took some time. 
I'd refer anyone willing to tackle this to try Joplin and observe their sync feedback, here's an example:

![image](https://github.com/getting-things-gnome/gtg/assets/317514/0a77ab74-d327-42f4-b0b9-f55625e9778a)

This is always visible on the bottom left part of the main window. That huge button does what it says.

jaesivsm:
While technically possible (the "sync" method call is pretty straightforward) in the current GTG implementation, backends sync are triggered by a mechanism independent from the actual plug. The mechanism kinda "oversee" the run of the sync process (technically it's run in a thread as far as I remember). So a better way to do that would be to have said button just talking to that mechanism and triggering an early sync.
So it'd be pretty independent from the plug implementation (and kinda out of my reach right now).

PR #1294, opened 2026-07-18
Add a manual "Synchronize now" button for remote backends (#929)

pentux-GitHub:
## What this does

Adds a "Synchronize now" button to the main window headerbar, visible only when a remote backend is enabled. Clicking it (or pressing F5) triggers an immediate synchronization cycle instead of waiting for the next periodic import. While any sync runs, the button turns into a spinner and is disabled, so it also acts as a sync activity indicator.

## History of the request

@cpitclaudel opened #929 in 2022: the CalDAV service only syncs periodically, and there was no way to trigger it on demand. @MagicFab later widened the point: as a new user he had no feedback at all that sync was happening, and pointed to Joplin's always-visible sync button as the model. @jaesivsm gave the implementation direction: don't touch the plug itself, have the button talk to the mechanism that oversees sync runs. This PR follows that direction.

## Implementation notes

Everything reuses existing machinery: `PeriodicImportBackend.start_get_tasks()` already cancels the pending timer, runs an import and reschedules (with reentrance coalesced by its urgent-iteration logic), `flush_all_tasks()` pushes local changes, and the `backend-sync-started`/`backend-sync-ended` signals were already emitted on every cycle but never consumed by the UI. The new `Datastore.sync_now()` mirrors the existing backend startup convention. ~100 insertions, no deletions, no new dependencies.

## Testing

Manually tested against a Nextcloud CalDAV server: immediate pull of server-side changes, spinner on manual and periodic cycles, button hidden when no remote backend is enabled. Note: functional testing was done with #1265 applied, since the CalDAV backend on current master crashes during import (the very issue #1265 fixes). The patch itself applies cleanly on master and is backend-agnostic.

## Open question for review

There is deliberately no menu entry, to keep the main menu lean; the shortcut is discoverable via the button tooltip. Happy to add a "Synchronize Now" item next to "Synchronization" if you prefer it there too.

### DIFF
--- a/GTG/core/datastore.py
+++ b/GTG/core/datastore.py
@@ -709,6 +709,34 @@ class Datastore:
         thread.start()
 
 
+    def sync_now(self, backend_id=None):
+        """
+        Trigger an immediate synchronization cycle on remote backends,
+        without waiting for the next periodic import.
+
+        @param backend_id: if given, only that backend is synced.
+        """
+
+        for backend in self.backends.values():
+            if backend_id and backend.get_id() != backend_id:
+                continue
+            if not backend.is_enabled():
+                continue
+            # Only remote (periodic import) backends can be forced.
+            # The local file backend syncs continuously and is skipped.
+            if not hasattr(backend, 'do_periodic_import'):
+                continue
+
+            def __sync_now(self, backend):
+                backend.start_get_tasks()
+                self.flush_all_tasks(backend.get_id())
+
+            thread = threading.Thread(target=__sync_now,
+                                      args=(self, backend))
+            thread.setDaemon(True)
+            thread.start()
+
+
     def set_backend_enabled(self, backend_id, state):
         """
         The backend corresponding to backend_id is enabled or disabled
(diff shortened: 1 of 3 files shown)


diegogangl:
Works as expected and it's a great idea. One small change is that I would use `update-symbolic` for the idle icon. 
I think it's more readable for users than `dirsync-symbolic` which has a folder and is more suited for file management.

pentux-GitHub:
Ok. Good to me. Nice.

I went looking for `update-symbolic` to apply this, but it doesn't exist
under that name in Adwaita (stable or git) nor in the icon-development-kit
/ Icon Library. Closest real candidates: `arrows-circle` from the dev kit
(would need bundling in `data/icons/`), or Adwaita's stock
`view-refresh-symbolic` (nothing to bundle). Which one did you have in
mind? I'll push it right away.

diegogangl:
That's weird, I found that icon name in the icon library app. Anyways, `view-refresh-symbolic` is ok.

pentux-GitHub:
Done — switched the idle icon to `view-refresh-symbolic` as you suggested (`8c238893`). Thanks for the review!

diegogangl:
Thanks!

PR #1297, opened 2026-07-21
fix(caldav): push local task deletions to the server

pentux-GitHub:
Follow-up to #1265, closing its explicit scope note: *"this PR does not
cover pushing local deletions to the server"*. Deleting a task in GTG
currently never deletes it on the CalDAV server, so it reappears on the
next import. This is not a regression (master never pushed deletions),
but it is the last missing piece of two-way sync.

## What was missing

The deletion plumbing already existed and worked — `queue_remove_task`,
the worker draining `to_remove` (with remove taking priority over set),
and `base_store.remove()` cascading one `removed` signal per descendant.
Only two wires were missing:

1. **Core** — nothing connected the TaskStore `removed` signal to
   `queue_remove_task` (the symmetric of what #1265 did for
   `queue_set_task`). *(commit 1)*
2. **Backend** — the public `remove_task` was inverted: it deleted the
   task from GTG's *own store* (which is where the signal came from in
   the first place) instead of deleting the remote todo. The real DAV
   deletion, `_remove_task`, was dead code with no caller. It now
   delegates to it, normalizing the tid with `str()` since the todo
   cache is keyed by `str(task.id)` while the queue hands over `UUID`
   objects (regression-tested). *(commit 2)*

Commit 3 makes the deletion idempotent: a todo already gone server-side
(404/NotFound) is a success — the goal, that it no longer exists, is
reached — logged as info instead of an exception.

## What was deliberately *not* reimplemented

`base_store.remove()` already cascades `removed` over the whole subtree
(children first), so deleting a parent naturally yields one
`queue_remove_task` — and one DELETE — per descendant. No tree walking
was added in the backend; a test pins this behavior instead.

## Tests

- `tests/core/test_datastore_removal_sync.py` (new): a removal on a
  Datastore with an enabled backend queues `queue_remove_task(task.id)`;
  a parent with 2 subtasks queues all three (children first); a disabled
  backend is not notified.
- `tests/backend/backend_caldav_test.py`, new `LocalDeletionPushTest`
  class written against the new core API (same style as
  `NonUuidUidRegressionTest` from #1265): the right `todo.delete()` is
  issued; the local store is *not* touched (regression on the inverted
  behavior); a `UUID` object tid still finds the todo (regression on the
  str-keyed cache); a task imported through `uid_to_task_id` (non-UUID
  server UID) is deleted; an unknown tid is a no-op; a call before the
  cache is ready is ignored; a 404 on delete is a success with no
  error-level log.
- Each commit is green in isolation. The pre-existing `CalDAVTest` class
  remains `@unittest.skip` (legacy old-core API); migrating it is a
  separate, mechanical follow-up I'm happy to take next.
- Live testing (Nextcloud ✅): created a parent with two subtasks,
  synced, verified server-side; deleted the parent in GTG → three
  DELETEs issued, children first, parent last; all three todos gone
  server-side; two full import cycles later, zero `created` — the tasks
  do not come back. The original symptom is gone.

## Known limitations, and where the answers live

- **The `to_remove` queue is in-memory** (same as `to_set`): a deletion
  made while the server is unreachable, or pending when GTG quits, is
  lost and the task will come back on import. A persistent
  deletion-journal replayed at startup would be the natural follow-up —
  a separate piece of work, same design for both queues.
- **Flush-at-quit is fragile**: `GenericBackend.quit()` runs
  `Thread(target=self.sync).run()`, which executes synchronously in the
  calling thread instead of spawning one. It happens to work today; a
  one-line follow-up (`self.sync()` directly, or `.start()`+`join()`)
  would make the shutdown flush deliberate rather than accidental.
- **Beat with the periodic import**: if an import runs while a DELETE is
  still queued, the todo can be re-imported just before being deleted,
  then deleted again next cycle — correct convergence, one cycle of
  flapping. The datastore mutex and the `to_remove > to_set` priority
  already bound this; documenting it beats over-engineering it. (Note
  the pending #1294 "Synchronize now" button currently triggers two
  imports per click, which would make this beat easier to observe
  manually — not a regression of this PR.)

Also closes #1240, made obsolete by #1265 (the added-date guard now
lives in `fill_task` on the new API).

### DIFF
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -158,15 +158,19 @@ class Backend(PeriodicImportBackend):
         self._set_task(task)
 
     @interruptible
-    def remove_task(self, tid: str) -> None:
+    def remove_task(self, tid) -> None:
+        """Push a local deletion to the server. The task is already
+        gone from the local store (that removal is what triggered the
+        'removed' signal), so this must only delete the remote todo."""
         if self._parameters["is-first-run"] or not self._cache.initialized:
-            logger.warning("not loaded yet, ignoring set_task")
+            logger.warning("not loaded yet, ignoring remove_task")
             return
         if not tid:
             logger.warning("no task id passed to remove_task call, ignoring")
             return
-        with self.datastore.mutex:
-            return self.datastore.tasks.remove(UUID(tid) if isinstance(tid, str) else tid)
+        # tid may arrive as a UUID object (task.id) from the queue;
+        # the todo cache is keyed by str(task.id), so normalize.
+        self._remove_task(str(tid))
 
     #
     # real main methods
@@ -222,11 +226,13 @@ class Backend(PeriodicImportBackend):
             self._create_todo(task, calendar)
 
     def _remove_task(self, tid: str) -> None:
-        todo = self._cache.get_todo(tid)
+        todo = self._cache.get_todo(str(tid))
         if todo:
-            self._remove_todo(tid, todo)
+            self._remove_todo(str(tid), todo)
         else:
-            logger.error("Could not find todo for task(%s)", tid)
+            # Not an error: the task may never have reached the server
+            # or may belong to another calendar. Nothing to delete.
+            logger.info("No known todo for task(%s), nothing to remove", tid)
 
     #
     # Dav functions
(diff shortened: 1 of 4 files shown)


diegogangl:
Works as advertised, thanks!

PR #1300, opened 2026-07-21
fix(core): load 0.6 data files whose task ids are not UUIDs

### LINKED ISSUE #1289: Non-UUID task ids in 0.6 data files abort the whole startup (real-world CalDAV-synced databases)
## Summary

`TaskStore.from_xml()` casts every task id with `UUID(element.get('id'))`
and no guard (GTG/core/tasks.py:897). A single task whose id is not a
canonical UUID raises `ValueError: badly formed hexadecimal UUID string`,
the exception propagates out of `find_and_load_file()`, and GTG fails to
start entirely. One malformed id makes the other N healthy tasks
unreachable.

Found while running current master against a real, long-lived 0.6
database (266 tasks): it contains 7 tasks whose ids are not UUIDs, and
startup aborts on the first one.

## Why such ids exist in legitimate 0.6 data

RFC 5545 defines the iCalendar UID as an opaque string; `name@host` is
even the traditional form. The 0.6 CalDAV backend used the remote UID
as the local task id, so any 0.6 database that ever synced tasks
created by another CalDAV client (Nextcloud Tasks, jtx Board, custom
clients...) can contain ids like:

    <task id="task-1772277627342-s60kypqm@someclient" ...>
    <task id="3046470725646703742" ...>

These files load fine in 0.6 and brick 0.7 as it stands. This affects
exactly the users a release wants to keep: people upgrading with years
of synced data.

## Reproduction

Take any 0.6 `gtg_data.xml`, change one task id to `test@example`,
launch. GTG 0.7 shows the error dialog and never opens; nothing loads.

## Prior art in-tree

The old-format converter already handles this exact situation:
`versioning.py:166` does `task.get('uuid') or str(uuid4())`. The
project has therefore already accepted the principle of regenerating an
id instead of failing. Only the direct 0.6 read path lacks it.

## Suggested direction (happy to submit a PR if this sounds right)

In `from_xml`, when `UUID(id)` fails, derive a deterministic UUID from
the original string (e.g. `uuid5(NAMESPACE, old_id)`) instead of
raising. Deterministic mapping means subtask/parent references written
with the same old string keep pointing to the same task, so hierarchies
survive without a remapping pass. Log one warning per converted id.
The original UID probably needs to be preserved (task attribute) so
CalDAV sync identity is not broken; that touches the same area as
#1265, so I'd rather align with reviewers before writing code.

No new feature, no behavior change for databases that already use
canonical UUIDs. Just making the 0.6 → 0.7 upgrade unable to brick on
standards-legal data.

## Noticed in passing

When startup fails this early, the exception dialog path itself crashes
(`application.py:216`, `'NoneType' object has no attribute
'set_application'`), so the user gets a stack trace instead of the
error dialog. Separate small bug, mentioned for completeness.

pentux-GitHub:
Fixes #1289.

`from_xml` cast every task id with `UUID()` and no guard. A 0.6 data
file that ever synced tasks created by another CalDAV client can carry
ids that are not canonical UUIDs — the old CalDAV backend used the
remote iCalendar UID verbatim as the local id, and RFC 5545 makes that
UID an opaque string (`name@host` is even the traditional form). A
single such id raised `ValueError`, aborted `find_and_load_file`, and
GTG failed to start entirely: one bad id made every other task
unreachable. This hits exactly the users a release wants to keep —
people upgrading with years of synced data.

Any non-canonical id is now mapped to a deterministic UUID (uuid5) at
all three conversion points in `from_xml` (the task id, the parent id,
and each `<sub>` reference). Deterministic mapping means references
written with the same original string keep pointing to the same task,
so hierarchies survive without a remapping pass. This mirrors
`uid_to_task_id` in the CalDAV backend (#1265), kept as a separate
helper so the core doesn't depend on a backend. Canonical UUIDs are
used as-is, so nothing changes for files that already use them; one
warning is logged per converted id.

Verified against a real reproduction: a `gtg_data.xml` carrying a
Nextcloud-Deck-style id (`deck-card-986`) crashes GTG at startup on
`master` (`ValueError` in `from_xml`) and loads cleanly on this branch.
Red-before/green-after unit tests cover the crash, the deterministic
mapping, the parent/child link survival, and that canonical UUIDs are
left untouched.

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -22,7 +22,7 @@
 from gi.repository import GObject, Gio, Gtk, Gdk # type: ignore[import-untyped]
 from gettext import gettext as _
 
-from uuid import uuid4, UUID
+from uuid import uuid4, uuid5, UUID, NAMESPACE_URL
 import logging
 from typing import Callable, Any, List, Optional, Set, Dict, Tuple, Union
 from enum import Enum
@@ -37,6 +37,27 @@ from GTG.core.tags import Tag, TagStore
 from GTG.core.dates import Date
 
 log = logging.getLogger(__name__)
+
+
+# Task ids are UUIDs, but 0.6 data files can carry non-canonical ids:
+# the old CalDAV backend used the remote iCalendar UID verbatim as the
+# local id, and RFC 5545 makes that UID an opaque string (name@host is
+# even the traditional form). Map anything that is not already a valid
+# UUID to a stable name-based one (RFC 4122, uuid5), so the same
+# original string always yields the same task id and subtask/parent
+# references written with that string keep pointing to the same task.
+# This mirrors uid_to_task_id in the CalDAV backend, kept separate to
+# avoid the core depending on a backend.
+ID_NAMESPACE = uuid5(NAMESPACE_URL, 'gtg://core/tasks')
+
+
+def id_to_uuid(task_id) -> UUID:
+    if isinstance(task_id, UUID):
+        return task_id
+    try:
+        return UUID(task_id)
+    except (TypeError, AttributeError, ValueError):
+        return uuid5(ID_NAMESPACE, str(task_id))
 
 
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


diegogangl:
LGTM, thanks!

PR #1301, opened 2026-07-21
fix(editor): dismiss the date popover when a preset button is clicked

### LINKED ISSUE #403: Clicking the "Now", "Soon", "Someday" buttons in the date picker widgets should dismiss the date pickers

pentux-GitHub:
Fixes #403.

Completes the second half of #403.

The date popovers' Now / Soon / Someday and Clear buttons set the date
but left the popover open — the user had to click away to dismiss it.
Selecting a day in the calendar already closes it; the preset buttons
now do the same, calling `popdown()` on the popover once the date is
set. Fix is centralised in the two convergence methods
(`on_duedate_fuzzy` and `on_date_cleared`), covering every preset
button.

The Escape half of #403 (Escape closing the whole editor instead of
just the popover) landed earlier in #1276; this closes the remaining
half.

Verified by hand: on `master` the popover stays open after clicking a
preset; on this branch it dismisses, matching the calendar
day-selection behaviour.

### DIFF
--- a/GTG/gtk/editor/editor.py
+++ b/GTG/gtk/editor/editor.py
@@ -606,11 +606,29 @@ class TaskEditor(Gtk.Window):
         year, month, day = gtime.get_year(), gtime.get_month(), gtime.get_day_of_month()
         return datetime.date(year, month, day)
 
+    def _dismiss_date_popover(self, kind):
+        """Close the date popover for the given kind and hand focus back.
+
+        The popovers open on the entry's focus 'enter' event, so simply
+        calling popdown() leaves the focus inside the entry: clicking it
+        again fires no new 'enter' and the popover won't reopen (#403).
+        Moving focus to the text view restores that, and it mirrors the
+        calendar day-selection UX -- pick a date, popover closes, you're
+        back in the task.
+        """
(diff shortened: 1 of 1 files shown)


diegogangl:
This only works on due date popovers, it should also work for "Today" and "Clear" on the start date. Also once the popover is auto-dismissed it doesn't come up again unless you click on the other date input

pentux-GitHub:
Good catches, both fixed:

- **Coverage**: now dismisses on every path that commits a date, for all three pickers (start, due, closed) — calendar day-selection, the Today / Now / Soon / Someday presets, and Clear. Not just due-date presets.
- **The reopen regression**: it happened because the popovers open on the entry's focus `enter` event, so `popdown()` while focus stayed in the entry meant re-clicking fired no new `enter`. The shared `_dismiss_date_popover` helper now hands focus back to the text view, which restores reopening.

Verified by hand on all three date fields. Thanks for the review!

diegogangl:
LGTM, thanks!

PR #1302, opened 2026-07-21
fix(core): marshal store signals to the main loop from worker threads

### LINKED ISSUE #1279: Backend sync thread mutates the task store without main-loop marshalling (GLib ref-count assertion, segfault)
While testing #1265 against a live Nextcloud, GTG occasionally crashed
hard during the first sync cycle:

    /usr/lib/python3/dist-packages/gi/overrides/Gio.py:138: Warning:
    g_atomic_ref_count_inc: assertion 'old_value > 0' failed
    Segmentation fault

Other runs of the same setup survive: it is a thread race, not a
deterministic bug, and it is independent of the #1265 changes (the
mutation pattern below predates them).

## Mechanism

- Backends are started in a worker thread
  (`Datastore.__backend_startup`, `datastore.py`).
- Backend *signals* are correctly marshalled to the main loop:
  `backend_signals.py` wraps every `emit` in `GLib.idle_add`.
- Backend *data mutations* are not: the sync thread calls
  `datastore.tasks.add()` / `remove()` / `parent()` directly. These
  stores are GObject-based and emit signals synchronously, so sidebar
  counters and list models react from the worker thread. GTK 4 is not
  thread-safe, hence the ref-count assertion and the occasional
  segfault, or a frozen UI.

## Historical corroboration

The original author of the CalDAV backend described the same beast in
#525 while rewriting it: "Some tasks created while a synchronization is
happening will often be considered as to be deleted and will be deleted
right after there creation. I've tried adding locks around potentially
conflicting operation but the problem, I think, is deeper than that".
The first-cycle "created N / deleted N" churn is still observable today.

## Scope

This affects any backend running on the new core, not just CalDAV.
Possible directions: marshal store mutations through `GLib.idle_add`,
or make the stores queue mutations for the main loop. Filing this
separately from #1265 as promised there; related to the #737 cleanup.

pentux-GitHub:
Backend sync runs in a worker thread and mutates the task store directly (add/remove/parent). Those stores are GObject-based and emit their signals synchronously, so sidebar counters and list models react on the spot -- from the worker thread. GTK 4 is not thread-safe, so that reaches GTK off the main loop and can raise
'g_atomic_ref_count_inc: assertion old_value > 0 failed' and segfault, or freeze the UI (#1279). Backend *signals* were already marshalled in backend_signals.py; the store's own *data* signals were not.

Route the four store emits (added, removed, parent-change, parent-removed) through a _emit helper that emits directly when called on the main thread -- preserving the synchronous behaviour existing callers rely on -- and defers via GLib.idle_add from any other thread. Fixing this in BaseStore covers every backend on the new core, not just CalDAV.

The race itself is intermittent and can't be pinned by a test; the added tests instead pin the marshalling contract (synchronous on the main thread, deferred off it).

Fixes #1279

### DIFF
--- a/GTG/core/base_store.py
+++ b/GTG/core/base_store.py
@@ -19,10 +19,11 @@
 """Base for all store classes."""
 
 
-from gi.repository import GObject # type: ignore[import-untyped]
+from gi.repository import GObject, GLib # type: ignore[import-untyped]
 
 from uuid import UUID
 import logging
+import threading
 
 from lxml.etree import _Element
 from typing import Dict, Optional, TypeVar, Generic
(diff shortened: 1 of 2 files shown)


diegogangl:
LGTM, thanks!

PR #1303, opened 2026-07-22
fix(caldav): refetch a missing todo by its server UID, not the GTG id

pentux-GitHub:
When an active task is absent from a fetch, the backend refetches it
from the server to decide whether it was really deleted there. It asked
with the task's GTG id (`str(task.id)`), but the server knows the task
by its own UID.

Those two only coincide for UUID-shaped server UIDs. `uid_to_task_id`
is deliberately one-way (uuid5), so for any server that issues
non-UUID UIDs — Nextcloud Deck, Evolution, `name@host`… — the GTG id
is a uuid5 the server never emitted. `todo_by_uid` then raises
`NotFoundError` every time, the task is treated as server-deleted, and
`datastore.tasks.remove()` runs: **the task is deleted on every
import.** Silent data loss, exactly on the users with a long
CalDAV-synced history.

The helper for this already exists — `remote_uid()`, which returns the
UID stored at import time and falls back on the id for GTG-created
tasks — it just wasn't wired in here.

**Verified live against a Nextcloud Deck calendar** (cards have
`deck-card-N` UIDs): importing the cards, then dropping one from the
next fetch while the server still holds it, deletes the card on
`master` and preserves it on this branch. Also covered by a
red-before/green-after unit test asserting the refetch asks for the
real UID and the task survives.

Found via a static audit of the CalDAV backend against the current
core; this was its highest-severity finding.

### DIFF
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -308,7 +308,7 @@ class Backend(PeriodicImportBackend):
                 logger.warning("Couldn't find calendar for %r", task)
                 return
             try:  # fetching missing todo from server
-                todo = calendar.todo_by_uid(uid)
+                todo = calendar.todo_by_uid(remote_uid(task, self.namespace))
             except caldav.lib.error.NotFoundError:
                 do_delete = True
             else:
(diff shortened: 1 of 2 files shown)


diegogangl:
LGTM, thanks!

PR #1304, opened 2026-07-22
WIP: fix FREQ→term parsing in the dormant CalDAV recurrence mapping (groundwork for #645)

pentux-GitHub:
**Groundwork for the recurrence discussion in #645. Not for merge as-is — help welcome.**

The CalDAV `Recurrence` field maps between iCalendar RRULE `FREQ` and GTG's recurring terms. It's currently commented out of the `Translator` field list, so it isn't active on the sync path yet.

The bug: `get_dav` derived the term with `freq.lower()[:-2]`, which works for `WEEKLY`→`week`, `MONTHLY`→`month`, `YEARLY`→`year`, but turns `DAILY` into `'dai'`, which GTG's date parser doesn't recognise. Replaced with an explicit `FREQ_TO_TERM` table that also ignores unsupported frequencies cleanly.

Red-before/green-after tests cover the round-trip. The wider design question (duplication-on-close vs a rule-based model) should be settled first — see #645, where I've laid out where GTG stands and what the field does now.

### DIFF
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -1015,6 +1015,9 @@ class OrderField(Field):
 
 class Recurrence(Field):
     DAV_DAYS = ['SU', 'MO', 'TU', 'WE', 'TH', 'FR', 'SA']
+    # iCalendar FREQ -> GTG recurring term (see dates.py parsing)
+    FREQ_TO_TERM = {'DAILY': 'day', 'WEEKLY': 'week',
+                    'MONTHLY': 'month', 'YEARLY': 'year'}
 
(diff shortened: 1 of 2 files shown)


pentux-GitHub:
Marking this ready for review following the discussion in #645.

Scope is deliberately minimal: it fixes the FREQ to term mapping in the CalDAV backend (a daily rule was read as an invalid term) and adds round-trip tests. The RRULE field stays commented out of the Translator list, so nothing changes on the sync path or the internal model. The recurrence redesign itself remains out of 0.7 scope as agreed.

diegogangl:
Thanks!

Issue #1305, opened 2026-07-22
CalDAV plugin: data loss in tag names after sync from Nextcloud


MagicFab:
I am using GTG (dev version) 0.7-dev-884baf58 synchronizing from NextCloud Hub 25 Autumn tasks (CalDAV), v.  to Thunderbird tasks (140.12.1esr in Ubuntu 26.04).

I implemented GTD with tags that all start with "@" - <span>@</span>1-Urgent, <span>@</span>2-Work, etc., as many online guides suggest to do.

This syncs well between Thunderbird tasks and Nextcloud tasks.

However when using GTG, the first sync from NextCloud tasks will change the tag names to "_1-Urgent" from the example above. "@" gets translated to "_" **and the new tag is assigned to the task, while replacing the original one**. This also removes the assigned color.

When syncing again in Thunderbird or looking at NextCloud tasks, they've all become "1-Urgent" from the example above ! The text in the task itself when examined from GTG is now showing "@_1-Urgent".

Something similar happens when using a single quote - "<span>@</span>7-N'importe où" becomes "7-N importe où" - why ?

I re-tagged ~25 tasks with multiple tags each and was able to reproduce this.

While I understand GTG uses "@" internally, it should support using it as it's a character suggested in the method's many guides and books.

I'll probably just abandon using "@" in any tag names but the problem (data loss) remains for tags with single quotes and probably other special characters in them.

A couple of references:

* https://www.todoist.com/productivity-methods/getting-things-done
* https://gettingthingsdone.com/wp-content/uploads/2021/05/GTD_Outlook_2019_SAMPLE_A4.pdf

ps. @nekohayo suggested mentioning @pentux-GitHub here as someone particularly active with contributions lately.

pentux-GitHub:
Thank you for such a precise report @MagicFab, the two examples made this quick to pin down. The data loss part is fixed in #1306; the `@` question I'd like to open up properly below, because it deserves a real answer rather than a workaround.

**What broke, and it was my doing.** In eec2faf (part of #1265) I made the CalDAV backend map every character outside `\w` and `-` onto `_`, to stop the editor truncating tags like `DAV_Deck:_Server` and forking a mangled duplicate on every open. But the reverse mapping turns every `_` back into a space, so the projection stopped being reversible: your `@` and your apostrophe collapsed into `_`, decoded as spaces, and got pushed back over the originals. A display fix that caused silent data loss. Sorry about the round of re-tagging that cost you.

[...]

PR #1306, opened 2026-07-23
fix(caldav): make the category-to-tag projection reversible again

pentux-GitHub:
Addresses the data loss in #1305 (leaves the issue open: one part of the report, the leading `@`, is a pre-existing limit that deserves its own discussion there).

**The bug.** Server categories synced from Nextcloud were mangled and written back mangled, overwriting the user's data: `@1-Urgent` came back as ` 1-Urgent`, `@7-N'importe où` as ` 7-N importe où`, colors detached from the renamed tags. Reported with real GTD data.

**The mechanism, and my own regression.** In eec2faf (part of #1265), to stop the editor from truncating tags like `DAV_Deck:_Server` and forking a mangled duplicate on every open, I mapped every character outside `\w` and `-` onto `_`. But the reverse mapping turns every `_` into a space, so the projection stopped being reversible: colons, quotes and `@` collapsed into `_`, decoded as spaces, pushed back to the server over the originals. It traded a display bug for silent data loss.

**The fix, two matched pieces.**
- *backend*: `to_tag()` only maps spaces to `_` again, the historical, exactly reversible contract. Every other character travels verbatim: what the server wrote is what the server gets back.
- *editor*: `insert_tags()` only writes a tag into the task text when the parser reads it back identically (new helper `tag_is_faithfully_representable`). Richer names keep living in the tag store and the editor's tag popover; they just stay out of the text. This contains the #1265 proliferation at its actual root, without touching the data.

The tag store itself accepts any name (names are XML attributes, tasks reference tags by UUID), so nothing is lost in between.

**Second commit: escaping tag names in the content regexes.** `remove_tag()` and `rename_tag()` splice the tag name straight into a regex, so `@x(y)` was never removed (the parentheses became a group) and `@a[b` raised `re.error`. The editor has always accepted those characters, so this predates the change above, but letting CalDAV names through verbatim makes the crash reachable from remote data, so it can't be left for later. Names are now `re.escape()`d, and the trailing `\b(?!-)` becomes an explicit lookahead on the characters a tag may continue with (`\b` needs a word character, so it silently failed for names ending in `)`, `%` or `*`). Removing `@work` still leaves `@workshop` and `@work-item` alone.

**Red-before/green-after** on both commits. `CategoryRoundTripTest` reproduced the exact corruption from the report (`' 1-Urgent'`, `' 7-N importe ou'`, `'Deck  Server'`) and now passes; the #1265 test is rewritten for the new contract (the calendar tag keeps the calendar's exact name); the editor guard has its own tests in `tests/gtk/test_taskview_tag_guard.py`; and the regex escaping is covered in `tests/core/test_task.py`, including the non-regression case. Full suite: 276 passed, 1 xfailed (the leading-`@` target, kept as `expectedFailure` to document where the discussion in #1305 points).

**Known pre-existing limits, out of scope here and flagged in #1305:** a leading `@` collides with GTG's tag sigil and is dropped by the store (`@1-Urgent` → `1-Urgent`, body intact); a `_` typed by the user inside GTG still reaches the server as a space.

### DIFF
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -744,18 +744,20 @@ class PercentComplete(Field):
 
 class Categories(Field):
     CAT_SPACE = '_'
-    # GTG's tag parsers do not agree on what a tag may contain: the
-    # editor's TAG_REGEX and core.tasks' TAG_LINE_REGEX both stop at
-    # anything outside \w and '-', while core.tags accepts more. A
-    # calendar named "Deck: Server" used to produce the tag
-    # "DAV_Deck:_Server", which the editor re-read as "@DAV_Deck" and
-    # then re-added as a second, truncated tag on every open. Only
-    # build tags every parser agrees on.
-    TAG_UNSAFE = re.compile(r'[^\w\-]', re.UNICODE)
+    # A server category may carry characters GTG's inline tag marker
+    # cannot (colons, quotes, ...). The historical contract only maps
+    # spaces to '_' and back, which is exactly reversible, so the
+    # server data survives the round trip untouched. Mapping *every*
+    # unsafe character onto '_' (as a previous fix did) is not
+    # reversible: '_' comes back as a space and the original category
+    # is overwritten on the server (#1305). Keeping richer names out
+    # of the task text is the editor's job (insert_tags only writes
+    # tags its parser reads back identically); the tag store itself
+    # accepts any name.
 
     @classmethod
     def to_tag(cls, category, prefix=''):
-        return f"{prefix}{cls.TAG_UNSAFE.sub(cls.CAT_SPACE, category)}"
+        return f"{prefix}{category.replace(' ', cls.CAT_SPACE)}"
 
     def get_gtg(self, task: Task, namespace: str = None) -> list:
         return [tag_name.lstrip('@').replace(self.CAT_SPACE, ' ')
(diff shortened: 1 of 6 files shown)


pentux-GitHub:
Verified live against a real Nextcloud instance, creating a temporary task carrying the categories from the report and reading back what GTG pushes:

| server category | on master | with this branch |
|---|---|---|
| `Deck: Server` | `Deck  Server` | `Deck: Server` |
| `@7-N'importe ou` | ` 7-N importe ou` | `7-N'importe ou` |
| `@1-Urgent` | ` 1-Urgent` | `1-Urgent` |
| `my first category` | `my first category` | `my first category` |

The server keeps every category verbatim, so the asymmetry is entirely ours. On master the apostrophe and the colon come back as spaces, overwriting the original; here they survive untouched, and the tags carry their real names (`@Deck:_Server`, `@7-N'importe_ou`) instead of mangled ones.

Checked in the running app too: those two tags no longer enter the task text (only `@1-Urgent`, `@DAV_NcGtgWeb` and `@my_first_category` are written), so the truncate-and-fork loop is gone, and they remain listed and ticked in the editor's tag popover, which reads from `task.tags` rather than the text.

One consequence worth flagging: a tag that stays out of the text currently can't be unticked either, because the `else` branch of `on_tag_toggled` was never implemented (`TODO(jakubbrindza)`). That gap predates this PR and affects every tag, but it becomes more visible for synced ones. Happy to do that in a follow-up PR if you'd like it.

The remaining `@1-Urgent` → `1-Urgent` is the pre-existing sigil collision, left open for the discussion in #1305.

diegogangl:
LGTM, thanks!

PR #1309, opened 2026-07-29
fix(core): load the 0.6 'Dismiss' status as DISMISSED

### LINKED ISSUE #1308: 0.6 data files use status "Dismiss", the 0.7 loader only knows "Dismissed": dismissed tasks silently come back as active
Found while test-driving a real 0.6 production file against current master (0e99f945). The 0.6 format stores dismissed tasks as `Dismiss` (this file: 114 Active, 152 Done, 8 Dismiss), but `TaskStore.from_xml` only matches `'Done'` and `'Dismissed'` (GTG/core/tasks.py:967-969). `Dismiss` falls through and loads as Active: all 8 dismissed tasks silently reappear in the active list (122 shown instead of 114). Affects any 0.6 user who ever dismissed a task, with no warning.

Related, possibly its own issue: `from_xml` assigns `task.status` directly instead of calling `set_status()`, so `is_active` keeps its default at load and the dimmed closed-task style is applied inconsistently.

The fix is a one-liner (accept `Dismiss` too); happy to send a small PR with a red/green test.

pentux-GitHub:
Fixes #1308.

A 0.6 data file stores dismissed tasks as `Dismiss`. The loader only matched `Dismissed`, so those tasks silently came back as active after migration (verified on a real 274-task production file: all 8 dismissed tasks resurrected). This accepts both spellings and sets `is_active` alongside, which also makes the dimmed closed-task style consistent. `set_status()` is deliberately not used in the load path, as it would overwrite `date_closed` and trigger recurrence duplication.

Red/green: the included test fails on master (`Dismiss` loads as `ACTIVE`) and passes with the fix. Full `tests/core/test_task.py` suite green (43 passed). Verified live against the production file: the 8 dismissed tasks are back in the Closed tab and the closed rows dim consistently.

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -964,10 +964,18 @@ class TaskStore(BaseStore[Task]):
                             'on the modification date', tid)
                 task.date_added = task.date_modified
 
+            # 0.6 wrote the dismissed status as 'Dismiss'; 0.7 renamed
+            # it to 'Dismissed'. Accept both so migrated files keep
+            # their dismissed tasks (#1308). Set is_active alongside:
+            # set_status() is avoided on purpose here, as it would
+            # overwrite date_closed with today and trigger the
+            # recurrence duplication logic.
             if status == 'Done':
                 task.status = Status.DONE
-            elif status == 'Dismissed':
+                task.is_active = False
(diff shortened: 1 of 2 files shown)


diegogangl:
LGTM, thanks!

Issue #1310, opened 2026-07-30
Add a toggle to hide tags that have no tasks in the current view


pentux-GitHub:
Since the GTK4 port the sidebar shows every tag in the store, including tags whose task count for the current view is zero. This was made consistent across view switches by commit bfcb78a8 which fixed #1007, and that commit introduced a show_zero field in TagEmptyFilter for this purpose. The field is currently hardcoded to True and nothing in the code changes it.

In 0.6 the behavior was the opposite: the sidebar view applied the activetag filter, which calls Tag.is_actively_used(), so a tag was only visible when it had at least one active task.

Both behaviors are legitimate. Showing everything gives access to historical tags, hiding empty ones keeps the sidebar focused. But the change is silent for people migrating from 0.6, and it can be dramatic on old datasets. Testing a 0.7 migration with a real six-year-old file, my sidebar roughly doubled: 19 legitimate tags only referenced by Done or Dismissed tasks became visible, one of them carrying 111 closed tasks. On files still containing keystroke-fragment tags from #605 the effect is much worse, since every fragment shows up at once.

Since TagEmptyFilter already has the show_zero field and per-pane counting logic, would you accept a small toggle, for instance a check item in the main menu or a preference, wiring show_zero to a stored setting? Defaulting to the current behavior would change nothing for existing 0.7 users, and a line in the release notes would warn 0.6 migrants. I can prepare the patch if the idea is accepted.

nekohayo:
This was fixed by #1311

PR #1311, opened 2026-07-30
feat(browser): add a 'Show Empty Tags' toggle in the main menu

### LINKED ISSUE #1007: Tags with zero issues are not shown on startup, but get shown after switching "Actionable" and "Open" view modes

pentux-GitHub:
This implements the proposal from #1310.

Since the GTK4 port the sidebar shows every tag in the store, including tags whose task count for the current pane is zero. Commit bfcb78a8, which fixed #1007, introduced the show_zero field in TagEmptyFilter for exactly this purpose, but the field is hardcoded to True and nothing in the code changes it. In 0.6 the sidebar applied the activetag filter through Tag.is_actively_used(), so tags without active tasks were hidden. Users migrating from 0.6 therefore see their sidebar grow considerably on old datasets, as described in the issue.

The change wires show_zero to a new stateful window action, exposed as a Show Empty Tags check item in the main menu next to Show Sidebar, and persisted in the browser configuration as show_empty_tags. The default is True, so the current 0.7 behavior is preserved exactly for anyone who does not touch the toggle. When unchecked, tags with no task in the current pane are hidden, matching the 0.6 behavior. The filter logic itself is untouched: the toggle only flips the existing field and notifies the filter.

Tested on a real six-year-old dataset migrated from 0.6 (54 tags, 276 tasks): the item appears in the menu, unchecking it hides the 19 tags that only have closed tasks, checking it brings them back, and the state survives a restart. The mechanism is also covered by a red and green check on TagEmptyFilter directly: a tag with zero open and five closed tasks matches by default, stops matching in the open view once show_zero is False, and still matches in the closed view.

Marked as draft pending feedback on #1310, in particular on the placement of the control, since a preference checkbox could be preferred over a menu item. If wanted I can add a line for the release notes.

### DIFF
--- a/GTG/core/config.py
+++ b/GTG/core/config.py
@@ -53,6 +53,7 @@ DEFAULTS = {
         'autoclean': True,
         'autoclean_days': 30,
         'dark_mode': False,
+        'show_empty_tags': True,
         'maximized': False,
         'sort_mode_open': 'title',
         'sort_mode_active': 'title',
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -348,6 +348,16 @@ class MainWindow(Gtk.ApplicationWindow):
         order_action.connect('change-state', self.on_sort_order)
         self.add_action(order_action)
 
+        show_empty = self.config.get('show_empty_tags')
+        empty_variant = GLib.Variant.new_boolean(show_empty)
+        empty_action = Gio.SimpleAction.new_stateful('show_empty_tags',
+                                                     None,
+                                                     empty_variant)
+
+        empty_action.connect('change-state', self.on_show_empty_tags)
+        self.add_action(empty_action)
+        self.sidebar.set_show_empty_tags(show_empty)
+
     def _init_icon_theme(self):
         """
         sets the deafault theme for icon and its directory
@@ -1134,6 +1144,16 @@ class MainWindow(Gtk.ApplicationWindow):
         else:
             self.get_pane().set_sort_order(reverse=True)
             self.change_sort_icon('DESC')
+
+
+    def on_show_empty_tags(self, action, value) -> None:
+        """Callback when toggling the display of tags without tasks."""
+
+        action.set_state(value)
+        show = value.get_boolean()
+        self.config.set('show_empty_tags', show)
+        self.sidebar.set_show_empty_tags(show)
+
 
     def store_sorting(self, mode: str) -> None:
         """Store sorting mode."""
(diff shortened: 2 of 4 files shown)


pentux-GitHub:
Updated after additional testing on the worst realistic case: the same six-year-old dataset before any cleanup, that is 270 tags of which 216 are keystroke fragments from #605, all with zero open tasks.

The toggle behaves as intended at this scale: unchecking collapses the sidebar from 270 tags to the active ones instantly, checking brings all 270 back, and the state survives restarts.

One issue was found and fixed along the way. Re-checking the item initially left an empty gap in the list where the filtered rows used to be. Measuring the model directly showed the cause: on this kind of bulk toggle the FilterListModel does not emit fine-grained changes, it emits a single huge splice, in my test replacing 53 visible rows with 269 in one items-changed signal. The final model content is correct, item for item, so the gap was purely a view allocation artifact: the ListView anchor was left pointing at recycled rows. This cannot happen on current master because show_zero is hardcoded to True and the filter never removes rows. The fix is a scroll_to(0) on the tag list view right after the refresh, re-anchoring the view at the top, with a comment explaining why. Filter change hints (MORE_STRICT, LESS_STRICT) were tested and produce the exact same splice, so they would not have helped.

Still draft pending feedback on #1310, in particular on whether a preference checkbox would be preferred over the menu item.

pentux-GitHub:
Marking this ready for review.

**What was verified locally** (large real-world dataset, 1771 tasks):
- The "Show Empty Tags" check item appears in the main menu, checked by default (preserving current 0.7 behavior)
- Unchecking it hides tags with no task in the current pane; the list re-anchors cleanly with no allocation gap
- Switching between Open / Actionable / Closed panes correctly re-applies the per-pane counts
- The state is persisted in the browser config and survives a restart

**Scope**: sidebar display only — this wires the existing `TagEmptyFilter.show_zero` field (introduced in bfcb78a8) to a stateful window action. No changes to tag storage, counters, or deletion.

Context: #1310 describes the underlying issue (0.6 hid empty tags, 0.7 shows them all, which can make the sidebar grow considerably on migrated datasets). This toggle gives users the 0.6-like view back as an opt-in, without changing the default.

diegogangl:
LGTM, thanks!

Issue #1153, opened 2024-11-09
Internal Error


jfoll31:
### **While simply editing a task I got this message**

  File "/app/lib/python3.9/site-packages/GTG/gtk/editor/taskview.py", line 200, in process
    if self.detect_subtasks(text, start):
  File "/app/lib/python3.9/site-packages/GTG/gtk/editor/taskview.py", line 302, in detect_subtasks
    parents = task.get_parents()
AttributeError: 'NoneType' object has no attribute 'get_parents'
```

**Software versions:**
* Getting Things GNOME! v0.6
* CPython 3.9.9 (main, Nov 10 2011, 15:00:00)   [GCC 11.3.0]
* GTK 3.24.34, GLib 2.70.5
* PyGLib 3.42.2, PyGObject 3.42.2
* Linux-6.8.0-48-generic-x86_64-with-glibc2.33

nekohayo:
This would need to have much more precise reproduction steps, because I haven't been encountering this issue (in years of using version 0.6) so it certainly does not affect every task. Ideally it should be tested against the git version, too.

nekohayo:
See also #924 in case this is the issue you are experiencing…

pentux-GitHub:
Housekeeping pass: same situation as #924, the get_parents API from this traceback is gone from the new core's UI code, and the requested reproduction details never came. I'd think it's safe to close as obsolete.

pentux-GitHub:
Same conclusion as #924 (linked by nekohayo above). Retested on current master after #1314 landed: editing tasks and creating/un-indenting subtasks no longer triggers the `get_parents` AttributeError — that code path was removed by the GTK4 rewrite. It seems safe to close as obsolete, with a fresh report welcome if it still reproduces on 0.7.

Issue #924, opened 2022-09-08
Issue Report "Internal error"


Disaster2life:
I was just creating some tasks and this popped up, and it now seems to do that everytime I press backspace? I haven't restarted GTG yet, but I might after this to see if it persists 
`**Context:**` Global generic exception

```python-traceback
Traceback (most recent call last):
  File "/app/lib/python3.9/site-packages/GTG/gtk/editor/taskview.py", line 200, in process
    if self.detect_subtasks(text, start):
  File "/app/lib/python3.9/site-packages/GTG/gtk/editor/taskview.py", line 302, in detect_subtasks
    parents = task.get_parents()
AttributeError: 'NoneType' object has no attribute 'get_parents'
```

**Software versions:**
* Getting Things GNOME! v0.6
* CPython 3.9.9 (main, Nov 10 2011, 15:00:00)   [GCC 11.3.0]
* GTK 3.24.34, GLib 2.70.5
* PyGLib 3.42.2, PyGObject 3.42.2
* Linux-5.19.6-200.fc36.x86_64-x86_64-with-glibc2.33`

Disaster2life:
Update, I think I fixed it without needing to restart, it seemed that having an open "-" for writing a task without anything next to it except the @ tags at the end of the task and that seemed to be conflicting

nekohayo:
Interestingly enough, I encountered the same sort of issue recently when GTG unexpectedly crashed (for the first time ever) while doing some subtask creation/typing in a task, and I don't know exactly how I triggered it. It somehow didn't end up in any major  problem, though I had to re-write my sub-tasks for it after restarting the app.

If you manage to figure out step-by-step reproduction instructions for this, this would be a good thing to test for when Diego's refactoring changes land in master eventually. Maybe this kind of torture case could even be part of a test suite somehow.

scovetta:
I'm not 100% sure this is the same bug, but it's on the same line, and here's a repro for v0.6:

1. Open GTG. Press [+] to create a new task.
2. Add a sub-task within the window.
3. Add another sub-task under the first one.
4. Indent the second sub-task so it's no longer highlighted.
5. Un-indent the second subtask.
6. Error is triggered.

Video is below:

https://user-images.githubusercontent.com/732166/204119944-45a8061a-39da-48d2-a4c4-a010a3e04fb6.mp4

pentux-GitHub:
[...]

PR #1314, opened 2026-08-02
fix(core): refuse a second parent claim instead of corrupting the tree

### LINKED ISSUE #1307: BaseStore.parent() silently corrupts the tree when a task is claimed as subtask by several parents
The "item not found in data" errors seen when loading the bryce dataset (see #109) reveal a store defect. Bryce has 36 tasks referenced as subtask by more than one parent. In `BaseStore.parent` the first call works, but any further call for the same child fails `self.data.remove` (hence the misleading error: the item exists, it just is not a root anymore), overwrites `item.parent`, and `add_child` appends the child to the new parent without detaching it from the previous one. Result: the child sits in several children lists while `parent` points to the last one, an inconsistent tree.

Design question to settle first: either multi-parenting is legitimate and `parent()` should detach from the previous parent, or it is not and the second claim should be rejected with an accurate warning. Happy to prepare a small PR with a red/green test once the intended semantics are confirmed.

pentux-GitHub:
First part of the fix agreed with Diego on #1307 (the bryce sample cleanup will come as a separate PR).

## The bug

Real 0.6 data files can list the same task as `<sub>` of several parents: the bryce sample does it 36 times, and it came from a real database. `BaseStore.parent()` honored every claim: the second one overwrote `item.parent`, failed `self.data.remove` (producing the misleading `item not found in data` error first seen while investigating #109: the item exists, it just is not a root anymore), and `add_child` appended the child to the new parent without detaching it from the previous one. The child ended up in several children lists at once while `parent` pointed at the last claimer — an inconsistent tree, built silently at load time.

## The fix

Multi-parenting was a liblarch feature, never exposed to the UI, and deliberately dropped from the new core (see Diego's confirmation in the issue). `parent()` now enforces that decision: **the first parent wins**, a further claim for a different parent is refused with an accurate warning, and a re-claim by the same parent is a silent no-op. Moving an item still works through `unparent()` first — which is what every UI call site already does (task and tag drag-and-drop, multi-selection reparenting all call `unparent()` before `parent()`), so no behavior change for healthy flows.

## Tests

Red/green demonstrated at two levels: new unit tests on `BaseStore` reproduce the corruption exactly (parent overwritten, child duplicated into the second parent's children, the same `item not found in data` log as on bryce), and an integration test loads an XML file where two tasks claim the same `<sub>` — the bryce scenario. Full test suite passes (287 passed, 1 pre-existing xfail).

@nekohayo regarding your question whether this could explain #1153 and #924: quite possibly for some occurrences — both are `get_parents`-flavored crashes on inconsistent objects, and this bug did build inconsistent trees from valid-looking files. I would suggest retesting them once this lands rather than closing them blindly; I can help with that.

Fixes #1307

### DIFF
--- a/GTG/core/base_store.py
+++ b/GTG/core/base_store.py
@@ -221,6 +221,22 @@ class BaseStore(GObject.Object,Generic[S]):
             raise KeyError("parent_id is not in store: "+str(parent_id))
 
         item = self.lookup[item_id]
+
+        if item.parent is not None:
+            if item.parent.id == parent_id:
+                return
+            # Multi-parenting is not supported in the new core, but real
+            # 0.6 data files can claim the same task as subtask of
+            # several parents (#1307). Honoring the second claim used to
+            # overwrite item.parent and leave the child in several
+            # children lists at once, an inconsistent tree. The first
+            # parent wins; moving an item requires unparent() first, as
+            # every UI call site already does.
+            log.warning('Refusing to parent %s to %s: already a child of '
+                        '%s (first parent wins, see #1307)',
+                        item_id, parent_id, item.parent.id)
+            return
+
         item.parent = self.lookup[parent_id]
 
         try:
--- a/tests/core/test_base_store.py
+++ b/tests/core/test_base_store.py
@@ -136,6 +136,57 @@ class TestBaseStoreParent(TestCase):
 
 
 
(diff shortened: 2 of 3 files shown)


diegogangl:
LGTM, thanks!

PR #1315, opened 2026-08-02
chore(data): remove multi-parent subtask claims from the bryce sample

pentux-GitHub:
Second part of the fix agreed with Diego on #1307 (the loader hardening is #1314; the two are independent and can land in any order).

## What

The bryce sample listed 36 tasks as `<subtask>` of more than one parent — 50 excess claims in total, up to 7 parents for a single task. Multi-parenting was never exposed to the UI and was deliberately dropped from the new core, so these entries only exercised the tree corruption documented in #1307 and produced `item not found in data` errors on every load of the sample.

## How

Cleanup policy matches the hardening in #1314: the first parent in document order wins, every later claim for the same child is removed. The diff is pure line deletions — 50 `<subtask>` lines removed, nothing else touched (no reformatting, no declaration change).

## Verification

- The file stays valid XML: 659 → 609 `<subtask>` entries, 0 multi-parent claims remaining, 0 ghost references
- Loaded live on the 1771-task dataset: zero `item not found in data` occurrences (previously one per excess claim), tree displays normally, the 36 affected tasks keep their first parent

### DIFF
--- a/data/test-data/bryce/xdg/data/gtg/gtg_tasks.xml
+++ b/data/test-data/bryce/xdg/data/gtg/gtg_tasks.xml
@@ -505,7 +505,6 @@ m mmmmmmmmmmmmmmmmmmmmmmm</content>
 		<startdate>2010-04-19</startdate>
 		<donedate>2010-04-27</donedate>
 		<subtask>312@1</subtask>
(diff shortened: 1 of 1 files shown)


diegogangl:
Thanks!

PR #1317, opened 2026-08-02
fix(core): repair backup purge targeting, then archive old dailies with gzip (#443)

### LINKED ISSUE #443: Space-saving automatic data files compression for archiving old backups
GTG has a built-in automatic daily backup system for user data.
The only thing is, after a while (years), the collection of backups starts becoming a bit heavy (even though disk space is cheap these days) and it also feels a bit messy when you look at the data folder as a user.

Personally I have found that instead of removing old backup files, I'm better off just zipping them (that way they are always around if I ever need to do some archeological digging, and it allows users to keep the data around for the eventual prophecized day where GTG could process it all into some sort of amazing local analytics thing, like #424 and/or #435). Once the filesize is minimized through compression, keeping the data around has no cost. To put things in perspective:
* my data files from November 2017 to September 2019 weighted 277 MB, and once compressed into a .tar.xz they become a single 2.2 MB file;
* similarly, my data from February 2014 to November 2017, weighted 418.4 MB on disk but shrunk down to 5.7 MB once put into a .tar.xz.

If someone wants a fun and easy project/challenge to make a plugin to offer a GUI to do that, it would be a welcome addition. There are various ways to do this:
* An action that shows up in the cleanup section of the preferences GUI that says "Your data files daily backups are currently using %d MB of disk space. You can compress and archive old backups to save approximately 90% of that space. ((Compress now))".
* The same thing, but in a GtkInfobar that gets shown in the main window (rather than the preferences GUI) everytime the data weighs more than 250 MB (maybe the threshold size could be configurable by users in the plugin's GUI settings if they somehow need something more, or less, aggressive)
* Just doing it automatically in the background when the threshold is reached, or simply packing one compressed archive per year, archiving whatever files are present from the previous (not the current) year, so compressed archives would clearly be organized yearly.

Ideally, any such processing operation should be async and not block the GTK mainloop.

Alternatively, there's the theoretical question, "what if GTG transparently used compression on individual files at all times as a core feature?" But the downsides of that would be that you'd lose the ability to use diff or Meld to compare files, and there could be potentially significant overhead with (un)compressing all the time.

### LINKED ISSUE #1316: purge_backups() prunes the data directory instead of the backup directory (silent data loss for 0.6 migrants)
While preparing a patch for #443 I found that `purge_backups()` in `GTG/core/datastore.py` walks `os.path.dirname(path)` — the **data directory** — while its docstring and its call site in `write_backups` clearly target the `backup/` subdirectory.

Some history for context: before 2022 the function called `os.listdir` on a file path and crashed, so it never actually ran. Commit `3dfbae88` fixed that crash by switching to `dirname(path)`, which made the code run — but on the wrong directory. Since it stopped crashing, the slip was invisible.

Two consequences, both reproduced with a scripted scenario on current master:

1. **The dated daily backups (`*.YYYY-MM-DD.bak`) live in `backup/` and are therefore never removed.** The unbounded accumulation described in #443 is still true on 0.7.
2. **Any file older than 30 days in the data directory is silently deleted.** For users migrating from 0.6, this includes their pre-migration `gtg_tasks.xml`, `tags.xml` and `projects.xml` — gone one month after the migration, which is exactly when they might want them back. This is why I would suggest considering it for 0.7 despite the milestone being frozen: every 0.6 migrant walks into it.

A fix is ready (red/green tested): target the backup directory, only match dated daily backups (the rotating `.bak.0-N` copies are managed by count in `write_backups` and must survive long periods without opening the app), ignore foreign files, tolerate a missing backup directory, and log instead of crashing on removal failures. PR follows as a draft together with the #443 archival feature it unblocks.

pentux-GitHub:
Draft pending design validation on #443 — two commits, deliberately separable:

**Commit 1 — the purge fix** (see #1316 for the full story): `purge_backups()` pruned the data directory instead of `backup/`. Dated dailies were never removed, and files older than 30 days in the data directory — including 0.6 migrants' pre-migration files — were silently deleted. The purge now targets `backup/`, matches only dated daily backups, never touches the rotating `.bak.0-N` copies (managed by count, and they must survive long periods without opening the app), and logs instead of crashing. If the archival design below needs more discussion, this commit can be cherry-picked on its own — the data-loss half is the urgent part.

**Commit 2 — #443, the archival**: `write_backups` now gzips dailies older than 30 days instead of deleting them (design rationale posted on #443). Byte-for-byte verification before removing any plain file; `.gz` archives kept forever; automatic restore path untouched.

**Tests**: 12 new tests in `tests/core/test_datastore_backups.py`, red/green demonstrated for the purge bug (both faces: data directory survival, old dailies removal). Full suite passes (293 passed, 1 pre-existing xfail). Also verified live: a staged data directory with 0.6 leftover files and a 40-day-old daily goes through `write_backups` with the leftovers intact and the old daily gzipped byte-identical.

Fixes #1316. Closes #443 if the design is accepted.

### DIFF
--- a/GTG/core/datastore.py
+++ b/GTG/core/datastore.py
@@ -18,6 +18,7 @@
 
 """The datastore ties together all the basic type stores and backends."""
 
+import gzip
 import os
 import re
 import threading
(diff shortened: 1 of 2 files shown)


nekohayo:
It seems this MR's first commit was already merged somewhere else? If so this would need to be rebased to drop it, description to be edited, and see the considerations for potential alternative format in https://github.com/getting-things-gnome/gtg/issues/443#issuecomment-5552837387

PR #1320, opened 2026-08-04
fix(editor): keep indented subtask bullets from leaking into the title

### LINKED ISSUE #1319: Subtask titles keep the bullet prefix when the "- " line is indented
In the task editor, typing `- something` on its own line turns it into a subtask and strips the `- ` bullet. But the stripping is inconsistent with the detection, so an indented bullet line leaks the prefix into the stored task title.

**Mechanism** (`GTG/gtk/editor/taskview.py`, `detect_subtasks`): the guard tolerates leading whitespace — `text.lstrip().startswith('- ')` — but the two removals downstream operate on raw positions: the buffer deletes a fixed 2 characters, and the title is built from `text[2:]`. When the line has leading whitespace, those first 2 characters are not the `- `, so the leftover ends up in the saved `<title>`.

**Reproduced on current master:**

| typed line | stored title |
|---|---|
| `- sousou` | `sousou` (correct) |
| ` - sousou` (indented) | ` sousou` |
| `  - sousou` | `- sousou` |

Confirmed directly in `gtg_data.xml`, so it is persisted data, not a display glitch. Fix incoming as a PR.

pentux-GitHub:
Fixes #1319.

`detect_subtasks()` accepts a `- ` bullet after leading whitespace (`text.lstrip().startswith('- ')`), but the removal cut a fixed 2 characters from the raw line and built the title from `text[2:]`. On an indented bullet those 2 characters were not the `- `, so the leftover whitespace or dash leaked into the stored `<title>`: a line ` - sousou` produced the title ` sousou`, and `  - sousou` produced `- sousou`. Confirmed as persisted data in `gtg_data.xml`, not just a display glitch.

The bullet span (leading whitespace + `- `) and the title are now computed once from the same stripped position in a small pure helper, `subtask_bullet_span()`, so the buffer deletion and the title stay in sync regardless of indentation. A genuine second dash typed by the user (`- - foo`) is preserved in the title.

**Tests**: 8 unit tests in `tests/gtk/test_taskview_subtask_bullet.py`, following the display-less pattern of `test_taskview_tag_guard.py`. Red/green verified (the indented cases fail without the fix). Full suite: 295 passed, 1 xfailed. Also verified live on master: typing ` - test` in an editor stores the title `test` cleanly, with correct parenting confirmed in the XML.

### DIFF
--- a/GTG/gtk/editor/taskview.py
+++ b/GTG/gtk/editor/taskview.py
@@ -59,6 +59,26 @@ def tag_is_faithfully_representable(name: str) -> bool:
     match = TAG_REGEX.match(candidate)
     return bool(match) and match.group(0) == candidate
 
+
+def subtask_bullet_span(text: str):
+    """Locate the "- " subtask bullet on a line, tolerating indentation.
+
+    Returns (offset, title) where offset is the number of characters to
+    delete from the start of the line (leading whitespace plus the "- ")
+    and title is the subtask name, or None when the line is not a subtask
+    bullet. Computing both from the same stripped position keeps the
+    buffer deletion and the stored title consistent, so an indented
+    bullet no longer leaks its leading whitespace into the task title.
+    """
+    stripped = text.lstrip()
+    if not stripped.startswith('- '):
+        return None
+    title = stripped[2:]
+    if not title:
+        return None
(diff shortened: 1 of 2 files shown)


diegogangl:
LGTM, thanks!

Issue #1323, opened 2026-08-04
CalDAV: no clear errors shown in GUI, and tons of tracebacks when configuring an account to sync only specific tags


nekohayo:
After #1322, with the latest git master version run through GNOME Builder and successfully connected to @pentux-GitHub's test server, if I try to configure it to sync only the "cool" tag, I get a million tracebacks in various situations below.

---

## When trying to specify a tag (or more) to synchronize:

When clicking "Enable":

```python-traceback
Traceback (most recent call last):
  File "/app/lib/python3.13/site-packages/GTG/gtk/backends/backendstree.py", line 121, in on_backend_state_changed
    new_tags = self._get_markup_for_tags(backend.get_attached_tags())
  File "/app/lib/python3.13/site-packages/GTG/gtk/backends/backendstree.py", line 134, in _get_markup_for_tags
    tags_txt = get_colored_tags_markup(self.ds, tag_names)
  File "/app/lib/python3.13/site-packages/GTG/gtk/colors.py", line 148, in get_colored_tags_markup
    tag_markups = [get_colored_tag_markup(ds, t) for t in tag_names]
                   ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
  File "/app/lib/python3.13/site-packages/GTG/gtk/colors.py", line 133, in get_colored_tag_markup
    tag_color = tag.get_attribute("color")
                ^^^^^^^^^^^^^^^^^
AttributeError: 'Tag' object has no attribute 'get_attribute'
```

Then disabling, closing the synchronization dialog, reopening it, you get:

```python-traceback
Traceback (most recent call last):
  File "/app/lib/python3.13/site-packages/GTG/gtk/application.py", line 416, in open_backends_manager
    self.open_edit_backends()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/lib/python3.13/site-packages/GTG/gtk/application.py", line 511, in open_edit_backends
    self.backends_dialog = BackendsDialog(self.ds)
                           ~~~~~~~~~~~~~~^^^^^^^^^
  File "/app/lib/python3.13/site-packages/GTG/gtk/backends/__init__.py", line 74, in __init__
    self._create_widgets_for_treeview()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/lib/python3.13/site-packages/GTG/gtk/backends/__init__.py", line 167, in _create_widgets_for_treeview
[...]

Issue #1322, opened 2026-08-04
CalDAV: no clear errors shown in GUI when setting up account with invalid username or password credentials, many tracebacks breaking the app


nekohayo:
With the latest git `master` version run through GNOME Builder, when trying to connect to your @pentux-GitHub's test server with incorrect credentials (badly copy-pasted password from Jitsi's chat), the sync failed on next startup with this traceback:

```python-traceback

Exception in thread Thread-1 (__backend_startup):
Traceback (most recent call last):
  File "/usr/lib/python3.13/threading.py", line 1044, in _bootstrap_inner
    self.run()
    ~~~~~~~~^^
  File "/usr/lib/python3.13/threading.py", line 995, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.13/site-packages/GTG/core/datastore.py", line 720, in __backend_startup
    backend.start_get_tasks()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/lib/python3.13/site-packages/GTG/core/interruptible.py", line 38, in new
    return fn(*args)
  File "/app/lib/python3.13/site-packages/GTG/backends/periodic_import_backend.py", line 79, in start_get_tasks
    self._start_get_tasks()
    ~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/lib/python3.13/site-packages/GTG/backends/periodic_import_backend.py", line 98, in _start_get_tasks
    self.do_periodic_import()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/lib/python3.13/site-packages/GTG/core/interruptible.py", line 38, in new
    return fn(*args)
  File "/app/lib/python3.13/site-packages/GTG/backends/backend_caldav.py", line 154, in do_periodic_import
    self._do_periodic_import()
    ~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/lib/python3.13/site-packages/GTG/backends/backend_caldav.py", line 188, in _do_periodic_import
    self._refresh_calendar_list()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/lib/python3.13/site-packages/GTG/backends/backend_caldav.py", line 286, in _refresh_calendar_list
    raise error
[...]

PR #1325, opened 2026-08-05
Fix CalDAV configuration dialog crashes (#1323, completes #1322)

pentux-GitHub:
Two distinct crashes hit the backend configuration dialog around CalDAV. This PR fixes both.

## 1. Flood of tracebacks when syncing by tag (#1323)

**Symptom:** configuring a CalDAV account to sync only certain tags triggered a cascade of AttributeError.

**Mechanism:** the GTK4 port turned tag attributes into GObject properties. A Tag no longer answers `get_attribute("color")`, it exposes `.color`. `colors.py` and the editor still called the removed method, raising AttributeError on every tag-color render. In the backends dialog this surfaced as a flood of tracebacks as soon as an account was set to sync a tag.

**Fix:** read `.color` instead of `get_attribute("color")`. While here, drop the dead GTK3 `Gdk.Color` branch (it does not exist under GTK4 and was itself raising, hidden by a broad except), and guard `find()`, which now raises KeyError rather than returning None.

## 2. Crash when removing a failed account (completes #1322)

**Symptom:** removing a broken CalDAV account raised an AttributeError that left the user stuck.

**Mechanism:** `on_backend_state_changed` looked up the backend by id then called `get_human_name()` on it, but a removed account can already be gone from the datastore by the time the callback runs (`get_backend` returns None). Likewise, `ConfigurePanel` connects global signals that can reach a panel whose `set_backend()` has not run, with `self.backend` undefined.

**Fix:** guard against a missing backend in the callback, default `self.backend` to None, and bail out cleanly from the handlers when no backend is selected.

## Tests

Red-before / green-after demonstrated, unit tests added (`test_colors_tag_markup.py`, `test_backends_signal_guards.py`), verified live on a real CalDAV account (sync-by-tag with no traceback, account removal with no crash).

@nekohayo having reported these crashes from the field, a retest on your side under real conditions would be especially welcome before merge.

### DIFF
--- a/GTG/gtk/backends/backendstree.py
+++ b/GTG/gtk/backends/backendstree.py
@@ -113,6 +113,11 @@ class BackendsTree(Gtk.TreeView):
             b_iter = self.backendid_to_iter[backend_id]
             b_path = self.liststore.get_path(b_iter)
             backend = self.ds.get_backend(backend_id)
+            if not backend:
+                # The backend was removed between the signal and this
+                # callback (e.g. a failed account being torn down);
+                # nothing left to refresh in the row.
+                return
             backend_name = backend.get_human_name()
             self.liststore[b_path][self.COLUMN_TEXT] = backend_name
             self.liststore[b_path][self.COLUMN_ENABLED] = backend.is_enabled()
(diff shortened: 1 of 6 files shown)


diegogangl (approved):

diegogangl:
Hi @pentux-GitHub , looks like this branch has some conflicts now

Issue #1324, opened 2026-08-04
CalDAV sync backend keeps trying to sync while computer is offline (or lost connection), leading to cascading tracebacks and errors


nekohayo:
After finding #1322 and #1323 for @pentux-github on git `master`, I suspended my laptop and resumed it somewhere else where I had either no internet connection, or a very poor wifi signal, and below is what I got. GTG (or at least the caldav backend) is apparently not using gnetworkmonitor (or networkmanager, or whatever) to monitor the actual "global connectivity" online status of the operating system's network, and it just silently encounters cascading failure tracebacks all over the place.

It should not even attempt any of this if there is no network, and if it does encounter network problems when the network is reportedly up, then it should handle those tracebacks correctly.

---

```
2026-08-04 15:36:03,706 - INFO - backend_caldav:_do_periodic_import:193 - Fetching todos from 'https://pentux-server/remote.php/dav/calendars/testgtg/personal/'
Gtk-Message: 15:36:04.253: GtkDialog mapped without a transient parent. This is discouraged.
```

```python-traceback
Traceback (most recent call last):
  File "/app/lib/python3.13/site-packages/GTG/gtk/application.py", line 416, in open_backends_manager
    self.open_edit_backends()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/lib/python3.13/site-packages/GTG/gtk/application.py", line 511, in open_edit_backends
    self.backends_dialog = BackendsDialog(self.ds)
                           ~~~~~~~~~~~~~~^^^^^^^^^
  File "/app/lib/python3.13/site-packages/GTG/gtk/backends/__init__.py", line 74, in __init__
    self._create_widgets_for_treeview()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/lib/python3.13/site-packages/GTG/gtk/backends/__init__.py", line 167, in _create_widgets_for_treeview
    self.backends_tv = BackendsTree(self)
                       ~~~~~~~~~~~~^^^^^^
  File "/app/lib/python3.13/site-packages/GTG/gtk/backends/backendstree.py", line 50, in __init__
    self.refresh()
    ~~~~~~~~~~~~^^
  File "/app/lib/python3.13/site-packages/GTG/gtk/backends/backendstree.py", line 66, in refresh
    self.add_backend(backend)
    ~~~~~~~~~~~~~~~~^^^^^^^^^
[...]

PR #1326, opened 2026-08-05
Fix CalDAV sync crashes on auth failure and offline edits (#1322, #1324)

pentux-GitHub:
CalDAV synchronization killed its sync thread on several error paths, with a silent data loss in one case. This PR addresses all of them.

## 1. Authentication failure (#1322)

**Symptom:** a wrong password raised an AuthorizationError that killed the sync thread, with no clear message.

**Mechanism:** `_refresh_calendar_list` did catch the error to notify the UI, but then re-raised it (`raise`), and nothing caught it on the thread path.

**Fix:** classify the failure (`_dav_failure_errno`: auth vs network) and report it to the UI via `backend_failed` (clear banner, Configure button) instead of re-raising. The thread survives, the periodic cycle retries.

## 2. Offline traceback cascade (#1324, import side)

**Symptom:** while the machine was offline, the backend kept attempting syncs, producing a cascade of connection errors.

**Fix:** an `is_connection_up()` guard at the top of `do_periodic_import`. Offline, the import is skipped cleanly; the cycle resumes when the network returns.

## 3. Silent data loss on offline edits (#1324, write side)

**Symptom:** a task edited offline was lost. When the network returned, the push failed, killed the write thread, and the edit never reached the server.

**Mechanism:** in `launch_setting_thread`, the tid is popped from the queue before the push. If the push raised a non-DAVError (typically a network error), it propagated, killed the thread, and the tid was never re-queued. Worse, the cleanup line (`to_set_timer = None`) was skipped, locking the queue for good until restart.

**Fix:** wrap the push and delete in a try/except that re-queues the task and stops the current drain cleanly (no spin), always reaching the cleanup line (no more lock). A retry cap (N=3) bounds a task that would fail in a loop, without blocking the queue head. An offline failure does not count toward that cap (transient). Finally, a trigger on connectivity return (`network-changed`) relaunches pending writes without waiting for a new edit.

**Scope:** the write-side fix lives in `generic_backend.py` (shared by all backends). In practice, CalDAV is the only concrete backend taking this path in the new core. The semantics change for any future backend: an exception from `set_task` no longer kills the thread, which is strictly safer.

## Tests

Red-before / green-after demonstrated. Unit tests added (`test_caldav_error_taxonomy.py`, `test_sync_queue_resilience.py`). Verified live: wrong password (banner, no crash), offline edit then network return (the change reaches the server, confirmed on the Nextcloud side, with no loss).

These three error paths were surfaced by @nekohayo's feedback during the hackfest; a retest on your side, especially on the offline scenario, would be valuable to confirm the behavior before merge.

### DIFF
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -29,12 +29,14 @@ from gettext import gettext as _
 from hashlib import md5
 
 import caldav
+import requests
 from dateutil.tz import UTC
 from GTG.backends.backend_signals import BackendSignals
 from GTG.backends.generic_backend import GenericBackend
 from GTG.backends.periodic_import_backend import PeriodicImportBackend
 from GTG.core.dates import LOCAL_TIMEZONE, Accuracy, Date
 from GTG.core.interruptible import interruptible
+from GTG.core.networkmanager import is_connection_up
 from GTG.core.tasks import Task, Status as TaskStatus
 from vobject import iCalendar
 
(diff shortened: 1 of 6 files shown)


diegogangl:
LGTM, thanks!

PR #1327, opened 2026-08-05
Fix backup purge targeting the data directory instead of backup/ (#1316)

pentux-GitHub:
`purge_backups()` pruned the wrong directory, with two consequences including silent data loss for users migrating from 0.6.

**Symptom:** the function walked `os.path.dirname(path)`, i.e. the *data* directory, not the `backup/` subdirectory its docstring and call site target.

**Mechanism:** before 2022 the function called `os.listdir` on a file path and crashed, so it never actually ran. Commit `3dfbae88` fixed that crash by switching to `dirname(path)`, which made the code run, but on the wrong directory. Since it stopped crashing, the slip went unnoticed.

**Two consequences, both reproduced with a scripted scenario:**
- dated daily backups (`*.YYYY-MM-DD.bak`) live in `backup/` and were therefore never removed: the unbounded accumulation described in #443 is still true on 0.7;
- any file older than 30 days in the data directory was silently deleted. For users migrating from 0.6 this includes their pre-migration `gtg_tasks.xml`, `tags.xml` and `projects.xml`, gone one month after the migration.

**Fix:** the purge now targets the `backup/` directory, only matches dated daily backups (the rotating `.bak.0-N` copies, managed by count in `write_backups`, must survive long periods without opening the app), ignores foreign files, tolerates a missing backup directory, and logs instead of crashing when a removal fails.

**Tests:** red-before / green-after demonstrated, unit tests added (`test_datastore_backups.py`, covering correct targeting, preservation of the `.bak.0-N` copies, a missing backup directory, and foreign files).

---

This extracts the critical purge fix from #1317 so it can land independently of the gzip archiving feature, which remains in #1317 pending the #443 design discussion.

### DIFF
--- a/GTG/core/datastore.py
+++ b/GTG/core/datastore.py
@@ -19,6 +19,7 @@
 """The datastore ties together all the basic type stores and backends."""
 
 import os
+import re
 import threading
 import logging
 import shutil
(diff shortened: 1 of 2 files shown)


diegogangl:
Nice catch, thanks

PR #1328, opened 2026-08-05
flatpak: regenerate python3-requirements against SDK 50, add setproctitle

### LINKED ISSUE #638: Set the app's process name (instead of "python3")

pentux-GitHub:
Fixes #638

## What

The `setproctitle` import added in #641 has been failing silently in
the flatpak ever since, because the dependency was never added to the
manifest — so the process still shows up as `python3` in ps and
system monitors.

While adding it, this regenerates `python3-requirements.yaml` fully:
it was still generated against SDK 47 while the manifest targets
runtime 50 since #1292, and lxml had drifted from requirements.txt
(5.3.0 in the yaml vs 6.1.0 pinned). The `six` wheel hand-patched in
#1292 is kept, now resolved as a regular caldav dependency.

Command used (same as the original, bumped to SDK 50):

    flatpak-pip-generator --runtime=org.gnome.Sdk//50 \
      --requirements-file=build-aux/requirements.txt --yaml \
      --output=build-aux/python3-requirements --ignore-installed=lxml

## Why setproctitle and not GLib.set_prgname

As @chergert explained in the issue, `GLib.set_prgname()` only changes
the process-internal representation; what `ps` shows comes from
`/proc/$pid/cmdline`, which on Linux requires overwriting argv memory —
which is exactly what the `setproctitle` module does. The code in
`gtg.in` was correct all along; only the packaging was missing.

## What the regeneration pulls in

Transitive dependencies are refreshed against today's PyPI, including
some major bumps: icalendar 6.1.1 → 7.2.2, recurring_ical_events
3.4.1 → 3.8.2, pytest 8 → 9. The flatpak builds and the app starts
with the CalDAV backend loading; I have not run a full CalDAV sync
cycle against these versions.

## Verified

Built and installed locally (flatpak-builder, runtime 50):

    $ ps -eo pid,comm,args | grep gtg
    68190 bwrap    bwrap --args 76 -- gtg
    68201 bwrap    bwrap --args 76 -- gtg
    68202 gtg      gtg

The two bwrap wrapper processes remain, as expected — that's flatpak
itself, out of reach of setproctitle.

Note: since this is a small packaging-only fix labeled
low-hanging-fruit, it could be a candidate for the 0.7 milestone
rather than 0.8, if you think it fits.

### DIFF
--- a/build-aux/python3-requirements.yaml
+++ b/build-aux/python3-requirements.yaml
@@ -1,17 +1,18 @@
-# Generated with flatpak-pip-generator --runtime=org.gnome.Sdk//47 --requirements-file=build-aux/requirements.txt --yaml --output=build-aux/python3-requirements --ignore-installed=lxml
-build-commands: []
+# Generated with flatpak-pip-generator --runtime=org.gnome.Sdk//50 --requirements-file=build-aux/requirements.txt --yaml --output=build-aux/python3-requirements --ignore-installed=lxml
+name: python3-requirements
 buildsystem: simple
+build-commands: []
 modules:
   - name: python3-lxml
     buildsystem: simple
     build-commands:
       - pip3 install --verbose --exists-action=i --no-index --find-links="file://${PWD}"
-        --prefix=${FLATPAK_DEST} "lxml==5.3.0" --ignore-installed --no-build-isolation
+        --prefix=${FLATPAK_DEST} "lxml==6.1.0" --ignore-installed --no-build-isolation
     sources:
       - &id001
         type: file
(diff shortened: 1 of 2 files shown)


diegogangl:
Definitely, thanks!

PR #1342, opened 2026-08-09
fix(core): inherit recurrence when duplicating a completed recurring task (#1339)

### LINKED ISSUE #1339: Recurring tasks completely disappear from the Open/Actionable tasks list when marking them as "Done" 2 or 3 times
To reproduce with git `master`:

1. Set a task as recurring everyday
2. Mark it as "Done"
3. Mark its new instance as "Done"
4. Mark its third new instance as "Done"

Result: it will eventually disappear from the Open/Actionable tasks completely.

As demonstrated here in front of @pentux-GitHub:

https://github.com/user-attachments/assets/44929a44-b547-41cc-80d8-701883bd1395

pentux-GitHub:
Completing a recurring task duplicates it as its next occurrence. On current master the duplicate is born non-recurring, so completing it produces no further occurrence and the task vanishes from the Open and Actionable lists, exactly as demonstrated in #1339. The disappearance is deterministic at the second completion at the core level.

The mechanism is a line lost in the port to the new core. The 0.6 core inherited the recurrence when duplicating, with an explicit comment reading "Inherit the recurrency". The new `duplicate_for_recurrent` copies title, tags, content and dates, but neither the recurring flag nor the recurring term, so the duplication condition in `set_status` is never met again.

The fix restores the inheritance: the duplicate receives `set_recurring(True, task.recurring_term)` before its due date is computed, which also covers children duplicated along with a recurring parent since the duplication recurses through them.

Red before, green after: four regression tests in `tests/core/test_recurrence_duplication.py`, of which three fail on unpatched master. They cover the inheritance itself, the next occurrence date, a chain of five consecutive completions, and the parent-with-children case. Verified live against the exact reproduction recipe from the issue.

Fixes #1339

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -887,6 +887,10 @@ class TaskStore(BaseStore[Task]):
         new_task.tags = task.tags
         new_task.content = task.content
         new_task.date_added = task.date_added
+        # Inherit the recurrence, as the 0.6 core did: without this the
+        # duplicate is born non-recurring and the chain dies at its next
+        # completion (#1339).
+        new_task.set_recurring(True, task.recurring_term)
         new_task.date_due = task.get_next_occurrence()
 
         # Only goes through for the first task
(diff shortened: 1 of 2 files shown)


diegogangl:
LGTM, thanks!

PR #1343, opened 2026-08-09
fix(browser): repair the two dead entries of the weekly recurrence menu (#1339)

pentux-GitHub:
Two entries of the Weekly recurrence submenu silently failed to enable recurrence. A task configured through them was never recurring at all, so marking it Done closed it without spawning a next occurrence, the same symptom as reported in #1339 through another path.

The "On Monday" item was the only weekday missing its `action` attribute in `recurring_menu.ui`: the translator comment sits exactly where the action stands in every sibling entry, and carries a TRNASLATORS typo suggesting the line was overwritten by hand. A GTK4 menu item without an action is inert, so clicking On Monday did nothing at all.

The "On This Day" item passed the target `THIS` as a recurring term, a value handled nowhere in the menu class or in `dates.py`: `set_recurring()` rejected it and silently fell back to non-recurring. The item now triggers the existing every-week action, which is exactly what its label promises.

A structural test on the menu definition (`tests/gtk/test_recurring_menu_ui.py`) now verifies that every item carries an action, that every action is one the menu class installs, and that every weekday target is a parsable term. It fails on the previous menu definition and would have caught both defects.

Together with the core inheritance fix in #1342, this completes #1339.

### DIFF
--- a/GTG/gtk/data/recurring_menu.ui
+++ b/GTG/gtk/data/recurring_menu.ui
@@ -22,7 +22,8 @@
         <section>
           <item>
             <attribute name="label" translatable="yes">On Monday</attribute>
-            <!-- TRNASLATORS: this is locale dependant and should be the correct full
+            <attribute name="action">recurring_menu.recurr_week_day</attribute>
+            <!-- TRANSLATORS: this is locale dependant and should be the correct full
             name of the weekday -->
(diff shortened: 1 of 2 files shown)


diegogangl:
Thanks!

PR #1344, opened 2026-08-09
fix(browser): restore task sorting fully across restarts and views (#1248)

### LINKED ISSUE #1248: Sorting order not restored on startup with 0.7 (take 3)

pentux-GitHub:
Retesting #1248 on current master narrowed the picture: the sort mode itself is now restored, but only for the view GTG starts on, and three more defects hide behind it.

The sort direction is never saved: whatever the user chose, every restart comes back ascending, while the button icon, hardcoded in the ui file, keeps claiming descending. The sort popovermenu is never synchronized either, so it always shows "Sort by Title" whatever the list actually does. And a naming slip makes it worse: store_sorting matches the pane name "actionable", which get_selected_pane() never returns (the real id is "workview"), so the Actionable view stores its mode into the Open view key and overwrites it.

The fix keeps the pane-to-key mapping in one table, adds three sort_order_* configuration keys defaulting to ASC (the previous effective behavior, so nothing changes for existing users until they pick a direction), restores mode and direction for all three panes at startup, persists the direction when it changes, reapplies it when the mode changes since a fresh sorter is born ascending, and synchronizes the menu state and button icon at startup and on every view switch.

Regression test on the configuration defaults: red on master, where the direction keys are missing, green after. Verified live: different modes and directions on the Open and Actionable views all come back after a restart, menu and icon included.

Fixes #1248

### DIFF
--- a/GTG/core/config.py
+++ b/GTG/core/config.py
@@ -58,6 +58,9 @@ DEFAULTS = {
         'sort_mode_open': 'title',
         'sort_mode_active': 'title',
         'sort_mode_closed': 'title',
+        'sort_order_open': 'ASC',
+        'sort_order_active': 'ASC',
+        'sort_order_closed': 'ASC',
         'selected_tag': '',
     },
     'tag_editor': {
--- a/GTG/gtk/browser/main_window.py
+++ b/GTG/gtk/browser/main_window.py
@@ -54,6 +54,15 @@ PANE_STACK_NAMES_MAP = {
 }
 PANE_STACK_NAMES_MAP_INVERTED = {v: k for k, v in PANE_STACK_NAMES_MAP.items()}
 
+# Per-pane config keys for sorting: pane id -> (mode key, direction key).
+# The historical key names do not match the pane ids: 'sort_mode_active'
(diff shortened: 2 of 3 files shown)


diegogangl:
Thanks!

PR #1345, opened 2026-08-09
fix(browser): degrade the Synchronization dialog cleanly when no backend loads (#1337)

### LINKED ISSUE #1337: When caldav dependency is missing, the "Synchronization" UI breaks in various ways (empty widgets, broken Cancel and OK buttons, etc.)
Testing the git `master` version of GTG in front of @pentux-GitHub, I encountered this very broken state that occurs when you use `./launch.sh` directly without using GNOME Builder's flatpak environment to launch the app, on my Fedora 44 "Workstation" laptop where the "caldav" Python library is not installed:

```
WARNING - __init__:__init__:68 - Backend backend_caldav could not be loaded: ModuleNotFoundError("No module named 'caldav'")
```

Behold the resulting collection of cascading bugs in this short demonstration video:

https://github.com/user-attachments/assets/12c40f19-8a70-4d0d-aed8-93fb21a50b1b

pentux-GitHub:
On a system where the caldav python library is not installed, backend_caldav fails to import and GTG is left with zero addable backends, a case the whole Synchronization dialog assumed impossible. As shown in the #1337 video: the add panel offers only empty widgets, OK switches to the configuration panel of nothing, showing an orphan unlabeled button that raises AttributeError when clicked, and Cancel does nothing at all.

The fix adds four layers of defense, following the error-paths-first approach. The backend factory now records why each module failed to load instead of only logging it. The add button shows a clear error dialog naming the missing library when nothing is addable, rather than entering the broken panel. The configuration panel gets an explicit empty state: its widgets stay hidden until a backend is actually assigned to it, and every handler that dereferenced the backend attribute is now guarded, which also silences the negative-size warnings coming from the unlabeled sync button. Finally, cancelling the add panel now visibly returns to the initial state, a path that was inert even with caldav installed.

Verified live both ways on the reproduction from the issue: with a simulated missing caldav module the dialog stays sane and explains itself with the exact import error, and with the library present the nominal add and configure flow is unchanged.

Two further defects observed during testing are left out of scope on purpose, one problem per PR: the dialog window is badly sized on first opening, and the CalDAV parameter fields show hardcoded sample values instead of greyed placeholder text. Both predate this change and deserve their own issues.

Fixes #1337

### DIFF
--- a/GTG/backends/__init__.py
+++ b/GTG/backends/__init__.py
@@ -54,6 +54,7 @@ class BackendFactory(Borg):
             # This object has already been constructed
             return
         self.backend_modules = {}
+        self.failed_modules = {}
         backend_files = self._find_backend_files()
         # Create module names
         module_names = [f.replace(".py", "") for f in backend_files]
@@ -64,7 +65,9 @@ class BackendFactory(Borg):
             try:
                 __import__(extended_module_name)
             except ImportError as exception:
-                # Something is wrong with this backend, skipping
+                # Something is wrong with this backend, skipping. Keep the
+                # reason: the Synchronization dialog reports it to the user.
+                self.failed_modules[module_name] = str(exception)
                 log.warning("Backend %s could not be loaded: %r",
                             module_name, exception)
                 continue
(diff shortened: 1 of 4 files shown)


diegogangl:
LGTM and works as expected. Thanks!

PR #1348, opened 2026-08-09
fix(browser): give the Synchronization dialog a sane default geometry (#1346)

### LINKED ISSUE #1346: Synchronization dialog opens with insufficient window size and horizontal scrolling in some languages
On current git master the Synchronization dialog opens with a wrong, cramped window size, and did so consistently across restarts while testing #1337 and #1345, both with and without a CalDAV account configured. The window definition in backends.ui sets no default width or height at all, so the initial geometry is left entirely to GTK4 size negotiation, which visibly does not produce a usable result here. Screenshot attached below.

Image

pentux-GitHub:
The backends window declared no default size at all, leaving the initial geometry entirely to GTK4 size negotiation, which opened the dialog cramped, as shown in the #1346 screenshot. The central pane also had no horizontal scrollbar policy, so any content one pixel wider than the allocation grew a horizontal scrollbar at the bottom of the panel.

Two properties in backends.ui: the window now opens at 700x500, and the central pane refuses horizontal scrolling, so a panel that genuinely needs more width makes the window grow instead of offering a scrollbar. Verified live: the dialog opens comfortable on first and subsequent openings, with no horizontal scrollbar on either the configuration or the add panel.

Fixes #1346

### DIFF
--- a/GTG/gtk/data/backends.ui
+++ b/GTG/gtk/data/backends.ui
@@ -9,6 +9,8 @@
     <property name="page_size">10</property>
   </object>
   <object class="GtkWindow" id="backends">
+    <property name="default_width">700</property>
+    <property name="default_height">500</property>
     <child>
       <object class="GtkBox" id="big_central_box">
         <child>
@@ -66,6 +68,7 @@
         <child>
           <object class="GtkScrolledWindow" id="central_pane_window">
             <property name="hexpand">True</property>
+            <property name="hscrollbar_policy">never</property>
             <property name="width_request">480</property>
             <property name="vadjustment">adjustment1</property>
             <child>


nekohayo:
Looking at the screenshot from #1346, I suspect that's because the 1st label at the top is longer in French than in English, but that label does not wrap,  so here we are. The thing is that just bumping up the default window size and not making it scrollable isn't going to be good for future mobile formfactor compatibility, however.

nekohayo:
To be clearer, I think this MR's current approach (with `<property name="hscrollbar_policy">never</property>`) is reasonable and will certainly solve the symptom, so no objection here, but in the longer term the real solution would be to redesign that dialog so that it isn't so terribly laid out :)

diegogangl:
Mergin then, thanks!

PR #1349, opened 2026-08-09
fix(browser): show CalDAV field hints as placeholders, not stored values (#1347)

### LINKED ISSUE #1329: Backends plugins configuration dialogs (ex: CalDAV) should use placeholder text in GtkEntry widgets, not real text
In git `master`, these GtkEntry widgets have real text set directly in the entry as the default (instead of placeholder text as a property) , this makes no sense:

Image

We need to use https://docs.gtk.org/gtk4/method.Entry.set_placeholder_text.html instead.

### LINKED ISSUE #1347: CalDAV parameter fields show sample values as real text instead of placeholders
The username and server fields of the CalDAV configuration panel are prefilled with the literal strings insert your username (translated) and https://example.com/webdav/. These are not greyed placeholder hints: backend_caldav.py declares them as PARAM_DEFAULT_VALUE, and parameters_ui/text.py injects them with set_text(), so they are real values. The user has to notice and erase them by hand, and if the account is saved untouched they are stored as the actual configuration, as backends.conf then shows verbatim. Switching these fields to empty defaults plus set_placeholder_text() would keep the hint while never storing it.

pentux-GitHub:
The username and server fields of the CalDAV configuration panel were prefilled with the literal strings insert your username (translated) and https://example.com/webdav/. These were not greyed hints: backend_caldav.py declared them as PARAM_DEFAULT_VALUE and parameters_ui injected them with set_text(), so they were real values. The user had to notice and erase them by hand, and saving the account untouched stored them verbatim in backends.conf as the actual configuration.

The two defaults are now empty, TextUI accepts an optional placeholder argument, and the hints moved there: greyed, vanishing as soon as the user types, and never stored. Verified live: a freshly created account shows both hints greyed and writes empty username and service-url values to backends.conf. One assumed limit: accounts already saved with the sample strings keep showing them as real text, since those are by now their stored values.

Fixes #1329, fixes #1347

### DIFF
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -134,13 +134,13 @@ class Backend(PeriodicImportBackend):
             GenericBackend.PARAM_DEFAULT_VALUE: 15},
         "username": {
             GenericBackend.PARAM_TYPE: GenericBackend.TYPE_STRING,
-            GenericBackend.PARAM_DEFAULT_VALUE: _('insert your username')},
+            GenericBackend.PARAM_DEFAULT_VALUE: ''},
         "password": {
             GenericBackend.PARAM_TYPE: GenericBackend.TYPE_PASSWORD,
             GenericBackend.PARAM_DEFAULT_VALUE: ''},
         "service-url": {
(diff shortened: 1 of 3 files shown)


diegogangl:
Thanks!

PR #1350, opened 2026-08-09
fix(browser): bring back the Closed Date sorting order (#1334)

### LINKED ISSUE #1334: No "Closed date" sorting order for Done or Dismissed tasks list
In GTG 0.6, the "Closed" tasks view was sorting tasks by their closed date, with the simple column header:

Image

In the current git `master` version leading to 0.7, this is no longer possible, as there is no "Closed date" sorting order available:

Image

Note that this view's sorting order should be remembered independently from the "Open" and "Actionable" task views' sorting order in #1248.

pentux-GitHub:
GTG 0.6 sorted the Closed view by closing date through its column header. The new task pane lost that ability entirely: no sorter compares date_closed and the sort menu offers no such entry, a 0.7 regression shown in the #1334 screenshots.

A TaskClosedSorter joins the existing sorters, comparing the date_closed field, which is always a valid Date (no_date for tasks never closed, so the comparison is total). A Sort by Closed Date entry joins the sort menu, and the task pane maps it like its siblings. The per-view memory asked for in the report is already provided by #1344: once both land, the Closed view remembers this mode and its direction independently from the two other views.

Verified live: the new entry appears in the sort menu, orders the Closed view by closing date, reverses with the direction toggle, and survives switching views back and forth.

Fixes #1334

### DIFF
--- a/GTG/core/sorters.py
+++ b/GTG/core/sorters.py
@@ -95,6 +95,20 @@ class TaskDueSorter(ReversibleSorter):
         return self.reversible_compare(first, second)
 
 
+class TaskClosedSorter(ReversibleSorter):
+    __gtype_name__ = 'ClosedSorter'
+
+    def do_compare(self, a, b) -> Gtk.Ordering:
+
+        a = unwrap(a, Task)
+        b = unwrap(b, Task)
+
+        first = a.date_closed
+        second = b.date_closed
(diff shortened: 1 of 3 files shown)


diegogangl:
LGTM, thanks!

PR #1351, opened 2026-08-09
fix(browser): re-sort the task lists when a task date or title changes (#1332)

### LINKED ISSUE #1332: When changing a task's start date, due date, closing date, etc., the task does not get re-sorted in the current list view's sorting order
As you can see in this short video, with git `master`, the sorting order position in the tasks list is not recalculated when updating a task's Due Date (for example):

https://github.com/user-attachments/assets/502fe6d0-7e11-4dff-b43f-89fc9cb6620b

pentux-GitHub:
Editing the due or start date of a task updated the row label in place but left the row at its old position, as shown in the #1332 video: nothing told the sort model that anything changed. The date setters already notify the date-due-str and date-start-str properties, which is exactly how the labels refresh; the sorting side simply had no listener.

The task store now relays those notifies through a new task-sortably-changed signal, twin of the existing task-filterably-changed, and each pane asks its sorter to recompute on either signal. Title, tag and date edits therefore all reposition the row immediately. There is no startup churn: at load time the dates are set before the task enters the store, hence before the relays are connected.

Verified live on the reproduction from the video: with the list sorted by due date, changing a task due date moves the row to its new position as soon as the editor closes, and renaming a task under title sorting does the same.

Fixes #1332

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -868,6 +868,11 @@ class TaskStore(BaseStore[Task]):
         """Signal to emit when a task was changed in a filterable way. (E.g., A tag was added.)"""
 
 
+    @GObject.Signal(name='task-sortably-changed', arg_types=(object,))
+    def task_sortably_changed_signal(self, *_):
+        """Signal to emit when a task changed in a way that can affect sorting."""
+
+
     def __str__(self) -> str:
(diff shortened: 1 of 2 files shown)


diegogangl:
Thanks!

PR #1352, opened 2026-08-09
fix(browser): hand focus back to the task list when the context menu closes (#1333)

### LINKED ISSUE #1333: After right-clicking tasks in the browser list, they have a gray background color persisting (vs blue on hover)
As you can see in the demonstration video below, tested with git `master`:

https://github.com/user-attachments/assets/d8d24bd6-4fa2-46b9-b9bc-a77978b2a9da

I think I would expect it to remain blue, and to not flash between colors like that "dependent on hover state"…

pentux-GitHub:
Right-clicking a task selects it and pops the shared context menu, which takes keyboard focus away from the list view. GTK4 then paints the selection in the unfocused gray color, and since nothing ever returned the focus to the list, the row stayed gray after the menu closed and flickered with hover, exactly as the #1333 video shows.

The pane now keeps a reference to its list view and hands the focus back when the popover emits closed. The connection is deliberately one-shot: the two context menus are shared between the three panes, so a persistent connection would pile up handlers pointing at stale panes.

One boundary stated for the record: selections still turn gray while the popover is open, or when another window such as the task editor holds the focus. That is the standard GTK unfocused rendering, identical in any GTK application, and out of scope here. The persisting gray in the focused browser window is gone, verified live on both the open and closed views, including after running a menu action.

Fixes #1333

### DIFF
--- a/GTG/gtk/browser/task_pane.py
+++ b/GTG/gtk/browser/task_pane.py
@@ -197,6 +197,7 @@ class TaskPane(Gtk.ScrolledWindow):
         tasks_signals.connect('unbind', self.task_unbind_cb)
 
         view = Gtk.ListView.new(self.task_selection, tasks_signals)
+        self.task_view = view
         view.set_show_separators(True)
         view.add_css_class('rich-list')
         view.add_css_class('task-list')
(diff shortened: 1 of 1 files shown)


diegogangl:
Thanks!

PR #1353, opened 2026-08-09
fix(browser): ellipsize long tag and saved search names in the sidebar (#1341)

### LINKED ISSUE #1341: Ensure counters count labels are visible: ellipsize text labels for tags names and saved searches names in the sidebar
This is an issue that was present in 0.6 (and previous versions), and it persists in the current git `master` (upcoming 0.7) version:

Image

As you can see, if you have long tag names, or a long saved search query, the tasks counters on the right will get "pushed out" to the right into what becomes a horizontally scrollable area.

Horizontal scroll makes no sense in the sidebar, I would expect those labels to get ellipsized (middle or end?) so that the counters can remain shown at all times instead.

pentux-GitHub:
A long tag name or saved search query used to force its full width as the minimum width of its sidebar row, pushing the task counters out into a horizontally scrollable area, a defect carried over from 0.6 as noted in #1341.

The name labels of tags and saved searches now ellipsize at the end, which collapses the row minimum width so the counters stay visible at every sidebar width and the horizontal scrolling disappears. End ellipsis is the GNOME convention; switching to middle is a one-word change if preferred. The full name remains reachable: it is bound to the label tooltip on both row kinds.

Verified live with an overlong tag name and an overlong saved search query: both truncate with an ellipsis, the counters stay visible with the sidebar shrunk to its minimum, and hovering shows the complete name.

Fixes #1341

### DIFF
--- a/GTG/gtk/browser/sidebar.py
+++ b/GTG/gtk/browser/sidebar.py
@@ -18,7 +18,7 @@
 
 """Sidebar widgets."""
 
-from gi.repository import Gtk, GObject, Gdk, Gio
+from gi.repository import Gtk, GObject, Gdk, Gio, Pango
 
 from gettext import gettext as _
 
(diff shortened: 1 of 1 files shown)


diegogangl:
LGTM, thanks!

Issue #1338, opened 2026-08-08
Task reaper: When purging old closed tasks from 0.6 in 0.7, tasks in the "Dismissed" status do not get purged?


nekohayo:
I'm not entirely sure what's going on there because I actually have no visual indication of whether the Purge button actually works or not in 0.7 (see #1336), but I _think_ the current git `master` version towards 0.7 has a problem where it can't purge some tasks using this action in the Preferences dialog:

<img width="833" height="243" alt="Image" src="https://github.com/user-attachments/assets/09b06f09-c4a8-4ae6-a427-0e360ea573c1" />

In my case, using my real dataset (from 0.6), when clicking that button nothing happens even if I set the threshold to 1 day, and looking at the list of Closed tasks in the main window that list does not seem to shrink to a small amount of tasks.

Furthermore, when opening some of those closed tasks, I saw that some of them had nothing in the "Closed date" field.

@pentux-GitHub are you able to test and reproduce this issue somehow? Did I hallucinate this bug, or does it actually exist in 0.7? If I have to guess, I would suspect something related to the import or parsing of closed date for done vs dismissed tasks of 0.6 vs 0.7, but I'm not sure…

pentux-GitHub:
Reproduced, and your hunch was close: the culprit is the 0.7 serializer, not the import. It only wrote the closed date of Done tasks, so every Dismissed task lost its date on the first save, and a dateless closed task is invisible to the reaper, today minus no_date being -9999 days. #1354 fixes the serializer and heals damaged files at load, verified on a real dataset: ten stripped dismissed tasks got honest dates back and were purged as expected.

Your done tasks not purging is a second, distinct defect: on my bench every done task had its closed date overwritten with the current day, so they all look too young to purge. Still investigating, it predates the serializer fix. The editor Dismiss button is also dead, it calls a method that does not exist, fix coming separately.

pentux-GitHub:
Both defects behind this issue are now fixed on master. #1354, merged, repairs the serializer that silently dropped the closed date of Dismissed tasks, and heals already damaged files at load. #1358, merged today, stops the CalDAV sync from overwriting closed dates with the day of the sync, which is what made every Done task look too young to purge. The dead Dismiss button in the editor is fixed as well (#1356).

[...]

PR #1354, opened 2026-08-09
fix(core): keep the closed date of dismissed tasks across saves (#1338)

pentux-GitHub:
The serializer only wrote the closed date of Done tasks: every Dismissed task lost its date_closed on the first save, and since GTG rewrites the whole file on exit, one 0.7 session was enough to strip every dismissed task. A closed task without a date is then invisible to the reaper, today minus no_date being -9999 days, never above any threshold. That is the dismissed half of #1338.

The serializer now writes the closed date for every closed status, and the loader heals closed tasks that have no closed date, falling back on the modification date with a warning, the same guard philosophy as the added date. That repairs files already damaged. Red before, green after: four round-trip tests, two failing on the old code. Verified live on a real dataset: ten stripped dismissed tasks healed with honest dates at load, then purged as expected at the next reaper run.

The remaining observation in #1338, done tasks not purged, traces to a distinct pre-existing defect, closed dates mass-overwritten with the current day, investigated separately.

Fixes the dismissed half of #1338

### DIFF
--- a/GTG/core/tasks.py
+++ b/GTG/core/tasks.py
@@ -982,6 +982,17 @@ class TaskStore(BaseStore[Task]):
             if done_element is not None and done_element.text is not None:
                 closed = Date.parse(done_element.text)
                 task.date_closed = closed
+            elif task.status is not Status.ACTIVE:
+                # A closed task without a closed date is invisible to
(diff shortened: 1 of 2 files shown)


diegogangl:
Thanks!

PR #1358, opened 2026-08-11
fix(caldav): stop stamping closed dates with the sync day (#1357)

### LINKED ISSUE #1357: CalDAV sync overwrites the closed dates of Done and Dismissed tasks with the current day
Every CalDAV sync that fills a closed task stamps its closed date with the day of the sync, destroying the historical date. On a real dataset synced since february, the damage looks like this: 107 tasks closed on 2026-02-28, 20 on 2026-07-12, 10 on 2026-03-23, each cluster matching a sync day rather than the days the tasks were actually closed. Only 2 tasks out of about 160 kept a real date.

Mechanism: Translator.fill_task() applies the completed field first, which sets the correct historical date, then the status field, whose write_gtg() goes through Task.set_status(). That method is the user action handler: it unconditionally stamps date_closed with today, overwriting the value just set. It also propagates the status to every subtask and can trigger the recurrence duplication logic, none of which a sync should do. The XML loader already avoids set_status() for exactly this reason.

Side effect visible in the data: dates written by the bug are plain dates (2026-02-28), while the rare survivors carry the UTC datetime the sync normally writes (2026-02-20T23:00:00+00:00), which is how the two code paths can be told apart in a damaged file.

Reproduction on a live Nextcloud calendar holding 138 closed todos with historical COMPLETED dates: importing on current master writes 138 tasks with <done>2026-08-11</done> (the import day) to the data file. With the fix, the same import preserves every historical date down to the file.

This is one of the two defects behind the reaper symptoms of #1338: mass-stamped Done tasks all look too recent to purge.

Fix incoming with regression tests.

pentux-GitHub:
The CalDAV sync stamped the closed date of every closed task it filled with the day of the sync, destroying historical dates in waves. Full mechanism and field evidence on #1357.

The status field applied its value through Task.set_status(), the user action handler, which stamps date_closed with today, propagates the status to every subtask and can trigger the recurrence duplication logic. A sync must do none of that: Status.write_gtg() now sets the status properties directly, like the XML loader does, and leaves the closed date to the completed field. A todo coming back reopened has its stale closed date cleared, since it no longer carries a COMPLETED field to overwrite it.

Red before, green after, on a live Nextcloud calendar of 138 closed todos: master writes 138 tasks closed on the import day down to the data file, the fix preserves every historical server date through the full import, save and reload cycle. Three regression tests added, whole backend suite green.

Out of scope, tracked separately: the pre-existing race between the import thread and the GTK sorters that can crash the app during an import, on master as on this branch.

Fixes #1357

### DIFF
--- a/GTG/backends/backend_caldav.py
+++ b/GTG/backends/backend_caldav.py
@@ -759,8 +759,18 @@ class Status(Field):
         return self._translate(dav_value=super().get_dav(todo, vtodo))[1]
 
     def write_gtg(self, task: Task, value, namespace: str = None):
-        value = self._translate(dav_value=value, gtg_value=value)[0]
-        return super().write_gtg(task, value, namespace)
+        status = self._translate(dav_value=value, gtg_value=value)[0]
+        # Applying the status through Task.set_status() would overwrite
+        # date_closed with today (the COMPLETED field owns that value),
+        # propagate the status to every subtask and trigger the
+        # recurrence duplication logic. A sync must do none of that:
+        # set the properties directly, like the XML loader does.
+        task.status = status
+        task.is_active = (status == TaskStatus.ACTIVE)
+        if status == TaskStatus.ACTIVE:
+            # A reopened todo carries no COMPLETED field anymore, so
+            # the date field leaves the stale value in place: clear it.
+            task.date_closed = Date.no_date()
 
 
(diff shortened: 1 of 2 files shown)


diegogangl:
Thanks!

PR #1370, opened 2026-09-05
appdata: Add keywords to improve discoverability

nekohayo:
The description is not used for searching by some software center apps, like GNOME Software.

Use keywords to make this app easier to find, as per this documentation: https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-keywords

This is similar to what I've done in other GNOME apps.

I'd like to get this merged first before I bring in more merge requests to fix up translatable strings, prepare the translation template, provide my own updated translation, etc.

Issue #1362, opened 2026-09-05
`delete_tag.py` deletes children tags without warning


nekohayo:
In #1361 I tried figuring out why `delete_tag.py` is so messy in terms of UI, wording and code implementation, and it turns out that the complexity conceptually came from `delete_task.py`'s dialog for deleting selected tasks... because tags, just like tasks, can be hierarchical (i.e. have children), so it's not just about the _selected_ tags, but also their children if any.

The problem: while deleting a parent task does list every child that might get deleted:

<img width="859" height="206" alt="Image" src="https://github.com/user-attachments/assets/4c311204-411a-4e59-a84a-45f25807665d" />

…selecting a parent tag does not list the children tags that will get deleted, and they _do_ get removed and deleted:

https://github.com/user-attachments/assets/9947511e-ab2d-42d9-b8d8-f561392b9e93

There are two potential solutions to this. Either:
* We list every tag (and children) that will get removed and deleted, keeping the same warning workflow as the one for tasks, or;
* We just don't delete non-selected children tags, we just deparent them before deleting the parent. If the user wants to delete various children at the same time as the parent, then they would need to select them as well.

As hinted in #1361, I think the 2nd approach would lead to a _much_ simpler UI (and translations), and possibly much simpler code.

Issue #1361, opened 2026-09-05
Refactor the `delete_tag.py` dialog so that it works with 10x simpler code, wording and translations


nekohayo:
While updating the translation template (POT file) and reviewing translations changes to detect accidental translatable strings problems today, I discovered a bunch of problems with  `GTG/gtk/browser/delete_tag.py` based on how it was last reworked in 2020, in commit 6b51ebf01.

_When it works,_ in looks terrible:

<img width="972" height="174" alt="Image" src="https://github.com/user-attachments/assets/ea42eb97-8e11-4144-a988-b3ff4b84254d" />

<img width="972" height="227" alt="Image" src="https://github.com/user-attachments/assets/96a823c9-e438-4851-b3d4-09f0db0cceca" />

…but also, sometimes it doesn't work at all, showing a blank dialog with only the tag names and the two buttons, because it does sketchy unexplained stuff like this line:

```python
label_text = label_text[0:label_text.find(":") + 1]
```

But also:
* Everything stuffed into a heading, nothing in the dialog window's title and nothing in the body
* The meaning of the strings is wrong, the label types is wrong (it's confusing, and lacking plurals in the plural form of the string, and sometimes just lying)
* It might be good to show the number of affected selected tags (if more than one), but who cares about seeing a list of the selected tags' names? The user just selected them, of course they know the names already!

My proposal:
* Set a "Confirm tag deletion" "Confirm tags deletion" window title with ngettext
* Provide the general explanation message via ngettext as a body-text-only styled paragraph.
* If we can make it ignore children tags (see #1362) without deleting them, then do not show the list of tag names, nuke all that overly complicated code, it's fragile and complexifies everything.
* Use the tags count in the confirmation button (with ngettext)

nekohayo:
The whole "list the name of every tag" complexity comes from an incomplete implementation that was supposed to take into account listing the children tags that would get deleted alongside a selected parent: #1362. If we were to eliminate that behavior, then we could simplify this dialog. Otherwise, we still have to refactor and bugfix it, but ideally find a way to clean it up visually and in a way that is not a nightmare for handling translations.

Issue #913, opened 2022-07-08
Replace confirmation dialogs with undo and toasts/notifications


diegogangl:
This is a common Gnome pattern, just do the action and then let the user `ctrl+z` to undo. This also implies adding undo at a global scale through the app, for instance there's no undo for marking a task as done

nekohayo:
Well, the only permanent action I can recall is _deleting_ tasks (thus purging them from the XML data file) or deleting tags. Why would I want to undo those? If I want to delete/purge tasks, that's because I really want them gone from the file, mainly for performance / cleanliness / reliability reasons.

Closing/dismissing tasks can always be undone by reopening them, so those actions already have a way for users to revert them.

Is there an additional usecase am I forgetting?

nekohayo:
This cannot work if deleting a parent task can delete its children, the dialog is there to warn the user with the list of (sub-)tasks that will get deleted.

Whether that should be needed for deleting tags is a different question (#1362)...

However, what we _can_ do is refactor the `delete_task.py` dialog to at least not look as terribly unbalanced as what we're seeing here (so let's retitle this ticket here to be about that):

<img width="859" height="206" alt="Image" src="https://github.com/user-attachments/assets/4c311204-411a-4e59-a84a-45f25807665d" />

PR #1372, opened 2026-09-05
Translations 2026-W36 bugfixes and preparatory cleanups

nekohayo:
See commits.

This does not commit the updated POT file yet (nor my FR translation), but it sets the ground work, alongside #1370.

Ideally I'd like to hear from y'all what approach we should take for #1362 and #1361, as this would have a big impact on how that dialog's translatable strings would be written (maybe likewise for the tasks deletion dialog in #913). It depends if we want to bother cleaning up the code and UI of those dialogs before the release, or if we just want to get it over with ASAP by simply correcting the meaning of the strings per the existing workflow once #1362 is solved somehow.

c.c. @pentux-GitHub


diegogangl (approved):

diegogangl:
LGTM!

PR #1374, opened 2026-09-05
fix(caldav): flush unresolved todos instead of dropping them at the depth cap (#1373)

### LINKED ISSUE #1373: CalDAV sync logs `Too deep, 500th recursion isn't allowed` on the first import
I added two Nextcloud CalDAV accounts, and the very first sync printed this in
the terminal:

```
2026-09-05 09:04:32,146 - ERROR - backend_caldav:__sort_todos:484 - Too deep, 500th recursion isn't allowed
```

It appeared once, between the `Fetching todos from '…/personal/'` line and the
line for the next calendar, so I assume it is about that one. My other calendars
printed nothing.

That sync was the first one, on an empty GTG database, and it created 415 tasks.
Starting GTG again afterwards, the message does not come back.

I am not sure if it has consequences or not.

GTG master (5045363b), Python 3.14, Arch Linux.

AmirF194:
Confirmed the recursion is a real cycle/dangling-parent case, not just a deep hierarchy: `__sort_todos` only yields a todo once its parent is already known, so a RELATED-TO cycle or a parent UID pointing outside the synced batch never resolves and the loop makes zero progress on those todos every pass. At the 500th pass the depth cap logs the error and `break`s, which exits the generator without ever yielding them, so `_import_calendar_todos`'s `for todo in self.__sort_todos(todos)` loop never sees them and they are never created.

Flushed the still-stuck todos when the cap fires instead of dropping them. They come in without confirmed parent ordering (the existing parent-wiring code already treats a missing parent as fine, it only links when the parent task is found), which beats not coming in at all. The log line now also names which UIDs were affected.

Added `test_caldav_sort_todos_depth_cap.py`: a two-todo RELATED-TO cycle and a dangling-parent UID both used to vanish silently, both are now flushed, and a normal single-todo batch is a control that must never hit the cap. Fails on master (2 of 3), passes on the branch, `tests/backend/` and the full suite (`xvfb-run pytest tests/`) both green, 335 passed / 1 skipped / 1 xfailed unrelated. `pre-commit run` clean on the changed files.

Not fixed: the same `__sort_todos(self, todos, max_depth=500)` signature has an unused `max_depth` parameter, the loop actually checks the module-level `MAX_CALENDAR_DEPTH` constant. Looked like a separate, pre-existing thing, left it alone.

Fixes #1373


diegogangl:
LGTM, thanks!

AmirF194:
Glad the flush-instead-of-drop approach works here, felt like the safer default over trying to resolve the cycle itself.

